PR #9, opened 2019-12-04
fix for index arithmetic

sudiptoguha:
*Description of changes:* The extrapolation code for the cyclic case had incorrect index arithmetic


wnbts:
Looks good. Is it possible to add tests to verify and prevent regressions?

jotok:
I believe that the existing code can throw an NPE. Given that we are still in alpha, there is no existing test coverage for the extrapolation methods, and writing unit tests for the extrapolate methods will require some finicky mocking, I'm inclined to merge this PR as is and create a new Issue to add test coverage. Does that work for you  @wnbts ?

wnbts:
OK

jotok:
Issue to add test coverage #10

PR #24, opened 2020-04-30
support binary serialization with ion

wnbts:
Current RCF serialization results are text json, which is less efficient than binary for storage. This change adds support for serialization into the [ion](https://github.com/amzn/ion-java/tree/master) binary format.

Experiments shows for the same model (100 trees, 256 samples, 1~10 dimension), binary format saves 57%~67% storage.

### DIFF
--- a/src/main/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDe.java
+++ b/src/main/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDe.java
@@ -15,6 +15,8 @@
 
 package com.amazon.randomcutforest.serialize;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ForkJoinPool;
@@ -22,9 +24,17 @@ import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import software.amazon.ion.system.IonBinaryWriterBuilder;
+import software.amazon.ion.system.IonReaderBuilder;
+import software.amazon.ion.system.IonTextWriterBuilder;
+import software.amazon.ion.IonException;
+import software.amazon.ion.IonReader;
+import software.amazon.ion.IonWriter;
 import com.amazon.randomcutforest.AbstractForestTraversalExecutor;
 import com.amazon.randomcutforest.RandomCutForest;
 import com.amazon.randomcutforest.TreeUpdater;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.dataformat.ion.IonObjectMapper;
 import com.google.gson.ExclusionStrategy;
 import com.google.gson.FieldAttributes;
 import com.google.gson.Gson;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDe.java
```diff
@@ -77,6 +87,29 @@ public String toJson(RandomCutForest rcf) {
...
+     * Serializes a RCF object to ion binary.
+     *
+     * @param rcf a RCF object
+     * @return ion binary serialized from the RCF
+     * @throws IonException when the input cannot be serialized
+     */
+    public byte[] toIon(RandomCutForest rcf) {
+        try {
+            IonReaderBuilder readerBuilder = IonReaderBuilder.standard();
+            IonReader reader = readerBuilder.build(toJson(rcf));
```
jotok, line 100: Ion supports more types than JSON. Are we losing out by going through the JSON serializer first? If we're not taking advantage of other Ion features, is this actually more efficient (in terms of compute time or storage space) than just gzipping the JSON output?

src/main/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDe.java
```diff
@@ -77,6 +87,29 @@ public String toJson(RandomCutForest rcf) {
...
+            IonObjectMapper mapper = new IonObjectMapper();
+            mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
```
jotok, line 102: Why do we need to set this?

src/main/java/com/amazon/randomcutforest/serialize/SerializedRandomCutForest.java
```diff
@@ -0,0 +1,305 @@
...
+
+    public Executor getExecutor() {
+        return executor;
+    }
+
+    public void setExecutor(Executor executor) {
+        this.executor = executor;
+    }
+
+    private static class Random {
```
jotok, line 111: What does this do?

src/main/java/com/amazon/randomcutforest/serialize/SerializedRandomCutForest.java
```diff
@@ -0,0 +1,305 @@
...
+            return threadPoolSize;
+        }
+
+        public void setThreadPoolSize(int threadPoolSize) {
+            this.threadPoolSize = threadPoolSize;
+        }
+    }
+
+    private static class Executor {
+        private String executor_type;
```
jotok, line 274: Why is this in snake_case?

src/main/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDe.java
wnbts: The primary reason of reusing existing json is it's ready as building block. I tried IonMapper, customization of serialization/deserialization for non-pojo objects with many final private fields proves a challenge, which Gson works around nicely.

For our use case, serialization cost is infrequent but storage cost is long-lived. I thought about gzipping but am concerned the savings are not consistent or optimal. It is however worth an experiment.

src/main/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDe.java
wnbts: Since during serialization random objects are removed and represented by empty objects, without which they won't be de-serialized and be left to null.

src/main/java/com/amazon/randomcutforest/serialize/SerializedRandomCutForest.java
wnbts: That was an additional property added to the json result for denoting the type. Since it's part of json, I habitually used snake_case.

src/main/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDe.java
wnbts: This pr is more of a test to see if this use case is specific-purpose.

jotok: I see. I think the only concern would be if at some point we want to create a "proper" Ion format, I think it would be incompatible with this change (we would either end up with two `toIon` methods, or we'd have a backwards-incompatible change to `toIon`/`fromIon`. How do you think we should handle that?

PR #26, opened 2020-05-14
Refactor into submodules

jotok:
This change refactors the project into the following maven modules:

* randomcutforest-core: The core algorithm library, which defines trees, forests, etc.
* randomcutforest-serialization: The JSON serialization code.
* randomcutforest-benchmark: The benchmarking code based on JMH.
* randomcutforest-testdata: Data generation classes that are used in the core library and the benchmarking code.

The purpose of this refactoring is to help us maintain a lean core library with few external dependencies. I don't think the existing code is bloated, but making this change now will make it easier to add, e.g., jpype support in the future in its own module.

Questions to discuss:
1. Do we like the artifact IDs defined here?
2. Should serialization be merged back into the core library? It's nice that under the current refactoring the core library doesn't take any runtime dependencies at all. We could also change the artifact ID for the existing code to randomcutforest-json, leaving it open to support other kinds of serialization in the future (and most users should only need one kind of serialization for their application). On the other hand, serialization a fundamental need for most production systems, and JSON serialization seems like a good default to ship with the library. 


README.md
```diff
@@ -2,11 +2,11 @@
 
 This package is an implementation of the Random Cut Forest probabilistic data structure. Random Cut Forests (RCFs) were 
 originally developed at Amazon to use in a nonparametric anomaly detection algorithm for streaming data. Later new 
-algorithms based on RCFs were developed for density estimation, imputation, and forecasting. The goal of this library 
+algorithms based on RCFs were developed for density estimation, com.amazon.randomcutforest.imputation, and forecasting. The goal of this library
```
jotok, line 5: Fix unintentional "refactoring".

README.md
```diff
@@ -2,11 +2,11 @@
 
 This package is an implementation of the Random Cut Forest probabilistic data structure. Random Cut Forests (RCFs) were 
 originally developed at Amazon to use in a nonparametric anomaly detection algorithm for streaming data. Later new 
-algorithms based on RCFs were developed for density estimation, imputation, and forecasting. The goal of this library 
+algorithms based on RCFs were developed for density estimation, com.amazon.randomcutforest.imputation, and forecasting. The goal of this library 
 is to be easy to use and to strike a balance between efficiency and extensibility.
 
 The public interface to this package is the RandomCutForest class, which defines methods for anomaly detection, anomaly 
-detection with attribution, density estimation, imputation, and forecasting.
+detection with attribution, density estimation, com.amazon.randomcutforest.imputation, and forecasting.
```
jotok, line 9: Fix unintentional "refactoring".

README.md
```diff
@@ -2,11 +2,11 @@
```
jotok, line 2: I still need to review the README and update the instructions so that they're accurate with the new project layout.

core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -47,7 +47,7 @@
 
 /**
  * The RandomCutForest class is the interface to the algorithms in this package, and includes methods for anomaly
- * detection, anomaly detection with attribution, density estimation, imputation, and forecasting. A Random Cut Forest
+ * detection, anomaly detection with attribution, density estimation, com.amazon.randomcutforest.imputation, and forecasting. A Random Cut Forest
```
jotok, line 50: Fix unintentional "refactoring".

core/src/main/java/com/amazon/randomcutforest/anomalydetection/AbstractAttributionVisitor.java
```diff
@@ -116,7 +116,7 @@ public void accept(Node node, int depthOfNode) {
         if (hitDuplicates || ignoreLeaf) {
             // use the sibling bounding box to represent counterfactual "what if point & the candidate near neighbor
             // had not been inserted in the tree"
-            Node sibling = isLeftOf(pointToScore, node) ? node.getRightChild() : node.getLeftChild();
+            Node sibling = Node.isLeftOf(pointToScore, node) ? node.getRightChild() : node.getLeftChild();
```
jotok, line 119: Check to see if this is necessary, might have been added by intellij while moving files around.

core/src/main/java/com/amazon/randomcutforest/anomalydetection/AbstractScalarScoreVisitor.java
```diff
@@ -150,7 +150,7 @@ public void accept(Node node, int depthOfNode) {
                 return;
             }
         } else {
-            Node sibling = isLeftOf(pointToScore, node) ? node.getRightChild() : node.getLeftChild();
+            Node sibling = Node.isLeftOf(pointToScore, node) ? node.getRightChild() : node.getLeftChild();
```
jotok, line 153: Check to see if this is necesary.

core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -55,7 +55,7 @@ public ImputeVisitor(double[] queryPoint, int numberOfMissingValues, int[] missi
         this.numberOfMissingValues = numberOfMissingValues;
 
         for (int i = 0; i < this.numberOfMissingValues; i++) {
-            checkArgument(0 <= missingIndexes[i] && missingIndexes[i] < queryPoint.length,
+            CommonUtils.checkArgument(0 <= missingIndexes[i] && missingIndexes[i] < queryPoint.length,
```
jotok, line 58: Use static import.

core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -93,7 +93,7 @@ public double getRank() {
      */
     public void accept(final Node node, final int depthOfNode) {
 
-        double probabilityOfSeparation = getProbabilityOfSeparation(node.getBoundingBox(), queryPoint);
+        double probabilityOfSeparation = CommonUtils.getProbabilityOfSeparation(node.getBoundingBox(), queryPoint);
```
jotok, line 96: Use static import.

core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -116,7 +116,7 @@ public void acceptLeaf(final Node leafNode, final int depthOfNode) {
                 queryPoint[i] = leafNode.getBoundingBox().getMinValue(i);
             }
         }
-        double probabilityOfSeparation = getProbabilityOfSeparation(leafNode.getBoundingBox(), queryPoint);
+        double probabilityOfSeparation = CommonUtils.getProbabilityOfSeparation(leafNode.getBoundingBox(), queryPoint);
```
jotok, line 119: Use static import.

core/src/main/java/com/amazon/randomcutforest/interpolation/SimpleInterpolationVisitor.java
```diff
@@ -98,7 +98,7 @@ public void accept(Node node, int depthOfNode) {
 
             // use the sibling bounding box to represent counterfactual "what if point & the candidate near neighbor
             // had not been inserted in the tree"
-            Node sibling = isLeftOf(pointToScore, node) ? node.getRightChild() : node.getLeftChild();
+            Node sibling = Node.isLeftOf(pointToScore, node) ? node.getRightChild() : node.getLeftChild();
```
jotok, line 101: Review.

core/src/test/java/com/amazon/randomcutforest/returntypes/DensityOutputTest.java
```diff
@@ -125,7 +126,7 @@ public void testGetDensity() {
             sumOfFactors += t;
         }
 
-        assertEquals(sumOfPoints / (q*sumOfPoints+sumOfFactors), density, EPSILON);
+        Assertions.assertEquals(sumOfPoints / (q*sumOfPoints+sumOfFactors), density, EPSILON);
```
jotok, line 129: use static import.

core/src/test/java/com/amazon/randomcutforest/interpolation/SimpleInterpolationVisitorTest.java
```diff
@@ -209,7 +210,7 @@ public void testAccept() {
         int sampleSize = 50;
         SimpleInterpolationVisitor visitor = new SimpleInterpolationVisitor(pointToScore, sampleSize,1,false);
 
-        Node leafNode = spy(new Node(new double[] {1.0, -2.0}));
+        Node leafNode = Mockito.spy(new Node(new double[] {1.0, -2.0}));
```
jotok, line 213: use stack import, here and in the rest of the file.

jotok:
It looks like intellij changed a bunch of static imports while I was copying stuff around. I'll try to fix them all so that we're using a consistent style throughout the project.

core/src/main/java/com/amazon/randomcutforest/anomalydetection/AbstractAttributionVisitor.java
jotok: For the `isLeftOf` method, I think I actually prefer to keep the class name as part of the method call.

core/src/main/java/com/amazon/randomcutforest/anomalydetection/AbstractScalarScoreVisitor.java
jotok: Keeping this line and removing unnecessary static import.

core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
jotok: Keeping the class name.

core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
jotok: Keeping the class name.

core/src/main/java/com/amazon/randomcutforest/interpolation/SimpleInterpolationVisitor.java
jotok: Keeping as is.

jotok:
Per outside discussion, I've updated the PR to use the following module names:

* randomcutforest-core
* randomcutforest-serialization-json
* randomcutforest-benchmark
* randomcutforest-testutils

benchmark/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.amazon</groupId>
+    <artifactId>randomcutforest-parent</artifactId>
+    <version>1.0</version>
```
wnbts, line 9: wonder if version should be bumped?

jotok: I've been thinking that it's not _really_ 1.0 until we push to Maven. We're also changing from an artifact ID of `random-cut-forest:1.0` for the whole package, versus `randomcutforest-core:1.0` for the core library, etc.

If you think it would be better, I could change the version numbers in this CR to something like 1.0-alpha, and we can remove the alpha just before pushing to Maven.

wnbts: yes, good point, I am good with 1.0 then. My minor concern was existing code already built on pre-change api. Since the artifacts are new, an explicit upgrade is needed.

Issue #20, opened 2020-01-28
Intermittent failure in RandomCutForestSerDeTests


jotok:
I'm seeing an occasional failuree in the RandomCutForestSerDeTests class. The tests usually pass on re-run. I'm creating this issue to track occurrences.

Example
```
[ERROR] Tests run: 16, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 22.914 s <<< FAILURE! - in com.amazon.randomcutforest.serialize.RandomCutForestSerDeTests
[ERROR] toJsonString{int, int, int, int, int, int, int}[9]  Time elapsed: 0.55 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <1.394635311942881> but was: <1.8229730367398675>
        at com.amazon.randomcutforest.serialize.RandomCutForestSerDeTests.toJsonString(RandomCutForestSerDeTests.java:75)
```

wnbts:
I tried multiple times locally but could not produce failures. When this issue re-occurs, the configuration for the failed test cases can be recorded in this issue.

wnbts:
Looks like this issue hasn't occurred for a while. Closing for now.

PR #33, opened 2020-05-23
use fixed test data for serialization test

wnbts:
#20 : Intermittent failure in RandomCutForestSerDeTests

This pr fixes the random seed used to generate test data so the test results are less randomized.


jotok:
Did you figure out what was causing the sporadic failure? Does this solve the root cause, or does the failure only happen with certain RNG seeds (so explicitly setting the seed will avoid the failure)?

wnbts:
There are two sources of randomicity. Test data is one and deserialization is the other. This change removes that from test data to mitigate the issue. If there is still occurrence, the test case can be updated to be more tolerant.

Issue #29, opened 2020-05-15
Review default forest parameters


jotok:
Before pushing to Maven, let's review and update the default forest parameters. Current defaults:

| Parameter Name | Type | Description | Default Value|
| --- | --- | --- | --- |
| centerOfMassEnabled | boolean | If true, then tree nodes in the forest will compute their center of mass as part of tree update operations. | false |
| dimensions | int | The number of dimensions in the input data. | Required, no default value |
| lambda | double | The decay factor (lambda value) used by stream samplers in this forest. | 1e-5 |
| numberOfTrees | int | The number of trees in this forest. | 100 |
| outputAfter | int | The number of points required by stream samplers before results are returned. | 0.25 * sampleSize |
| parallelExecutionEnabled | boolean | If true, then the forest will create an internal threadpool. Forest updates and traversals will be submitted to this threadpool, and individual trees will be updated or traversed in parallel. | true |
| randomSeed | long | A seed value used to initialize the random number generators in this forest. | |
| sampleSize | int | The sample size used by stream samplers in this forest | 256 |
| storeSequenceIndexesEnabled | boolean | If true, then sequence indexes (ordinals indicating when a point was added to a tree) will be stored in the forest along with poitn values. | false |
| threadPoolSize | int | The number of threads to use in the internal threadpool. | Number of available processors - 1 |
| windowSize | int | An alternate way of specifying the lambda value. Using this parameter will set lambda to 1 / windowSize. | |

To discuss:
* Given what we know about convergence of anomaly scores, 100 trees is probably to many for a default. Proposal: 50.
* Should we change the default for parallel execution to false? From our benchmark results for 50 trees, parallel execution only consistently wins for 256 dimensional points.
* The lambda value in this code was taken from an older implementation and not re-examined. Can we pick a better default? What guidance can we give to users to help them pick a meaningful value?
* I like the idea of windowSize as a more user-friendly way of setting lambda, but does it have a precise mathematical interpretation that we can explain? If not, could we come up with a better choice for a user-friendly setting? For example, something like: "if you set this parameter to 100, then the probability that more than half of your in-sample points comes from the last 100 submitted points is greater than 90%".

wnbts:
1. 50 is fine. is there any benchmarking dataset?
2. yes.
3/4. math is hard. my guess it might depend on a. lambda/window size range, and b. specific implementation to be summarized into one/multiple bounds.

sudiptoguha:
[...]

PR #34, opened 2020-05-26
update default config changing parallel execution to sequential execu…

wnbts:
*Issue #, if available:* https://github.com/aws/random-cut-forest-by-aws/issues/29

*Description of changes:* This pr changes default execution config from parallel to sequential.

### DIFF
--- a/core/src/test/java/com/amazon/randomcutforest/RandomCutForestBuilderTest.java
+++ b/core/src/test/java/com/amazon/randomcutforest/RandomCutForestBuilderTest.java
@@ -73,8 +73,8 @@ public class RandomCutForestBuilderTest {
         assertEquals(256 / DEFAULT_OUTPUT_AFTER_FRACTION, f.getOutputAfter());
         assertFalse(f.storeSequenceIndexesEnabled());
         assertFalse(f.centerOfMassEnabled());
-        assertTrue(f.parallelExecutionEnabled());
-        assertEquals(Runtime.getRuntime().availableProcessors() - 1, f.getThreadPoolSize());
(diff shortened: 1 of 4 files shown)


jotok:
We also need to update the default parameters table in the README. Can you also add something like the following to the description of the parallelExecutionEnabled parameter: "For larger shingle sizes, dimensions, and number of trees, parallelization may improve throughput. We recommend users benchmark against their target use case."

wnbts:
I updated the readme documentation there.

core/src/test/java/com/amazon/randomcutforest/RandomCutForestBuilderTest.java
```diff
@@ -73,8 +73,8 @@ public void testDefaultForestWithDimensionArgument() {
         assertEquals(256 / DEFAULT_OUTPUT_AFTER_FRACTION, f.getOutputAfter());
         assertFalse(f.storeSequenceIndexesEnabled());
         assertFalse(f.centerOfMassEnabled());
-        assertTrue(f.parallelExecutionEnabled());
-        assertEquals(Runtime.getRuntime().availableProcessors() - 1, f.getThreadPoolSize());
```
jotok, line 77: Can we add a new test to verify that thread pool size is being set correctly when parallel execution is set to true but no explicit value is given for thread pool size?

wnbts: yes, i just added back this test case named testForestBuilderWithDefaultParallelExecutionThreadPoolSize .

PR #36, opened 2020-05-26
Move Maven project into Java subdirectory

### LINKED ISSUE #28: Proposal: new directory structure to multiple language implementations
We are currently developing a second implementation of RCFs in C. We intend to support both the Java and C versions going forward, since the two implementations have different tradeoffs. This issue is to discuss how we should reorganize the project directory to accomodate both implementations.

**Proposal 1**: Root level language directories
```
c/
java/
python/
```
With this approach, the existing Maven project would be moved to a new java/ subdirectory, and we would add a top-level c/ directory to contain the C implementation. We could then continue this pattern for other language bindings. For example, the python/ directory could contain a Python project with bindings to the C library.

If we take this approach we'll have to create a new top-level README explaining the setup and possibly giving some information about RCF the algorithm. We'll also have to double-check that we can execute our github workflow from a subdirectory.

**Proposal 2**: Separate repositories

This proposal is to create a new repository for the C implementation. This approach would all both repos to conform to the standard project layout wherein the root directory for the project contains the build configuration files for your project.

By taking this approach we would lose some discoverability between the two implementations, and it would be less clear that the two implementations conform to a single "vision" for RCFs.

**Also considered**: Add the C library to a directory in the Maven project
```
benchmark/
core/
...
rcfc/
```
In this approach, the folder containing the C code would live in the Maven project, but it would be ignored by the Maven build. I think the main drawback here is that the C and Java versions would not appear as different-but-equally-good implementations. People who get to the repository via search might not even notice that there is a C implementation.

jotok:
Closes #28

Per discussion in #28, this PR moves the existing Maven project into a Java subdirectory in our repository. This setup will allow us to release a C implementation of RCFs into the same repository in the future without affecting the Java implementation.


Java/README.md
```diff
@@ -0,0 +1,169 @@
...
+This directory contains a Java implementation of the Random Cut Forest data structure and algorithms
+for anomaly detection, density estimation, imputation, and forecast. The goal of this library 
+is to be easy to use and to strike a balance between efficiency and extensibility.
+
+## Basic operations
+
+To create a RandomCutForest instance with all parameters set to defaults:
+
+```java
+int dimensions = 1; // The number of dimensions in the input data, required
```
sudiptoguha, line 12: Can we make this 5 as an example different from 1?

Java/README.md
```diff
@@ -0,0 +1,169 @@
...
+The full test suite including functional tests currently takes over 10 minutes to complete. If you are contributing to
+this package, we recommend excluding the functional tests while actively developing, and only running the full test
+suite before creating a pull request. Functional tests can be excluded from Maven build targets by passing
+`-DexcludedGroups=functional` at the command line. For example:
+
+```text
+% mvn test -DexcludedGroups=functional
+```
+
+We currently have 90% line coverage with the full test suite, and 80% line coverage when running the unit tests only
```
jotok, line 137: We should change this to reflect coverage in the core library, which is currently at 79% according to jacoco.

Java/README.md
jotok: Fixed in new commit.

Issue #47, opened 2020-07-16
Signed zero causes NullPointerException


kaituo:
Running the unit test can reproduce the issue:

```
RandomCutForestFunctionalTest:

    // Use this ArgumentsProvider to run a test on both single-threaded and
    // multi-threaded forests
    static class TestForestProvider2 implements ArgumentsProvider {
        @Override
        public Stream<? extends Arguments> provideArguments(ExtensionContext context) throws Exception {
            RandomCutForest forest = RandomCutForest.builder().numberOfTrees(numberOfTrees).sampleSize(sampleSize)
                    .dimensions(1).randomSeed(randomSeed).centerOfMassEnabled(true)
                    .storeSequenceIndexesEnabled(true).parallelExecutionEnabled(false).build();
            return Stream.of(forest).map(Arguments::of);
        }
    }

    @ParameterizedTest
    @ArgumentsSource(TestForestProvider2.class)
    public void testSideEffectsC(RandomCutForest forest) {
        /* the changes to score and attribution should be in sync */
        forest.update(new double[] { 0.0 });
        forest.update(new double[] { -0.0 });
    }

```

NullPointerException will be thrown:

```
java.lang.NullPointerException: null
        at com.amazon.randomcutforest.tree.Cut.isLeftOf(Cut.java:52) ~[random-cut-forest-1.0.jar:?]
        at com.amazon.randomcutforest.tree.Node.isLeftOf(Node.java:134) ~[random-cut-forest-1.0.jar:?]
        at com.amazon.randomcutforest.tree.RandomCutTree.addPoint(RandomCutTree.java:348) ~[random-cut-forest-1.0.jar:?]
        at com.amazon.randomcutforest.tree.RandomCutTree.addPoint(RandomCutTree.java:349) ~[random-cut-forest-1.0.jar:?]
        at com.amazon.randomcutforest.tree.RandomCutTree.addPoint(RandomCutTree.java:349) ~[random-cut-forest-1.0.jar:?]
        at com.amazon.randomcutforest.tree.RandomCutTree.addPoint(RandomCutTree.java:349) ~[random-cut-forest-1.0.jar:?]
        at com.amazon.randomcutforest.tree.RandomCutTree.addPoint(RandomCutTree.java:349) ~[random-cut-forest-1.0.jar:?]
        at com.amazon.randomcutforest.tree.RandomCutTree.addPoint(RandomCutTree.java:349) ~[random-cut-forest-1.0.jar:?]
[...]

PR #48, opened 2020-07-16
handle signed zeros in data points

wnbts:
*Issue #47* 

*Description of changes:* This is a fix for the issue caused by signed zeros described in the issue. This change maps negative zeros to positive zeros before passing them down for sampling and updating. 

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/AbstractForestTraversalExecutor.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/AbstractForestTraversalExecutor.java
@@ -47,7 +47,7 @@ public abstract class AbstractForestTraversalExecutor {
      */
     public void update(double[] point) {
         totalUpdates++;
-        double[] pointCopy = Arrays.copyOf(point, point.length);
+        double[] pointCopy = cleanCopy(point);
         update(pointCopy, totalUpdates);
     }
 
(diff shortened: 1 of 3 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/AbstractForestTraversalExecutor.java
```diff
@@ -192,4 +192,22 @@ public void update(double[] point) {
...
+     *
+     * Current clean-ups include changing negative zero -0.0 to positive zero 0.0.
+     *
+     * @param point The original data point.
+     * @return a clean deep copy of the original point.
+     */
+    protected double[] cleanCopy(double[] point) {
+        double[] pointCopy = Arrays.copyOf(point, point.length);
+        for (int i = 0; i < point.length; i++) {
+            if (pointCopy[i] == 0.) {
```
jotok, line 207: Elsewhere in this code base we're using explicit `.0`. Can we write it out here for consistency?

jotok:
The GitHub build workflow wasn't triggered because of the change in the default branch name. Let's try merging #49 first and rebasing this change to validate the workflow fix.

Java/core/src/main/java/com/amazon/randomcutforest/AbstractForestTraversalExecutor.java
wnbts: yes, i have changed them to 0.0

Issue #46, opened 2020-07-14
deprecate master branch


wnbts:
The name "master" is founded on inequality. Since it is used as the name of the most current branch, it is inevitable for contributors to use it for pulling/pushing. For current or potential contributors, this may be or may have been a source of discomfort/rejection, against [Code of Conduct](https://opendistro.github.io/for-elasticsearch/codeofconduct.html). Neutral names such as "main" have been proposed and used.

The [RFC](https://tools.ietf.org/id/draft-knodel-terminology-00.html) provides reasonable arguments against this language. [Github](https://www.bbc.com/news/technology-53050955) , as well as [many others](https://www.forbes.com/sites/elanagross/2020/07/03/twitter-and-jpmorgan-chase-are-the-latest-companies-to-remove-terms-master-and-slave-from-programming-code/#68da80223ced) , are taking the actions to end the use of the term.

jotok:
Thanks for creating this issue, let's make the change.

jotok:
I executed the steps in [this guide](https://dev.to/afrodevgirl/replacing-master-with-main-in-github-2fjf) to move the commit history from 'master' to 'main', push 'main' to github, and set 'main' as the default branch. Let's leave 'master' up for a couple days so people can sync, and if everything looks good we can delete it.

wnbts:
Thank you so much, @jotok !

jotok:
I've deleted the 'master' branch.

PR #49, opened 2020-07-21
Update GitHub workflow to use default branch name 'main'

jotok:
We recently updated our repository to change the name of the default branch to 'main' (#46). This change updates our GitHub build workflow accordingly.


jotok:
I briefly changed the explicit name 'main' to the expression `$default-branch` which I saw in an example I found, but I'm not able to find any GitHub documentation for this usage so I'm reverting to use the explicit branch name for now.

PR #51, opened 2020-08-14
Use floats instead of double in PointStore

jotok:
*Description of changes:*

The PointStore will be used with as part of the new compact tree data structure. The purpose of creating compact trees is to save space, and storing floats instead of doubles will help with that. Moreover, we believe that typical uses for this algorithm should not depend on the additional precision provided by doubles.


sudiptoguha:
We should revisit double -> float (internally) for the whole library once this branch is finished.

PR #52, opened 2020-08-15
Add new store classes for review

jotok:
This PR introduces new Store classes:

* NodeStore and LeafStore are "column-oriented" store classes. Node data and leaf data is stored in arrays in the store classes and retrieved by index value. These classes have minimal memory overhead to store the data required to represent a tree, but the way they're used is a little unusual for Java.
* The `genericstore` package shows how we might implement stores using Java generics. These classes are "row-oriented": each storage class maintains an `ArrayList<T>` and elements in the list are managed by index. The idea is that, instead of having `NodeStore` and `LeafStore`, we would have`SmallStore<CompactNode>` and `SmallStore<CompactLeaf>`. The Java in these classes looks more natural, but we pay the cost of storing an array of pointers in addition to the data fields.


Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @param <T> The type of value being stored.
+ */
+public class Store<T> {
+
+    private final int capacity;
+    private final List<T> store;
+    private final int[] freeBlockStack;
+    private int freeBlockPointer;
+    private final BitSet occupied;
```
wnbts, line 43: question. is this for error check?

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
```diff
@@ -0,0 +1,129 @@
...
+     * Save a value to this store and return an index which can be used to look up
+     * the value later.
+     * 
+     * @param t The value being stored.
+     * @return an index that can be used to look up the value later.
+     */
+    public int add(T t) {
+        checkState(freeBlockPointer >= 0, "store is full");
+        int index = freeBlockStack[freeBlockPointer--];
+        occupied.set(index);
```
wnbts, line 87: question. does it need to check if the index is already set?

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/SmallStore.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.List;
+
+/**
+ * This class provides a fixed amount of storage slots for a given type. When an
+ * object is stored, the class returns an index value that can be used to
+ * retrieve the object later. The purpose of this class is to reduce the overall
+ * memory footprint of an applicaiton by allowing classes to keep index values
+ * instead of pointers. The index values used in this class are shorts, which
+ * occupy 2 bytes (Java pointers occupy 8 bytes). Using shorts as index values,
+ * this class has a maximum capacity of 32,767. If more capacity is neede, see
```
wnbts, line 32: minor typo. neede should be needed

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/SmallStore.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return the maximum number of objects that can be stored.
+     */
+    public short getCapacity() {
+        return capacity;
+    }
+
+    /**
+     * @return the number of objects currently being stored.
+     */
+    public int size() {
```
wnbts, line 74: minor. the return type should be short

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
```diff
@@ -0,0 +1,129 @@
...
+ * retrieve the object later. The purpose of this class is to reduce the overall
+ * memory footprint of an applicaiton by allowing classes to keep index values
+ * instead of pointers. The index values used in this class are ints, which
+ * occupy 4 bytes (Java pointers occupy 8 bytes). For further compression, see
+ * the {@link SmallStore} class which uses 2 byte short integers as index
+ * values.
+ *
+ * @param <T> The type of value being stored.
+ */
+public class Store<T> {
```
wnbts, line 37: question. should store depend on/reuse storecommon?

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactLeaf.java
```diff
@@ -0,0 +1,53 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest.tree;
+
+public class CompactLeaf {
+
+    private final int pointIndex;
+    private short parentIndex;
```
wnbts, line 21: question. why is one index int and the other short?

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
```diff
@@ -0,0 +1,142 @@
...
+ * {@link com.amazon.randomcutforest.tree.Node} class.
+ */
+public class NodeStore extends SmallStoreCommon {
+
+    public final short[] parentIndex;
+    public final short[] leftIndex;
+    public final short[] rightIndex;
+    public final int[] cutDimension;
+    public final float[] cutValue;
+    public final int[] mass;
```
wnbts, line 41: question. does mass need int?

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/RCStore.java
```diff
@@ -0,0 +1,64 @@
...
+package com.amazon.randomcutforest.genericstore;
+
+import static com.amazon.randomcutforest.CommonUtils.checkState;
+
+/**
+ * Reference counted store.
+ * 
+ * @param <T>
+ */
+public class RCStore<T> extends Store<T> {
```
theadroit-m, line 25: Should we keep the reference counting a separate concept and not extend the Store<T> class?

This could expose the following APIs
public void incrementRefCount(int index)
public void decrementRefCount(int index)
public int getRefCount(int index)

The PointStore can extend both the Store<T> class and RCStore class

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private final int capacity;
+    private final List<T> store;
```
theadroit-m, line 40: Should this not be an Array? It would consume lesser memory than a List

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
theadroit-m: +1 We are already keeping track of the free "indexes" as part of the freeBlockStack, This is basically duplicate information correct? 

is the plan to maintain this until we get more confidence on freeBlockStack?

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * Note that a NodeStore does not store instances of the
+ * {@link com.amazon.randomcutforest.tree.Node} class.
+ */
+public class NodeStore extends SmallStoreCommon {
+
+    public final short[] parentIndex;
+    public final short[] leftIndex;
+    public final short[] rightIndex;
+    public final int[] cutDimension;
```
theadroit-m, line 39: The dimension can also be a short? 

or maybe a byte if the number of dimensions can be limited to 127? :)

Java/core/src/main/java/com/amazon/randomcutforest/store/StoreCommon.java
```diff
@@ -0,0 +1,94 @@
...
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+import static com.amazon.randomcutforest.CommonUtils.checkState;
+
+import java.util.BitSet;
+
+/**
+ * This class defines common functionality for Store classes, including
+ * maintaining the stack of free pointers.
+ */
+public class StoreCommon {
```
theadroit-m, line 27: 1/ Should this class be an abstract class? 

2/ This and the Store class looks very similar. Should we abstract out the common parts into a base class? This abstraction can also be extended by the SmallStoreCommon.. 

3/ Or Should we make a concrete class called IndexStore? This could just maintain the indexes, the same logic can be used by other Storage classes which want to maintain the indexes? (Still need to think about this)

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
jotok: This is to provide an O(1) check for whether an index is in use or not. The freeBlockStack will contain a list of index values that are not currently being used, but to check whether a given index was in use we would have to diff the freeBlockStack against a list of all valid indexes.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
jotok: That's a good question. It _shouldn't_ be possible to reach a state where an index value is available in the freeBlockStack but the occupied bit is set. We could consider adding a check and throwing an illegal state exception for this.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/SmallStore.java
jotok: Fixed.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/SmallStore.java
jotok: Fixed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactLeaf.java
jotok: The PointStore uses int indexes, because the maximum number of points that we need to store (number of trees * sample size) is on the same order as the maximum short value (~32,000). We thought that using a short index for points could could unnecessarily restrict valid use cases.

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
jotok: It probably doesn't, but the mass value is already defined as an int in the existing Node class and I thought we should keep the int here for consistency. I think it would make sense to change both to short, but maybe in a separate PR since this one is already pretty big.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/RCStore.java
jotok: Java doesn't support extending more than one class. I think the best we could do along these lines would be to make a class like ReferenceCounter which implements those methods, and then give the PointStore a ReferenceCounter instance. Between the two approaches (defining a separate class or keeping the current subclass approach), I don't have a strong opinion.

Keep in mind that the generic store classes are somewhat exploratory: I wanted to see if taking a more standard Java approach would be a big enough improvement over the column-oriented array based approach that it would be worth paying the price in terms of additional memory used. Currently, I feel that it's not enough of an improvement and I think we'll end up deleting these classes from the package.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
jotok: My feeling was that the overhead we were paying to keep an array of pointers was already big enough that it didn't make sense to be stingy about creating the ArrayList<T> versus just creating a T[]. If we decide to keep the generics, I'm fine with making that change, but as I mentioned in another comment my vote is to get rid of the generic versions and keep the column-oriented stores.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
jotok: This PR contains two independent implementations of the store concept: one using generics, and one using non-generic classes with primitive arrays. Eventually, we'll keep only one of the implementations (and in particular, we'll only have one or the other of Store<T> and StoreCommon).

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
jotok: Like with node mass, I left this as an int to be consistent with the existing Cut class. I think we should revisit making that value a short in a separate PR.

Java/core/src/main/java/com/amazon/randomcutforest/store/StoreCommon.java
jotok: This PR contains two completely independent implementations of the store concept:  one based on generics, and one using primitive arrays. So eventually, we won't have both Store<T> and StoreCommon.

The StoreCommon class can be used as the base class for a column-oriented store class that uses ints as indexes. Currently I don't think there are any methods that we would want to make abstract. The best candidate would be an abstract `add` method, but the problem with that is that the add methods in different subclasses will have different method signatures depending on the type of data being stored. For example, compare `NodeStore::add` to `LeafStore::add`.

StoreCommon contains just the logic to keep track of what indexes are in use or not. It's similar to the other comment about making ReferenceCounter a separate class: instead of subclassing StoreCommon, another approach could be to create a storage class that keeps an instance of StoreCommon for tracking indexes. And similar to that case, I don't have a strong opoinion about which approach is better.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
wnbts: the question is O(1) lookup only for error case checks? which are not expected to happen because if it were to happen, the code would not be functional

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
wnbts: since there is the structure for helping maintaining index integrity, it might as well be used. also since the cases this structure protects are also unlikely in correct implementation, i am also ok with discarding it.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
wnbts: not an issue. the refactoring can be done later.

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
wnbts: ok. this be a TODO.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactLeaf.java
wnbts: is the short parent index already a restriction? might as well make them consistent.

Java/core/src/main/java/com/amazon/randomcutforest/store/StoreCommon.java
theadroit-m: if this class is only keeping track of what indexes are in use, should we rename it?

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
theadroit-m: If the motivation for the code change is to make the tree more compact, we should use an array as we know the maximum size. Else ArrayList would try to grow the size as it reaches the limit even though we would never use the additional storage.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/RCStore.java
theadroit-m: Keeping the details of reference counting in a separate class would be better conceptually. But not a blocker as of now.

Java/core/src/main/java/com/amazon/randomcutforest/store/StoreCommon.java
jotok: In the latest update, I changed the base classes to IndexManager and SmallIndexManager. I opted to keep have the store classes inherit from them (as opposed to having the store classes create an instance of IndexManager), because it makes sense for those classes to inherit the size() and getCapacity() methods.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
```diff
@@ -0,0 +1,133 @@
...
+     * 
+     * @param capacity The maximum number of objects that can be added to this
+     *                 store.
+     */
+    public Store(Class<T> clazz, int capacity) {
+        checkArgument(capacity > 0, "capacity must be greater than 0");
+        this.capacity = capacity;
+
+        @SuppressWarnings("unchecked")
+        final T[] temp = (T[]) Array.newInstance(clazz, capacity);
```
wnbts, line 55: question. is temp variable required?

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return the maximum number of nodes whose data can be stored.
+     */
+    public short getCapacity() {
+        return capacity;
+    }
+
+    /**
+     * @return the number of nodes whose data is currently stored.
+     */
+    public int size() {
```
wnbts, line 61: minor. size should be of short type.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactLeaf.java
jotok: The reason for the discrepancy is that the current design uses one point store for the forest, but it uses one node store and one leaf store for each tree in the forest.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
```diff
@@ -0,0 +1,133 @@
...
+    private final int capacity;
+    private final T[] store;
+    private final int[] freeBlockStack;
+    private int freeBlockPointer;
+    private final BitSet occupied;
+
+    /**
+     * Create a new store with the given capacity.
+     * 
+     * @param capacity The maximum number of objects that can be added to this
```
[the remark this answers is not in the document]
jotok: I've decided to remove the generic classes from the PR since there seems to be agreement that we should use the column-oriented implementation.

Java/core/src/main/java/com/amazon/randomcutforest/genericstore/Store.java
jotok: This class will be removed in the next revision.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
jotok: Fixed.

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
sudiptoguha: mass can be a short (sample size of a tree)

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
sudiptoguha: Dimensions can be a short for axis parallel cuts.

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
sudiptoguha: mass can be short.

PR #54, opened 2020-09-10
Add new generic interfaces to support different tree implementations.

jotok:
This PR adds new generic interfaces with the goal of support the new compact implementation of Random Cut Tree. 

New interfaces:
* ITree<P> - Defines basic tree operations, like `addPoint`, `deletePoint`, and `traverseTree`.
* IUpdatabaleTree<P> - Extends ITree to add an `update` method. This interface plays a similar role to the existing TreeUpdater class.
* IStreamSampler<P> - A basic sampler interface that assumes that the input values will have a sequence index assigned.
* IUpdateCoordinator<P> - In the new forest traversal executor class, IUpdateCoordinator is used to convert a `double[]` into the point representation expected by an `IUpdatableTree`.

New classes:
* Sequential<P> - A container class that contains a value of type P and a sequence index.
* Weighted<P> - A container class that extends Sequential value. This is used to create a generic time-decay sampler.
* SimpleStreamSamplerV2 - A reimplementation of SimpleStreamSampler that implements the IStreamSampler interface
* GenericForestTraversalExecutor<P> - A modified version of AbstractForestTraveralExecutor. Each GenericForestTraversalExecutor gets an IUpdateCoordinator and a list of IUpdatableTree instances. The IUpdateCoordinator is used to initiate an udpate by converting a `double[]` into the point type expected by the IUpdatableTrees. The update step returns a list of points that were deleted from individual trees (if any), and passes them to back to the IUpdateCoordinator to complete the update. When compact trees are added, we will create an implementation of IUpdateCoordinator to manage the shared point store.
* SequentialForestTraversalExecutorV2<P>, ParallelForestTraversalExecutorV2<P> - Reimplementations of the existing concrete executor classes that extend from the new generic executor base class.
* GenericForestTraversalExecutorBenchmark - I created this new benchmark class just to have a code example of how the new generic executors would be created.

The classes with V2 in the name were created because I didn't want to make any changes to RandomCutForest in this PR (because such a change would require updating the JSON serializer, and I'd like to have consensus on our strategy before making that change). With that in mind, I can think of 3 options:

* Keep the V2 classes (possibly with renaming). We either create a RandomCutForestV2 class, or else we add a new instance variable of type GenericForestTraversalExecutor<?> to the class (so that it would have both and old-style and a new-style executor instance). The reason for taking the second approach is that existing JSON deserializer might still work by creating a forest that instantiates an old-style executor and leaves the new-style executor null.
* Replace the existing classes with their V2 counterparts and replace the AbstractForestTraversalExecutor in RandomCutForest with a GenericForestTraversalExecutor<?>. Then update the JSON deserializer logic to detect if a JSON file was created from the older version and deserialize it into the new version.
* Replace the existing classes with their V2 counterparts and replace the AbstractForestTraversalExecutor in RandomCutForest with a GenericForestTraversalExecutor<?>. Update the library version number to 2. Then the problem of serialization versions is external to the library (although we can provide tooling). Elasticsearch can update to version 2 on their own schedule.


sudiptoguha:
StreamSamplerv2 -> CompactSampler ??

jotok:
> StreamSamplerv2 -> CompactSampler ??

The current implementation is not any more compact than the original implementation though, it just supports a generic type parameter (so that it can be used to store both a weighted double array or a weighted index value).

sudiptoguha:
The executors are not part of any API and are internal - can we simply update the older ones?

jotok:
I've added an update that makes the following changes:

* Removes traversal methods from ITree
* Removes the IUpdatableTree interface
* Adds ITraversable and IUpdatable interfaces
* Adds an initial implementation of CompactRandomCutForest
* Adds an IPointStore interface that contains read-only methods. PointStore is now an implementation of IPointStore.

Java/core/src/main/java/com/amazon/randomcutforest/ITraversable.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest;
+
+public interface ITraversable {
```
wnbts, line 18: Question. It seems strange this class is not generic.

Java/core/src/main/java/com/amazon/randomcutforest/IUpdatable.java
```diff
@@ -0,0 +1,36 @@
...
+     * A given call to `update` may choose not to do anything with the point (and
+     * therefore the model state may not change). For example, in a
+     * {@link SamplingTree} the tree is only modified if the point is accepted by
+     * the sampler.
+     *
+     * @param point The point submitted to the model
+     * @return the point that was removed from the model as part of the update, or
+     *         null if no point was removed.
+     */
+    P update(P point);
```
wnbts, line 35: style. i would use Optional<T> to make possible absence clear.

Java/core/src/main/java/com/amazon/randomcutforest/IUpdateCoordinator.java
```diff
@@ -0,0 +1,26 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest;
+
+import java.util.List;
+
+public interface IUpdateCoordinator<P> {
```
wnbts, line 20: minor. this class can use some documentation, like what the methods do&return, or the expected call order.

Java/core/src/main/java/com/amazon/randomcutforest/AbstractForestUpdateExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Update the forest with the given point. The point is submitted to each
+     * sampler in the forest. If the sampler accepts the point, the point is
+     * submitted to the update method in the corresponding Random Cut Tree.
+     *
+     * @param point The point used to update the forest.
+     */
+    public void update(double[] point) {
```
wnbts, line 43: question. if a client is using float[] for compact rcf, should it also take float[] input?

Java/core/src/main/java/com/amazon/randomcutforest/PointStoreCoordinator.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.amazon.randomcutforest;
+
+import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
+
+import java.util.List;
+
+import com.amazon.randomcutforest.store.PointStore;
+
+public class PointStoreCoordinator implements IUpdateCoordinator<Sequential<Integer>> {
```
wnbts, line 24: minor. is it a good idea to have a pair of sequential and updateInput so the updateInput doesn't need to be stored as a state, different from sequential?

Java/core/src/main/java/com/amazon/randomcutforest/Sequential.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (!(o instanceof Sequential)) {
+            return false;
+        }
+
+        Sequential<?> other = (Sequential<?>) o;
+        return other.value.equals(value) && other.sequenceIndex == sequenceIndex;
```
wnbts, line 73: note. if the value is array type, this will be a same check.

wnbts (commented):
gone through 20 files, no major issues.

Java/core/src/main/java/com/amazon/randomcutforest/AbstractForestUpdateExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public abstract class AbstractForestUpdateExecutor<P> {
```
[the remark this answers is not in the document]
jotok: Agreed. There are some documentation misses in the PR as I hurried a little to have something to share. I'll add some comments.

Java/core/src/main/java/com/amazon/randomcutforest/ITraversable.java
jotok: We want the traversal methods to do the right thing regardless of the internal representation of a tree. E.g., if you're writing a new Visitor implementation, you shouldn't care if the tree stores points internally or looks them up on the fly during traversal (which is what the compact tree does).

Java/core/src/main/java/com/amazon/randomcutforest/IUpdatable.java
jotok: Will fix on revision.

Java/core/src/main/java/com/amazon/randomcutforest/IUpdateCoordinator.java
jotok: Will fix on revision.

Java/core/src/main/java/com/amazon/randomcutforest/AbstractForestUpdateExecutor.java
jotok: The repository is still in an intermediate state, with some classes using double[] and others (mostly the new ones) using float[]. Now that we've agreed to release this branch as a 2.0 version, let's address this inconsistency in a separate, targeted PR.

Java/core/src/main/java/com/amazon/randomcutforest/PointStoreCoordinator.java
jotok: Sorry, I don't understand the suggestion.

Java/core/src/main/java/com/amazon/randomcutforest/Sequential.java
jotok: I think I added this equals implementation for testing but didn't end up using it for exactly that reason. I'll try removing it from the next revision so we don't give the incorrect impression that we've implemented semantic equality.

Java/benchmark/src/main/java/com/amazon/randomcutforest/GenericForestTraversalExecutorBenchmark.java
```diff
@@ -0,0 +1,151 @@
...
+                    SimpleStreamSamplerV2<Integer> sampler = new SimpleStreamSamplerV2<>(Integer.class, sampleSize,
+                            lambda, random.nextLong());
+                    SamplingTree<Integer> samplingTree = new SamplingTree<>(sampler, tree);
+                    trees.add(samplingTree);
+                }
+
+                if (parallelExecutionEnabled) {
+                    executor = new ParallelForestUpdateExecutor<>(updateCoordinator, trees, threadPoolSize);
+                } else {
+                    executor = new ParallelForestUpdateExecutor<>(updateCoordinator, trees, threadPoolSize);
```
jotok, line 110: This should be sequential.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,538 @@
...
+            //
+
+            if (leafNodes.mass[leafOffset] > 1) {
+                --leafNodes.mass[leafOffset];
+                return leafNodes.pointIndex[leafOffset];
+            }
+
+            short parent = leafNodes.parentIndex[leafOffset];
+            int saved = leafNodes.pointIndex[leafOffset];
+            if (parent == -1) {
```
wnbts, line 240: minor. this check can also have a named method isEmpty/None for readability.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,538 @@
...
+ * The offsets are encoded as follows: an offset greater or equal maxSize
+ * corresponds to a leaf node of offset (offset - maxSize) otherwise the offset
+ * corresponds to an internal node
+ *
+ * The main use of this class is to be updated with points sampled from a
+ * stream, and to define traversal methods. Users can then implement a
+ * {@link Visitor} which can be submitted to a traversal method in order to
+ * compute a statistic from the tree.
+ */
+public class CompactRandomCutTree implements ITree<Sequential<Integer>> {
```
wnbts, line 48: remark. halfway through tree implementation. some general issues other than comments. there seems to be some data type inconsistencies such as float[]/double[], int/short. is the plan to later iron them out eliminating data type casting?

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,538 @@
...
+    private final int maxSize;
+    protected final NodeStore internalNodes;
+    protected final LeafStore leafNodes;
+    protected final IPointStore pointStore;
+    protected short rootIndex;
+
+    public CompactRandomCutTree(int maxSize, long seed, IPointStore pointStore) {
+        checkArgument(maxSize > 0, "maxSize must be greater than 0");
+        checkNotNull(pointStore, "pointStore must not be null");
+        this.maxSize = maxSize;
```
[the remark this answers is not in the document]
sudiptoguha: Yes. The number of internal nodes is at most 15, so only [0,14] should be used.

Java/core/src/main/java/com/amazon/randomcutforest/IUpdatable.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Optional;
+
+import com.amazon.randomcutforest.tree.SamplingTree;
+
+public interface IUpdatable<P> {
+    /**
+     * Update the model with the given point. If a point is removed from the model
+     * as part of the update, then it is returned. Otherwise the return value is
+     * null.
```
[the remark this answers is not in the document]
jotok: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/PointStoreCoordinator.java
jotok: I think I see your point. In the next revision I've changed the method signature of the `completeUpdate` method to:

```
public abstract void completeUpdate(P updateInput, List<P> deletedPoints);
```

The new implementation of `update` looks like this:

```
public void update(double[] point) {
        double[] pointCopy = cleanCopy(point);
        P updateInput = updateCoordinator.initUpdate(pointCopy);
        List<P> results = update(updateInput);
        updateCoordinator.completeUpdate(updateInput, results);
}
```

So now implementations of IUpdateCoordinator don't need to keep an instance variable to store any part of the updateInput value.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * @return the list of sequential points currently making up the sample.
+     */
+    List<Sequential<P>> getSamples();
+
+    /**
+     * @return the point that was evicted from the sample in the most recent call to
+     *         {@link #sample}, or null if no point was evicted.
+     */
+    Sequential<P> getEvictedPoint();
```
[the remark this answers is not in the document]
jotok: Changed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,538 @@
...
+            } else if (internalNodes.rightIndex[parent] == nodeOffset) {
+                return internalNodes.leftIndex[parent];
+            } else {
+                throw new IllegalArgumentException("node parent does not link back to node");
+            }
+        } else
+            return -1; // root is leaf
+    }
+
+    public boolean leftOf(float[] point, short nodeOffset) {
```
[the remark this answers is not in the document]
jotok: I don't think it needs to be. I'll change it to protected so that we can add tests for it later.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,538 @@
...
+        return (point[internalNodes.cutDimension[nodeOffset]] <= internalNodes.cutValue[nodeOffset]);
+    }
+
+    /**
+     * Delete the given point from this tree.
+     *
+     * @param sequential An offset of the point in the tree that we wish to delete.
+     */
+    /*
+     * public void deletePoint(WeightedPoint weightedPoint) { checkState(root !=
```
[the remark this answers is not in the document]
jotok: I deleted the commented-out code.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    /**
+     * This function deletes the point from the tree recursively. We traverse the
+     * tree based on the cut stored in each interior node until we reach a leaf
+     * node. We then delete the leaf node if the mass of the node is 1, otherwise we
+     * reduce the mass by 1.
+     *
+     * @param nodeOffset node that we are visiting in the tree.
+     * @param point      the point that is being deleted from the tree.
```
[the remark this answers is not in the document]
jotok: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
jotok: I split the difference and defined a new constant
```
public final static short NULL = -1;
```
So now this looks like
```
if (parent == NULL)
```
I felt that was clearer than `isNone(parent)`, but let me know what you think.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
jotok: Yes, exactly. When I created this PR we still hadn't settled the 1.0/2.0 question, and I didn't want to change existing classes like Node and BoundingBox. Given the size of this PR, I'd like to leave the inconsistencies for now and fix them in a later PR (possibly the next one).

Java/core/src/main/java/com/amazon/randomcutforest/PointStoreCoordinator.java
```diff
@@ -0,0 +1,55 @@
...
+        return new Sequential<>(pointIndex, currentIndex);
+    }
+
+    @Override
+    public void completeUpdate(Sequential<Integer> updateInput, List<Sequential<Integer>> updateResults) {
+        updateResults.stream().map(Sequential::getValue).forEach(deletedIndex -> {
+            store.incrementRefCount(updateInput.getValue());
+            store.decrementRefCount(deletedIndex);
+        });
+        store.decrementRefCount(updateInput.getValue());
```
wnbts, line 47: question. why is the input decremented?

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,543 @@
...
+     * we have created earlier.</li>
+     * <li>If the cut falls inside the existing box. We follow the cut of the
+     * existing box and move to right or left child of the current node based on the
+     * existing cut.</li>
+     * </ol>
+     *
+     * @param nodeOffset the current node in the tree we are on
+     * @param point      the point that we want to add to the tree
+     * @param pointIndex is the location of the point in pointstore
+     */
```
wnbts, line 328: minor. the return of the method is not documented.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTree.java
```diff
@@ -0,0 +1,538 @@
...
+     *
+     * @param oldNodeOffset The node we are replacing.
+     * @param newNodeOffset The new node we are inserting into the tree.
+     */
+
+    void replaceNode(short oldNodeOffset, short newNodeOffset) {
+
+        short parent;
+        if (isLeaf(oldNodeOffset)) {
+            parent = leafNodes.parentIndex[oldNodeOffset - maxSize];
```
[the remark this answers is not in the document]
wnbts: some comments were hidden, probably there is too many

Java/core/src/main/java/com/amazon/randomcutforest/PointStoreCoordinator.java
wnbts: answer it myself. to counter the initial add at line 37.

wnbts (approved):

PR #64, opened 2020-10-19
Add new interfaces and classes for mapping between models and states

jotok:
This PR is a proposal for handling state serialization in the library.

* For every class in the library whose instance variables encode part of the state of the Random Cut Forest model, we will create a corresponding "state" class. Every state class is  POJO with getter and setter methods and a no-argument constructor. State classes do not define any behavior by design. The Random Cut Forest state will be an object tree of state classes. This should make it possible for Java serialization libraries that rely on reflection to serialize the forest state with minimal effort.
* State classes for serialization should not be generic.
* There is an interface `IStateMapper<Model, State>` to encapsulate the logic for transforming a Model class to a State class.
* Where appropriate I've added new constructors for model classes for creating a new Model instance from a saved State. These constructors do not take an instance of the State class (because that logic is owned by State Mapper object). Indeed, the constructors should define minimal logic and should not have to know anything about how the State class is created.
* In addition to `IStateMapper` there is a second interface `IConextualStateMapper`. The method for constructing a model from a state in `IContextualStateMapper` takes a second argument that I'm calling the Context State. Consider the problem of deserializing a sampler. The sampler requires a lambda parameter. In a typical forest, the same lambda parameter is used in all samplers, and it would be redundant to include this value in the serialized sampler state. The Context State represents these kinds of parameters, which may be shared by other model classes.
* Methods for creating a Model from a State take an optional `long` value for a random seed. Some of our model classes, like samplers and trees, require a random number generator, but it doesn't make sense to include the random seed in the serialized State. This optional field is a convenience for such classes.


jotok:
The main thing I’d like to accomplish in this PR is to get agreement on the basic approach establish naming conventions:
* I use Model to refer to domain objects and State to refer to serializable data: so RandomCutTree and RandomCutTreeState. * Currently some of the classes use Data instead of state (like PointStoreDoubleData). I think both State and Data are fine, let’s pick one and use it consistently throughout.
* I’ve created an interface `IStateMapper<Model, State>` which defines methods for mapping from state to model and vice versa. There is also `IContextualStateMapper<Model, State, ContextState>` to handle classes where not all the fields needed to reconstitute the class are contained in the state class. In samplers, for example, we don’t include the lambda decay value in the serialized state class (because the same value would be repeated for every sampler. When deserializing a sampler, this value is passed in via the ContextState (which is just the forest state in the current implementation).
* I’ve added lombok to the library, and all serializable state classes are `@Data` classes with no methods.
* State classes are not generic, since generics can be a problem for serialization libraries. The only current exception is that I’ve created a generic class `SimpleStreamSamplerState<P>` to avoid code duplication. I extend it to define a concrete type, and the idea is that we’ll only use the concrete type when building up the forest state. Example: class `ArraySamplerState extends SimpleStreamSamplerState<double[]> {}` (that’s the entire class definition).
* I think a State class should match the Model that it’s representing. Currently we don’t serialize pointer-based RandomCutTrees, we only serialize the sampler state and then reinitialize the trees. Even so, I think the forest state should contain a SamplerPlusTreeState class to contain the sampler state. That will allow us to change the way serialization works in the future (e.g., by adding tree state to the class) in a backward compatible way.
* All these state classes are cluttering up the library. I think we should make a new `state` package to contain them.

sudiptoguha (commented):
Overall makes sense. We should get the double[] version down first. The change from "...Data" too "...State" seems to be the correct direction.

The ArrayState/SequentialState may be avoided if we decide that for both the implementations we have, we use the compact representation (TreeState can be Null/optional  for the pointer based version). This may reduce the proliferation of state classes.

Should we make addPoint() in the trees take in reference, weight and sequencenum explicit? That may help us eliminate the Sequential<P>, Weighted<P> from all places except the SimpleStreamSampler? The V 1.0 used WeightedPoint, so the compatibility is already broken.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -1004,42 +1062,107 @@ public RandomCutForest build() {
...
+                    componentModelStates.add(mapper.toState(sampler));
+                }
+            }
+
+            return state;
+        }
+
+        @Override
+        public RandomCutForest toModel(RandomCutForestState state) {
+            return null;
```
theadroit-m, line 1149: Can we add a todo or maybe throw UnsupportedOperationException?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -18,52 +18,55 @@
...
+import com.amazon.randomcutforest.tree.CompactRandomCutTreeState;
 
 /**
  * A class that encapsulates most of the data used in a RandomCutForest such
  * that the forest can be serialized and deserialized.
  */
-public class ForestState {
-
-    public String typeOfForest = null;
+@Data
```
theadroit-m, line 33: np: you can also use the @Builder it makes creating these objects easier to read

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -18,52 +18,55 @@
...
-    public ArrayList<CompactSamplerData> compactSamplerData;
+    public ArrayList<CompactSamplerState> compactSamplerData;
 
     public ArrayList<List<Sequential<double[]>>> sequentialSamplerData;
 
     public ArrayList<List<Weighted<double[]>>> smallSamplerData;
 
-    public ArrayList<TreeData> treeData;
-
+    public ArrayList<CompactRandomCutTreeState> treeData;
```
theadroit-m, line 71: Can we make all of these private as well?

Java/core/src/main/java/com/amazon/randomcutforest/IStateMapper.java
```diff
@@ -0,0 +1,28 @@
...
+public interface IStateMapper<Model, State> {
+    State toState(Model model);
+
+    default Model toModel(State state) {
+        return toModel(state, 0L);
+    }
+
+    default Model toModel(State state, long seed) {
+        return toModel(state);
+    }
```
theadroit-m, line 27: Question: Since java supports function overloading, having #to() functions makes the code more readable:

```
public interface IStateMapper<Model, State> {
    State to(Model model);

    default Model from(State state) {
        return toModel(state, 0L);
    }

    default Model from(State state, long seed) {
        return toModel(state);
    }
}
```

theadroit-m (commented):
Looks good. thanks for raising this PR.

Did one pass, will go over it again later today.

Java/core/src/main/java/com/amazon/randomcutforest/IContextualStateMapper.java
```diff
@@ -0,0 +1,28 @@
...
+public interface IContextualStateMapper<Model, State, ContextState> {
+    State toState(Model model);
+
+    default Model toModel(State state, ContextState contextState) {
+        return toModel(state, contextState, 0L);
+    }
+
+    default Model toModel(State state, ContextState contextState, long seed) {
+        return toModel(state, contextState);
+    }
```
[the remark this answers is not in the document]
jotok: I liked that you only had to define the one that you wanted, but point taken. I'll make the version that takes a random seed the required version, and mappers that don't need that argument can simply ignore it.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -187,33 +215,36 @@ protected RandomCutForest(Builder<?> builder) {
...
+                components.add(samplerPlusTree);
             } else {
                 CompactRandomCutTreeDouble tree = new CompactRandomCutTreeDouble(sampleSize, rng.nextLong(),
                         pointStore);
                 CompactSampler newSampler = new CompactSampler(sampleSize, lambda, rng.nextLong(),
                         storeSequenceIndexesEnabled);
-                components.add(new SamplerPlusTree(newSampler, tree));
+                components.add(new SamplerPlusTree<>(newSampler, tree));
             }
         }
```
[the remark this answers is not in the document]
jotok: There's going to be some cleanup in a later PR. The rough plan is to separate the part of the constructor that sets the metadata/config variables from the part that sets up the model classes. That way we can add a new constructor that can be initialized with models restored from serialized state. Something like this:

```
public RandomCutForest(Builder builder) {
  initConfig(builder);
  if (compactEnabled) {
    initCompact();
  } else {
    initNonCompact();
  } 
}

public <P> RandomCutForest(Builder builder, IUpdateCoordinator<P> updateCoordinator, List<IComponentModel<P>> componentModels) {
  initConfig(builder);
  this.updateCoordinator = updateCoordinator;
  this.componentModels = componentModels;
}
```

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -222,12 +253,39 @@ protected RandomCutForest(Builder<?> builder) {
...
+                .threadPoolSize(state.getThreadPoolSize()).parallelExecutionEnabled(state.isParallelExecutionEnabled())
+                .build();
+
+        if (!state.isCompactEnabled()) {
+            forest.updateExecutor.initializeModels(state.smallSamplerData, state.sequentialSamplerData);
+        } else {
+            forest.pointStore.reInitialize(state.pointStoreDoubleData);
+            if (state.saveTreeData) {
+                checkState(state.treeData != null, " error, missing tree information");
+            }
```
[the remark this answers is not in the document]
jotok: That's a good callout. Note that this particular function will be replaced in a future CR with the StateMapper implementation for RandomCutForest. We can make this a parameter of the constructor for that StateMapper class.

Java/core/src/main/java/com/amazon/randomcutforest/CommonUtils.java
```diff
@@ -63,8 +63,9 @@ public static void checkState(boolean condition, String message) {
      * @param message the error message that will be used if the condition is not
      *                met
      */
-    public static void checkNotNull(Object object, String message) {
+    public static <T> T checkNotNull(T object, String message) {
```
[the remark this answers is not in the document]
jotok: Yes, but let's do that in a separate, less-sprawling PR :)

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: Done.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
jotok: That's the end goal. I'm pretty sure I left this because existing code that I want to replace with the mapper classes expects it to be public and I didn't want to increase the scope of this PR. We can make this change after we remove the old code.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
jotok: I like using builders in place of constructors with long argument lists, but right now all the state classes have 0-argument constructors and are populated using setters. In this case, I don't think using builders would be a big readability gain, but I'm not opposed to adding them.

Java/core/src/main/java/com/amazon/randomcutforest/IStateMapper.java
jotok: The `to` method doesn't read very well to me. I think it's because the object of "to" is CompactRandomCutTreeState, which is all the way to the left. If you read this line in your head, you say "to tree", which is the opposite of what we're doing.

```
CompactRandomCutTreeState state = mapper.to(tree);
```

The `from` method reads well, though. Would it be weird to use `from` in both directions?

```
CompactRandomCutTreeState state = mapper.from(tree);
CompactRandomCutTree tree = mapper.from(state);
```

jotok:
> Should we make addPoint() in the trees take in reference, weight and sequencenum explicit? That may help us eliminate the Sequential
> 
> , Weighted
> 
> from all places except the SimpleStreamSampler? The V 1.0 used WeightedPoint, so the compatibility is already broken.

Weights are an implementation detail of the sampler, right? Is there ever a reason for a tree to see the weight value

jotok:
Updates:
* All state classes are renamed to end in State
* All state classes except RandomCutForestState have been moved to a new state package. RandomCutForestState will be moved in a future PR when we complete the state mapper implementation for RandomCutForest.
* Both compact and noncompact forests are now serialized into a PointStoreDoubleData instance and a list of CompactRandomCutTreeState instances. I've removed a number of state classes that are no longer required after this change.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public CompactSamplerState toState(CompactSampler model) {
+        CompactSamplerState state = new CompactSamplerState();
+        state.setSize(model.size());
+        state.setCapacity(model.getCapacity());
+        state.setWeightArray(Arrays.copyOf(model.getWeightArray(), model.size()));
+        state.setReferenceArray(Arrays.copyOf(model.getReferenceArray(), model.size()));
+        if (model.getSequenceArray() != null) {
```
wnbts, line 42: question. should storeSequences be used?

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerState.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Arrays;
+
+import lombok.Data;
+
+import com.amazon.randomcutforest.sampler.CompactSampler;
+
+@Data
+public class CompactSamplerState {
+
+    public float[] weightArray;
```
wnbts, line 27: minor. the fields can be private with lombok.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerState.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public int size;
+
+    public int capacity;
+
+    public CompactSamplerState() {
+
+    }
+
+    public CompactSamplerState(CompactSampler sampler) {
```
wnbts, line 41: remark. it seems like storeSequence should be used here storing model into state.

wnbts (commented):
i've gone through 30 files with a dozen or so to go.

some of the things i am thinking about might be off topic.
1. is model and state be exactly the same? it does not always appear so. A model might contain data that is not in the state. Trees are an example.
2. Within model, there is optional data. And from model to state, there is also optional data. So for clients, there are two things to do. First, they need to choose the data a model `uses`, which is required at constructor time. During serialization, they should be able to choose the data to `save` at runtime. Different decisions can be made at different times. For example, for a model configured to use trees (as opposed to no trees), it can be serialized twice. Once with trees and another without trees. It might not be the best example but reflects the overall idea of keeping forest config to forest as a model. Other options (executor, serialization-related) do not need to be one-way-door (at construction time) and can move to elsewhere later. Mixing all config together could add burden on the forest interface and client.
3. does sequence number need to be stored? if not, i am for removing it for simplifying the interface as there are many options already.

jotok:
>     1. is model and state be exactly the same? it does not always appear so. A model might contain data that is not in the state. Trees are an example.

Currently a model is a data structure with some attached behaviors (methods) whose data representation will strike some balance between memory use and runtime performance, and a state records enough data to restore a state (possibly with extra contextual state) but aims to be space-efficient and does not define any behaviors. 
 
>     2. Within model, there is optional data. And from model to state, there is also optional data. So for clients, there are two things to do. First, they need to choose the data a model `uses`, which is required at constructor time. During serialization, they should be able to choose the data to `save` at runtime. Different decisions can be made at different times. For example, for a model configured to use trees (as opposed to no trees), it can be serialized twice. Once with trees and another without trees. It might not be the best example but reflects the overall idea of keeping forest config to forest as a model. Other options (executor, serialization-related) do not need to be one-way-door (at construction time) and can move to elsewhere later. Mixing all config together could add burden on the forest interface and client.

I assume you're referring in part to the `saveTreeData` flag in the forest. I agree that this shouldn't be a configuration option for the forest itself. It would make sense to add this as a constructor argument for the relevant mapper class:

```
public RandomCutForestMapper(boolean saveTreeData)
```

I think I agree on the point about executors, too. We're already planning to remove them from the serialization path. In the new model, forest configuration will consist of the config/metadata variables, an IUpdateCoordinator (shared state), and a list of IComponentModels. Executors will be constructed to act on models, but will not own any state themselves. With this change it should be possible to make changes to executors at runtime (e.g., switch from serial to parallel).

>     3. does sequence number need to be stored? if not, i am for removing it for simplifying the interface as there are many options already.

The idea of sequence indexes, which we haven't realized in this library, is that they will be useful for understanding the state of the model and interpreting things like anomaly scores. For example, if you expect a particular point to be scored as an anomaly, but it's not, you could use the near neighbors method to find similar points and you could see the sequence indexes to know when they were added to the system. We should think through some use cases like this and write some examples to show that sequence indexes are helpful, and if we can't do that then I would agree with getting rid of them (because the fact that we made them optional as an optimization leaks through the entire library).

Java/benchmark/src/main/java/com/amazon/randomcutforest/ForestUpdateExecutorBenchmark.java
```diff
@@ -84,11 +100,12 @@ public void setUpExecutor() {
                     executor = new SequentialForestUpdateExecutor<>(updateCoordinator, trees);
                 }
             } else {
-                PointStore store = new PointStore(dimensions, numberOfTrees * sampleSize);
+                PointStoreDouble store = new PointStoreDouble(dimensions, numberOfTrees * sampleSize);
```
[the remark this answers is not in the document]
jotok: I prefer suffixes, so that if we have two versions of a class they appear next to each other alphabetically. That said, changing the basic point type from double[] to float[] is something that will have to be done carefully, and we may not include that in 2.0 work. In that case, we may delete the current PointStore and rename PointStoreDouble to PointStore.

Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
```diff
@@ -0,0 +1,22 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest;
+
+import com.amazon.randomcutforest.executor.ITraversable;
+import com.amazon.randomcutforest.executor.IUpdatable;
+
+public interface IComponentModel<P> extends ITraversable, IUpdatable<P> {
```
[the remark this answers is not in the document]
jotok: We want to leave open the possibility that a component model is itself a collection. The reason we still need separate ITraversable and IUpdatable interfaces is that a tree is traversable but not updatable.

Java/benchmark/src/main/java/com/amazon/randomcutforest/ForestUpdateExecutorBenchmark.java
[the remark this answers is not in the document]
sudiptoguha: I think the float[] will get included in 2.0, but we we should do so carefully. It's probably a few commits by itself; plus we may have to look at the tests, etc. It may be easier to get a clean double[] version first and then change double[] -> float[]. We may have to look at the precision of the cut carefully.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
 
     public boolean saveTreeData;
 
-    public PointStoreDoubleData pointStoreDoubleData;
-
-    public ArrayList<CompactSamplerData> compactSamplerData;
+    public PointStoreDoubleState pointStoreDoubleState;
 
-    public ArrayList<List<Sequential<double[]>>> sequentialSamplerData;
+    public List<CompactSamplerState> compactSamplerStates;
```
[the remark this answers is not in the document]
jotok: I'd like to leave this for now. The previous PR added methods for creating state classes and this one moves and rewrites some of that logic into the new mapper classes, but I didn't remove the old  code and I didn't complete the rewrite because the mapper for the RandomCutForest class doesn't implement the toModel() method yet. If I make this change, it will require additional changes in code that will be deleted in a future PR.

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
```diff
@@ -16,58 +16,43 @@
...
-    }
-
-    public PointStoreCoordinator(PointStore store) {
-        checkNotNull(store, "store must not be null");
-        this.doubleStore = null;
         this.store = store;
     }
 
     @Override
     public Integer initUpdate(double[] point, long seqNum) {
```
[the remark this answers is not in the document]
jotok: I'm open to this idea, but I prefer the current names. The important thing about these methods isn't just that they are called before and after `update` is called on each `IUpdatable`. `initUpdate` produces the value that's consumed in the `update` calls, and `completeUpdate` takes the result of the update calls. So `initUpdate` and `completeUpdate` are really part of the forest update.

That said, do you find the names confusing? What if we changed initUpdate to beginUpdate? Do you still prefer preUpdate and postUpdate?

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
```diff
@@ -16,58 +16,43 @@
...
     }
 
     @Override
-    public PointStoreDoubleData getPointStoreData() {
-        if (doubleStore != null) {
-            return new PointStoreDoubleData(doubleStore);
-        } else { // To be filled in
-            return null;
-        }
+    public PointStoreDoubleState getPointStoreData() {
```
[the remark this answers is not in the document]
jotok: Done. I missed renaming this method when I changed the name of the state class. Note that this method will eventually be removed because we can use the new mapper class to do this transform.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/ArraySamplersToCompactStateConverter.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public List<CompactSamplerState> getCompactSamplerStates() {
+        return compactSamplerStates;
+    }
+
+    public void addSampler(SimpleStreamSampler<double[]> sampler) {
+        CompactSamplerState compactState = new CompactSamplerState(sampler.size(), sampler.getCapacity(),
+                storeSequences);
+        List<Sequential<double[]>> samples = sampler.getSequentialSamples();
```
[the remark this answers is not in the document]
jotok: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/ArraySamplersToCompactStateConverter.java
```diff
@@ -0,0 +1,104 @@
...
+
+        int i = 0;
+        for (Sequential<double[]> sample : sampler.getSequentialSamples()) {
```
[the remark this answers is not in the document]
jotok: `getSequentialSamples()` returns a List and not an ArrayList specifically, so I didn't want to assume that `samples.get(i)` was an efficient operation. Would you still switch to a `for(;;)` loop in this scenario?

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/ArraySamplersToCompactStateConverter.java
```diff
@@ -0,0 +1,104 @@
...
+                compactState.sequenceArray[i] = sample.getSequenceIndex();
+            }
+
+            i++;
+        }
+
+        compactSamplerStates.add(compactState);
+    }
+
+    public void addSamples(ArrayList<List<Sequential<double[]>>> sampleList, int capacity) {
```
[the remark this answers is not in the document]
jotok: Agreed. This is another case where I'm leaving something that will be deleted later, after I've completed the end-to-end refactoring to use mapper classes.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/ArraySamplersToCompactStateConverter.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public PointStoreDoubleState getPointStoreDoubleState() {
+        return new PointStoreDoubleMapper().toState(pointStoreDouble);
+    }
+
+    public List<CompactSamplerState> getCompactSamplerStates() {
+        return compactSamplerStates;
+    }
+
+    public void addSampler(SimpleStreamSampler<double[]> sampler) {
```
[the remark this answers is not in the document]
jotok: The purpose of this class is to capture the state of a `List<SimpleStreamSampler<double[]>` into one `PointStoreDoubleState` and a `List<CompactSamplerState>`. So the order goes like this:

1. Create an `ArraySamplersToCompactStateConverter`
2. For each `SimplerStreamSampler<double[]>`, call `addSampler` to extract its state
3. Add the final `PointStoreDoubleState` and the `List<CompactSamplerState>` to the forest state.

One simplification could be to create a CompactForestState class that contains a PointStoreDoubleState and a List<CompactSamplerState>, and make it a top-level field in RandomCutForestState. I don't know if that would be an improvement.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
jotok: Yes, I think that's better.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerState.java
jotok: These fields are being used in code that will be deleted, and I don't want to add those changes to this PR. We'll have another chance to review all the data classes when we add them to the 2.0 branch.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerState.java
jotok: storeSequence really is a forest configuration option, because if it's enabled then sequences are propagated through a bunch of different places in the library. I think it's different than a saveTreeData flag as part of serialization, because in that case there's a tradeoff between between being able to restore the exact tree structure (at the cost of a larger serialized object), versus creating a smaller serialized object and then getting back a tree that's equal in distribution only. Throwing away sequence indexes during serialization would be a lossy operation.

Java/core/src/main/java/com/amazon/randomcutforest/IStateMapper.java
sudiptoguha: Speaking for me, a "from" in both directions make sense -- but what if there are more than one target? It's hypothetical though, in the context of this library. Why not use mapper.toRandomCutTreeState ?

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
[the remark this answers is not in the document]
sudiptoguha: I am a skeptical about "postUpdate" -- the "post processing" is a part of the update because otherwise the state will not remain meaningful. Its "postParallel" but not postUpdate ...

Java/core/src/main/java/com/amazon/randomcutforest/IStateMapper.java
jotok: These methods are defined by an interface with generic type parameters, so the method names are also "generic". Hence `toState` and `toModel`.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/LeafStoreMapper.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public LeafStore toModel(LeafStoreState state, long seed) {
+        int capacity = state.getPointIndex().length;
+        int[] pointIndex = Arrays.copyOf(state.getPointIndex(), capacity);
+        short[] parentIndex = Arrays.copyOf(state.getParentIndex(), capacity);
+        int[] mass = Arrays.copyOf(state.getMass(), capacity);
+
+        short freeIndexPointer = (short) (state.getFreeIndexes().length - 1);
+        short[] freeIndexes = new short[capacity];
+        System.arraycopy(state.getFreeIndexes(), 0, freeIndexes, 0, freeIndexPointer + 1);
```
wnbts, line 33: minor. using `state.getFreeIndexes().length` for the last parameter might be clearer.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/LeafStoreMapper.java
```diff
@@ -0,0 +1,48 @@
...
+
+        return new LeafStore(pointIndex, parentIndex, mass, freeIndexes, freeIndexPointer);
+    }
+
+    @Override
+    public LeafStoreState toState(LeafStore model) {
```
wnbts, line 39: question. does freeIndexPointer need to be saved to the state?

Java/core/src/main/java/com/amazon/randomcutforest/state/store/LeafStoreState.java
```diff
@@ -13,22 +13,29 @@
...
+import com.amazon.randomcutforest.store.LeafStore;
 
-    public LeafStoreData(LeafStore leafStore) {
+@Data
+public class LeafStoreState {
+
+    public int[] pointIndex;
+    public short[] parentIndex;
+    public int[] mass;
+    public short[] freeIndexes;
```
wnbts, line 30: minor. this name seems inconsistent with previous variables. freeIndex?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
  * that the forest can be serialized and deserialized.
  */
-public class ForestState {
-
-    public String typeOfForest = null;
+@Data
+public class RandomCutForestState {
 
-    public long entreesSeen;
+    private long entreesSeen;
```
wnbts, line 35: is this variable part of forest state?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
+    private int sampleSize;
 
-    public int dimensions;
+    private int dimensions;
 
-    public int outputAfter;
+    private int outputAfter;
 
-    public boolean storeSequenceIndexesEnabled;
+    private boolean storeSequenceIndexesEnabled;
```
wnbts, line 47: raise this for final decision

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
 
-    public boolean compactEnabled;
+    private boolean compactEnabled;
 
-    public boolean centerOfMassEnabled;
+    private boolean centerOfMassEnabled;
 
-    public boolean parallelExecutionEnabled;
+    private boolean parallelExecutionEnabled;
```
wnbts, line 53: raise this for final decision

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
 
-    public int threadPoolSize;
+    private int threadPoolSize;
```
wnbts, line 55: raise this for final decision

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
 
     public boolean saveTreeData;
```
wnbts, line 57: raise this for final decision

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
 
-    public PointStoreDoubleData pointStoreDoubleData;
-
-    public ArrayList<CompactSamplerData> compactSamplerData;
+    public PointStoreDoubleState pointStoreDoubleState;
```
wnbts, line 59: suggestion. it might be clearer to add comments for variables exclusively for compact.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
```diff
@@ -15,55 +15,54 @@
...
 
-    public ArrayList<List<Sequential<double[]>>> sequentialSamplerData;
+    public List<CompactSamplerState> compactSamplerStates;
 
-    public ArrayList<List<Weighted<double[]>>> smallSamplerData;
+    public List<List<Sequential<double[]>>> sequentialSamplerData;
 
-    public ArrayList<TreeData> treeData;
+    public List<List<Weighted<double[]>>> smallSamplerData;
```
wnbts, line 65: question. are these two variables both for current rcf? the difference is whether there is sequence or not?

they might use a better name to be differentiated from other sample/state.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeDouble.java
```diff
@@ -584,18 +604,40 @@ private BoundingBox reflateNode(int offset) {
...
+    public void reInitialize(CompactRandomCutTreeState treeData) {
         rootIndex = treeData.rootIndex;
-        leafNodes.reInitialize(treeData.leafData);
-        internalNodes.reInitialize(treeData.nodeData);
+        leafNodes.reInitialize(treeData.leafStoreState);
+        internalNodes.reInitialize(treeData.nodeStoreState);
         reflateTree();
     }
+
+    public static class StateMapper implements
```
wnbts, line 622: this state mapper is nested

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
sudiptoguha: It is needed for to "restore the clock" used in weight computation of the samplers.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
sudiptoguha: One suggestion, via theadroit-m, was to have saveTreeData as an argument to getForestState (or equivalent name). That does appear cleaner and we can remove this.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
sudiptoguha: I think we are using the compact representation for everything --- for the original style trees, we cannot save the tree state (short of rebuilding the tree as a CompactRCF, which defeats any intended purpose of using the original trees),  which remains as a null list/array.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
sudiptoguha: I'd keep this for now. a) if does help some of the inspection/explanation, b) as you point out, ion this information is available and we are not saving tree data, we can use this information to rebuild trees.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
sudiptoguha: We discussed this - I think we can separate data and execution config. But for now going with a catchall "state" may be easier? The use case is that it saves the  effort of a person/program to remember/decide what are the execution configs, and simply reuse them in deserialization. If the information is saved, we can choose to not use it or use to separately. If it is not saved, then we have no such choice.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
sudiptoguha: Same comment as parallelExecutionEnabled.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForestState.java
jotok: With our new plan of converting a list of noncompact samplers into a pointstorestate and a list of compactsamplerstates, these should eventually be removed.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/LeafStoreMapper.java
jotok: Done.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/LeafStoreState.java
jotok: The current naming matches in instance variable names in the corresponding model class. `pointIndex`, `parentIndex`, and `mass` are all public fields in LeafStore, and freeIndexes is a private field in SmallIndexManager, the superclass of LeafStore.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeDouble.java
jotok: Oops, fixed.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -222,12 +257,39 @@ protected RandomCutForest(Builder<?> builder) {
...
-            updateExecutor = new SequentialForestUpdateExecutor(coordinator, components);
+            updateExecutor = new SequentialForestUpdateExecutor(updateCoordinator, components);
+        }
+    }
+
+    /**
+     * The following initialization must happen as soon as the forest is created and
+     * trees are empty
+     */
+    public static RandomCutForest getForest(RandomCutForestState state) {
```
wnbts, line 272: minor. duplicate method with line 1069.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -1004,42 +1066,117 @@ public RandomCutForest build() {
...
-        answer.lambda = getLambda();
-        answer.sampleSize = getSampleSize();
-        answer.centerOfMassEnabled = centerOfMassEnabled();
-        answer.outputAfter = getOutputAfter();
-        answer.parallelExecutionEnabled = parallelExecutionEnabled();
-        answer.threadPoolSize = getThreadPoolSize();
-        answer.storeSequenceIndexesEnabled = storeSequenceIndexesEnabled();
-        answer.entreesSeen = getTotalUpdates();
-        answer.compactEnabled = compactEnabled();
+    public static class StateMapper implements IStateMapper<RandomCutForest, RandomCutForestState> {
```
wnbts, line 1093: this mapper is nested.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -1004,42 +1066,117 @@ public RandomCutForest build() {
...
+            return state;
+        }
+
+        @Override
+        public RandomCutForest toModel(RandomCutForestState state, long seed) {
+            throw new UnsupportedOperationException();
+        }
+    }
+
+    public RandomCutForestState getForestState() {
```
wnbts, line 1167: question. is this a duplicate with Forest.StateMapper line 1093?

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeDouble.java
```diff
@@ -584,18 +602,23 @@ private BoundingBox reflateNode(int offset) {
...
+    public LeafStoreState getLeaves() {
+        return new LeafStoreState(leafNodes);
     }
 
-    public void reInitialize(TreeData treeData) {
+    public short getRootIndex() {
+        return rootIndex;
+    }
+
+    public void reInitialize(CompactRandomCutTreeState treeData) {
```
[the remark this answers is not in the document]
jotok: This method is being used in the old code path for restoring a compact tree, and will be removed when we remove that code path in a future PR.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: Removed unused duplicate method.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: I'd like to leave this for now becase we haven't completed migrating the forest state class to the new paradigm and I'd like to keep those changes together. If I move this to the top level that will bring additional changes into this PR because the implementation code currently uses inner class visibility rules. Alternatively, I could delete this partial implementation from this PR.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: Yes, eventually the new state mapper could should replace the getForestState(). Currently the state mapper code is incomplete, and getForestState() is being used in the serde tests.

sudiptoguha:
Merging. We can iterate on the deferred pieces and hopefully the PRs will be smaller!

PR #65, opened 2020-10-29
Modify the forest constructor to create a list of component models

jotok:
This is a first step toward removing state from executor classes.
Eventually all state should be owned by the forest, the update
coordinator, or the component models. This change modifies the forest
constructor to initialize component models, but those models are not yet
in the execution path.

* I've defined a new `ComponentList<P>` class which is really just a typedef for `ArrayList<IComponentModel<P>>`. The main reason for this is that Java has a hard time casting nested generics. In particular, casting `ComponentList<double[]>` to `ComponentList<?>` is fine, but casting `ArrayList<IComponentModel<double[]>>` to `ArrayList<IComponentModel<?>>` is a compile-time error.
* This PR adds new code to the forest constructor to configure an `IUpdateCoordinator` and a `ComponentList`. After initialization, the `ComponentList` is not used and the `IUpdateCoordinator` is overwritten in a subsequent block of code. This was intentional to reduce the scope of this PR. Currently the executor classes expect a list of `SamplerPlusTree`s. Updating them to operate on a `ComponentList` would have significantly increased the size of this PR, so I plan to do it in a follow-up.
* The new forest constructor takes arguments for an `IUpdateCoordinator` and a `ComponentList`. When we complete implementing the state mapper class for RandomCutforest, these arguments will be used to pass in instances that are reconstructed from the serialized state.


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -255,13 +246,75 @@ protected RandomCutForest(Builder<?> builder) {
...
+                    .centerOfMassEnabled(centerOfMassEnabled).randomSeed(rng.nextLong()).build();
+            IStreamSampler<double[]> sampler = new SimpleStreamSampler<>(sampleSize, lambda, rng.nextLong(),
+                    storeSequenceIndexesEnabled);
+            componentModels.add(new SamplerPlusTree<>(sampler, tree));
+        }
+        this.updateCoordinator = updateCoordinator;
+        this.componentModels = componentModels;
+    }
+
+    private RandomCutForest(Builder<?> builder) {
```
wnbts, line 285: question. why not keep this constructor as protected for the creation use case? then the other constructor will take three non-null parameters

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -179,62 +180,52 @@
...
-
-        // If a random seed was given, use it to create a new Random. Otherwise, call
-        // the 0-argument constructor
-        rng = builder.randomSeed.map(Random::new).orElseGet(Random::new);
+        // TODO the rest of this constructor will be updated in a future revision
 
         if (compactEnabled) {
             pointStore = new PointStoreDouble(dimensions, sampleSize * numberOfTrees + 1);
-            updateCoordinator = new PointStoreCoordinator(pointStore);
+            this.updateCoordinator = new PointStoreCoordinator(pointStore);
```
wnbts, line 221: question. if updateCoordinator is from input, should it be reset here?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -255,13 +246,75 @@ protected RandomCutForest(Builder<?> builder) {
...
         } else {
-            threadPoolSize = 0;
             traversalExecutor = new SequentialForestTraversalExecutor(components);
-            updateExecutor = new SequentialForestUpdateExecutor(updateCoordinator, components);
+            updateExecutor = new SequentialForestUpdateExecutor(this.updateCoordinator, components);
+        }
+    }
+
+    private void initCompact() {
+        PointStoreDouble pointStore = new PointStoreDouble(dimensions, sampleSize * numberOfTrees + 1);
```
wnbts, line 258: question. why +1 capacity?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -179,62 +180,52 @@
...
             pointStore = new PointStoreDouble(dimensions, sampleSize * numberOfTrees + 1);
-            updateCoordinator = new PointStoreCoordinator(pointStore);
+            this.updateCoordinator = new PointStoreCoordinator(pointStore);
         } else {
             pointStore = null;
-            updateCoordinator = new PointSequencer();
+            this.updateCoordinator = new PassThroughCoordinator();
         }
 
+        ArrayList<SamplerPlusTree> components = new ArrayList<>(numberOfTrees);
```
wnbts, line 227: issues? what happen to trees created in line 211, 213?

Java/core/src/main/java/com/amazon/randomcutforest/executor/IUpdateCoordinator.java
```diff
@@ -47,9 +44,18 @@
...
     void completeUpdate(List<Optional<UpdateReturn<P>>> updateResults, P updateInput);
 
-    PointStoreDoubleState getPointStoreState();
+    long getTotalUpdates();
+
+    // TODO we can remove this from the interface once this value is being set in
+    // the appropriate mapper
+    void setTotalUpdates(long totalUpdates);
 
+    default long getCurrentIndex() {
```
wnbts, line 58: question. does index need to be 1-based?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: This constructor exists mostly for code organization, and all it does is set the RandomCutForest config variables from the builder. In particular, it doesn't set up models or executors. Since it doesn't create models, it doesn't really fully initialize the RCF state. Originally I wanted this to be a private method named `initConfig`, but it has to be a constructor because some of the forest instance variables are final.

Taking a step back, we want to be able to construct a forest in 2 ways:
* Create a forest from a config, where the models are initialized in the construcrtor
* Create a forest from a config and models that were initialized outside of the constructor. This is what we'll use during deserialization.

The current approach in this PR is to expose one constructor that takes a builder and model arguments. Then the decision to initialize models or not is based on whether the model arguments are null. I agree that it would be better two expose two constructors, one that takes model arguments and one that doesn't, but then we either have to duplicate the configuration initialization code, or else we have to make all the instance variables non-final so that they can be set in a method. Which solution do you like? Do you have different suggestion?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: As described in another comment, I want to support two cases: when models are passed in and when they need to be created from scratch.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: I think this is to handle the case when the point store is completely full and we need to update the forest with a new point. The new point has to be added to the point store. If it's accepted into a tree, then at least one other point will be removed from the point store. If it's not accepted by any tree, than the submitted point will be removed from the point store at the end of the update operation.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: Currently nothing. In the next PR I'll update the executor classes so that they take in a ComponentList argument (currently they expect a List<SamplerPlusTree>), and that would made this PR bigger and more complicated.

Java/core/src/main/java/com/amazon/randomcutforest/executor/IUpdateCoordinator.java
jotok: I think we've gone back and forth on this in the 2.0 code. I don't have a strong opinion about it, but this PR does have an opinion about who should own counting updates and keeping track of the current index (namely, the update coordinator). If we're fine with being 0-based, I can remove this method entirely.

The reason I put a +1 there, is because this value used to be maintained in the update executor, and the first point submitted to the forest would get an index of 1 (https://github.com/aws/random-cut-forest-by-aws/blob/compact-rcf/Java/core/src/main/java/com/amazon/randomcutforest/executor/AbstractForestUpdateExecutor.java#L80).

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: Actually, these are lines that will be removed in the next update.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
wnbts: To work around that a private method cannot set final members, how about a private constructor C which takes a config and an extra parameter (type and value doesn't matter) to differentiate from the config constructor A? C does the shared initialization logic. Constructors A and B call C and handle the rest.

jotok: I like it, I'll include this change in the next revision.

Java/core/src/main/java/com/amazon/randomcutforest/executor/IUpdateCoordinator.java
jotok: Agreed, I'll change this.

wnbts (approved):

PR #69, opened 2020-11-20
Compact rcf

sudiptoguha:
*Description of changes:* This is the first of three PRs to clean up creation of temporary nodes for compactRCF. This change removes the access to sibling in the two visitor main classes and replaces that by a function to the sibling bounding box.

Next change would change the visitor classes to use iNodeView interface instead of the nodes directly.

The final change would implement the iNodeView for CompactRCF and remove the need for temporary nodes being created.


Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
```diff
@@ -422,4 +422,24 @@ protected void deleteSequenceIndex(long sequenceIndex) {
...
+     * The following function returns the bounding box corresponding to the child
+     * that is not the path from the node to leaf traversed by a point. If such a
+     * box cannot be defined then the function returns null.
+     * 
+     * @param point the point as above
+     * @param node  the node as above
+     * @return the bounding box of the node
+     */
+
+    public static BoundingBox getSiblingBoundingBox(double[] point, Node node) {
```
jotok, line 436: hIt doesn't look the the `point` argument is being used.

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
```diff
@@ -422,4 +422,24 @@ protected void deleteSequenceIndex(long sequenceIndex) {
...
+        if (node == null)
```
jotok, line 437: Should this method every be called when node is null? Maybe this should throw an exception (e.g. using `checkNotNull` instead of silently returning null.

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
```diff
@@ -422,4 +422,24 @@ protected void deleteSequenceIndex(long sequenceIndex) {
...
+            return null;
+        Node sibling = Node.isLeftOf(point, node) ? node.getRightChild() : node.getLeftChild();
+        if (sibling == null)
+            return null;
```
jotok, line 441: It's only the root node that doesn't have a sibling, right? We could have this method return an `Optional<BoundingBox>` instead, although that would create a transient wrapper class which I know you're trying to avoid.

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
sudiptoguha: isLeftOf uses the argument "point" ... that may be a lengthier change ..

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
sudiptoguha: True. The function should not be called on a null node. Will fix.

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
sudiptoguha: The "sibling" is not the sibling of the node, it is the sibling of the path taken by point. Will change the name, it may avoid confusion.

More specifically, the function should return null for any leaf node (because the path is non-existent).

Java/core/src/main/java/com/amazon/randomcutforest/anomalydetection/AbstractAttributionVisitor.java
```diff
@@ -121,9 +121,9 @@ public void accept(Node node, int depthOfNode) {
             // use the sibling bounding box to represent counterfactual "what if point & the
             // candidate near neighbor
             // had not been inserted in the tree"
-            Node sibling = Node.isLeftOf(pointToScore, node) ? node.getRightChild() : node.getLeftChild();
 
-            shadowBox = shadowBox == null ? sibling.getBoundingBox() : shadowBox.getMergedBox(sibling.getBoundingBox());
+            shadowBox = shadowBox == null ? Node.getSiblingBoundingBox(pointToScore, node)
```
wnbts, line 125: discussion. Should this method be non-static given the node instance?
it can be node.sib(point) or even node.sib(point, shadow)

sudiptoguha: The answer depends on the goal :) For the current context, these three are equivalent. But for CompactRCF, there is no actual node -- the node in this function call would become an instance of INodeView (other suggestions welcome!). The  CompactRCFTree would create a class that provides this interface and spoofs the nodes without creating transient data  structures (which would mean more frequent GC) -- that is the current scenario which we are trying to avoid.

sudiptoguha: But we can also make it node.getSiblingBoundingBox(pointToScore)  which would become nodeView.getSiblingBoundingBox(pointToScore). If we do that would we not increase the size of the Node class for the default context? Admittedly, that may not matter ... but still

wnbts: i am not clear how making it a static method helps compact rcf

sudiptoguha: Static/Non-Static will not matter to CompactRCF. But non-static may increase the space required by the Node class and be detrimental to the current non-compact algorithm. I am happy to make it non-static. @jotok?

sudiptoguha: Actually making it non-static helps parts 2 and 3 of the PR for CompactRCF. Going with node.getSiblingBox() ...

Java/core/src/test/java/com/amazon/randomcutforest/imputation/ImputeVisitorTest.java
```diff
@@ -152,17 +146,17 @@ public void testAccept() {
...
-        when(cut.getDimension()).thenReturn(1).thenReturn(0);
-
-        Node node = mock(Node.class);
-        when(node.getCut()).thenReturn(cut);
-
-        assertTrue(visitor.trigger(node));
-        assertFalse(visitor.trigger(node));
-    }
+    /**
+     * TODO: fix the mocks
```
jotok, line 150: Could we fix this in this PR?

jotok (approved):

PR #74, opened 2020-12-02
Compact rcf

sudiptoguha:
*Description of changes:* Refactored internal state for addPoint


Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeDouble.java
```diff
@@ -368,73 +349,78 @@ private int deletePoint(short nodeOffset, double[] point, int level) {
      * any bounding box.
      */
 
-    class CandidateForSiblingInAddPoint {
+    class AddPointState {
         short siblingOffset;
         int cutDimension;
         double cutValue;
         BoundingBox savedBox;
+        boolean resolved;
```
wnbts, line 357: remark. This is essentially the same as before but better organized. But one consideration is whether this programming paradigm is scalable, i.e. all needed local variables are gathered in a state to pass around together. The state object can grow and get bloated. An alternative is to have variables such as isResolved in the method and use it where needed. For mutable types, there are support like AtomicBoolean.

wnbts (approved):

PR #76, opened 2020-12-10
Compact rcf

sudiptoguha:
*Description of changes:* Links up float[] for compact trees and also allows for bounding boxed to not be cached. The arrays in NodeStore/LeafStore are now behind interfaces and not accessed directly. The logic for add point is also refactored.


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -154,9 +169,21 @@
...
      */
     protected final boolean compactEnabled;
+    /**
+     * Enable float representation; the input/output is double[] but the internal
+     * representation would be float[]. There will be loss of precision in near
+     * neighbors because of the truncation. Otherwise, unless all numbers are in a
+     * very small range, the loss of precision is unlikely to have larger impact
+     * than changing the random seed.
+     */
+    protected final boolean floatEnabled;
```
jotok, line 182: Instead of a boolean, what if we defined an enum for precision?
```
public enum Precision {
  SINGLE, DOUBLE;
}

public static final DEFAULT_PRECISION = Precision.DOUBLE;

protected final Precision precision;
```

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -154,9 +169,21 @@
...
+    /**
+     * The following set to false saves space, but enforces compact representation.
+     */
+    protected final boolean cacheEnabled;
```
jotok, line 186: Maybe this variable name should indicate that it's specifically for caching or saving bounding boxes.

Also, are we going to throw an exception if compact enabled is false and cache enabled is true, or will we just silently ignore the value of compact enabled?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -183,7 +210,7 @@
      */
     protected AbstractForestUpdateExecutor<?> updateExecutor;
 
-    protected PointStoreDouble pointStore;
+    protected IPointStore<?> pointStore;
```
jotok, line 213: Can we remove this instance variable now? Our goal is that all state should be owned by the update coordinator or the component list.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -202,19 +229,38 @@
...
+        if (floatEnabled) {
+            initCompactFloat();
+        } else if (compactEnabled) {
             initCompact();
         } else {
             initNoncompact();
         }
     }
 
     private void initCompact() {
```
jotok, line 241: Should we change this to initCompactDouble() for consistency?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -1051,7 +1103,25 @@ public T threadPoolSize(int threadPoolSize) {
         }
 
         public T compactEnabled(boolean compactEnabled) {
-            this.compactEnabled = compactEnabled;
+            if (!this.compactEnabled) {
```
jotok, line 1106: Did you mean to write `if (!floatEnabled)` here?

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
```diff
@@ -47,6 +48,6 @@ public void completeUpdate(List<Optional<UpdateReturn<Integer>>> updateResults,
     }
 
     public PointStoreDouble getStore() {
-        return store;
+        return (PointStoreDouble) store;
```
jotok, line 51: We should update this method to return IPointStore and then update the corresponding mapper class to do the casting.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStore.java
```diff
@@ -27,12 +27,26 @@
...
+    void decrementRefCount(int index);
+
+    /**
+     * Adds to the store; there may be a loss of precision if enableFloat is on in
+     * the Forest level. But external interface of the forest is double[]
+     * 
+     * @param point point to be added
+     * @return index of the stored point
+     */
+    int add(double[] point);
```
jotok, line 41: Previously this interface exposed a read-only interface, which I thought was good because it meant that trees couldn't change the point store (only the update coordinator could do that). If we still think that's useful, we could create an IReadOnlyPointStore (or something) and add the mutating methods to a sub-interface.

sudiptoguha: Great suggestion. Done. I used IPointStoreReadOnly instead of IReadOnlyPointStore -- the latter seems to be like a statement.

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
sudiptoguha: Agree with the change - let's postpone changing the mapper in this PR?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: yes. Fixed.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Name changed to boundingBoxCachingEnable and is true by default (original behavior). In this PR we silently ignore compactEnabled in the builder  if some other feature mandates the compact representation. We check the consistency at the forest level.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Name changed to singlePrecisionEnabled. Agree with the possibility of enum. All of these flags could do with some rethinking in a future PR.

jotok: https://github.com/aws/random-cut-forest-by-aws/issues/77

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -283,6 +330,10 @@ private RandomCutForest(Builder<?> builder, boolean notUsed) {
         centerOfMassEnabled = builder.centerOfMassEnabled;
         parallelExecutionEnabled = builder.parallelExecutionEnabled;
         compactEnabled = builder.compactEnabled;
+        singlePrecisionEnabled = builder.singlePrecisionEnabled;
+        checkArgument(!singlePrecisionEnabled || compactEnabled, "float representation requires compact trees");
```
jotok, line 334: I think we should move this checks to the constructor (where we perform other validity checks on the builder, like checking that number of trees is positive).

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: If I'm reading this correctly, it means that the builder will let you set compact enabled to true once, but will silently ignore you if you try to change it after that. If we later change the default to having compact enabled be true, this would prevent a user from ever creating a non-compact forest.

My suggestion is that we don't add logic to the builder methods. Let's keep all config validation logic in the constructor. Then, if the user choose an inconsistent set of options we can either throw an exception or override them and print a warning to stderr.

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
jotok: https://github.com/aws/random-cut-forest-by-aws/issues/78

Java/core/src/main/java/com/amazon/randomcutforest/store/ILeafStore.java
```diff
@@ -0,0 +1,41 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest.store;
+
+/**
+ * A read-only point store interface.
```
jotok, line 19: 1. not a point store interface
2. not read-only

Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
```diff
@@ -0,0 +1,45 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest.store;
+
+/**
+ * A read-only point store interface.
```
jotok, line 19: Fix copy-pasted documentation.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStore.java
```diff
@@ -18,21 +18,16 @@
 /**
  * A read-only point store interface.
```
jotok, line 19: Fix documentation.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStoreReadOnly.java
```diff
@@ -0,0 +1,43 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest.store;
+
+/**
+ * A read-only point store interface.
+ */
+public interface IPointStoreReadOnly<P> {
```
jotok, line 21: What do you think of `IPointStoreView` to match `INodeView`?

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoundingBox.java
```diff
@@ -79,8 +81,10 @@ protected BoundingBox(final double[] minValues, final double[] maxValues) {
         dimensions = minValues.length;
     }
 
-    public int getDimensions() {
-        return dimensions;
+    public BoundingBox copy() {
```
jotok, line 84: I think I'd like this better as a copy constructor, rather than starting a new convention where we define a `copy()` method to create a copy.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoundingBox.java
```diff
@@ -129,13 +141,23 @@ public BoundingBox getMergedBox(double[] point) {
     }
 
     @Override
-    public BoundingBox convertBoxToDouble() {
-        return this;
+    public BoundingBox copyBoxToDouble() {
```
jotok, line 144: Is this method defined in the interface? I think we did something wrong if an interface depends on one of its implementations.

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoundingBox.java
```diff
@@ -29,13 +29,25 @@
...
+    IBoundingBox<P> addPoint(P point);
 
-    BoundingBoxFloat convertBoxToFloat();
+    // merges and keeops the older box unchaged
+    IBoundingBox<P> getMergedBox(IBoundingBox<P> otherBox);
+
+    // merges in place
+    IBoundingBox<P> addBox(IBoundingBox<P> otherBox);
+
+    BoundingBox copyBoxToDouble();
```
jotok, line 47: Would everything work if the return type of this method was `IBoundingBox<double[]>`?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: I agree with that in principle -- but right now the constructors are being called based on these flags! We need to rethink the flags as a whole.

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoundingBox.java
```diff
@@ -29,13 +29,25 @@
...
+
+    float getMinValueFloat(int i);
```
jotok, line 49: Where are these methods used? Can't we just cast to float there instead?

sudiptoguha: Repeatedly casting values can be an issue. But anyway these are removed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoundingBox.java
sudiptoguha: At the current moment the Visitors need BoundingBox -- we should change that in a subsequent PR.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoundingBox.java
sudiptoguha: The visitors use BoundingBox  ... unless we support those functions, we cannot remove this.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoundingBox.java
sudiptoguha: Fixed, changed to copyBox.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStoreReadOnly.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStore.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/ILeafStore.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStoreView.java
```diff
@@ -0,0 +1,43 @@
...
+
+    int getCapacity();
+
+    boolean pointEquals(int index, P point);
+
+    P get(int index);
+
+    void incrementRefCount(int index);
+
+    void decrementRefCount(int index);
```
wnbts, line 32: these two methods do not appear a read only view

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -124,10 +124,10 @@ protected short takeIndex() {
      * 
      * @param index The index value to release.
      */
-    protected void releaseIndex(short index) {
+    protected void releaseIndex(int index) {
         checkValidIndex(index);
         occupied.clear(index);
-        freeIndexes[++freeIndexPointer] = index;
+        freeIndexes[++freeIndexPointer] = (short) index;
```
wnbts, line 130: this will cause some very hard problems

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStore.java
jotok: The other nice thing about this naming convention is that IPointStore and IPointStoreReadOnly will appear next to each other when listed alphabetically.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -154,9 +168,21 @@
...
      */
     protected final boolean compactEnabled;
+    /**
+     * Enable float representation; the input/output is double[] but the internal
+     * representation would be float[]. There will be loss of precision in near
+     * neighbors because of the truncation. Otherwise, unless all numbers are in a
+     * very small range, the loss of precision is unlikely to have larger impact
+     * than changing the random seed.
+     */
+    protected final boolean singlePrecisionEnabled;
```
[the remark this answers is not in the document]
jotok: We are currently proposing to support both single and double precision in 2.0. However, we're also planing to change this boolean flag to an enum value because think it would be clearer. https://github.com/aws/random-cut-forest-by-aws/issues/77

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: https://github.com/aws/random-cut-forest-by-aws/issues/82

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: https://github.com/aws/random-cut-forest-by-aws/issues/82

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStore.java
sudiptoguha: Yes, as we specialize implementation, it'll be good to keep library internal assumptions separate from use.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStoreView.java
sudiptoguha: Indeed!

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: Locally or globally?

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoundingBox.java
```diff
@@ -29,13 +29,25 @@
...
-    BoundingBox convertBoxToDouble();
+    // merges in place
+    IBoundingBox<P> addPoint(P point);
 
-    BoundingBoxFloat convertBoxToFloat();
+    // merges and keeops the older box unchaged
+    IBoundingBox<P> getMergedBox(IBoundingBox<P> otherBox);
+
+    // merges in place
+    IBoundingBox<P> addBox(IBoundingBox<P> otherBox);
```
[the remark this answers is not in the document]
sudiptoguha: Yes, but if we want minValues/maxValues to be final (there are good reasons for that) and we wish to use minValues = maxValues = point (there is good for reason for that, why should we copy a point in  all the 10+ trees twice) -- then the addBox becomes a best-effort. There are some scenarios where a new box will be created. But we can merge .. in. a subsequent PR we probably should have BoundingBox as Abstract classes and not as interfaces.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -107,6 +109,18 @@
      */
     public static final boolean DEFAULT_COMPACT_ENABLED = false;
 
+    /**
+     * By default, trees will not use float[] to save space. Enabling this will
+     * force enableCompact
```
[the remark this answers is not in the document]
sudiptoguha: Agree. We should rethink all these flags together - I think we have an issue for this.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -107,6 +109,18 @@
...
+     */
+    public static final boolean DEFAULT_FLOAT_ENABLED = false;
+
+    /**
+     * By default, bounding boxes will be used. Disabling this will force
+     * enableCompact .
+     */
+    public static final boolean DEFAULT_CACHE_ENABLED = true;
```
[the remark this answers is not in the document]
sudiptoguha: Agree.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -183,7 +209,7 @@
      */
     protected AbstractForestUpdateExecutor<?> updateExecutor;
 
-    protected PointStoreDouble pointStore;
+    // protected IPointStore<?> pointStore;
```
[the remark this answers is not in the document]
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/state/store/LeafStoreState.java
```diff
@@ -26,7 +26,7 @@
 
     public int[] pointIndex;
     public short[] parentIndex;
```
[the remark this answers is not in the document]
sudiptoguha: Need not be. pointIndex is for the whole forest, so the number of points can be sampleSize*numberOfTrees in the worst case and can easily exceed max short value. The parentIndex however applies locally to a tree and is unlikely to more than 16K. We should someday probably enable LafStoreStateLarge ... but there hards seems reasons for that at this point. The allocation will save a bit of space per tree, and can matter with more trees.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/LeafStoreState.java
```diff
@@ -26,7 +26,7 @@
...
-    public int[] mass;
+    public short[] mass;
```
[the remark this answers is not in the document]
sudiptoguha: mass refers to the local tree -- so the same statement as parentIndex applies.

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeState.java
```diff
@@ -25,7 +25,7 @@
 public class CompactRandomCutTreeState {
     public LeafStoreState leafStoreState;
     public NodeStoreState nodeStoreState;
-    public short rootIndex;
+    public int rootIndex;
```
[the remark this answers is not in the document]
sudiptoguha: Yes. It probably makes sense that the trees assume int/double for all variables (so that we do not copy codes for int/short). The specific storage classes can perform the conversations (if needed)

Java/core/src/main/java/com/amazon/randomcutforest/store/ILeafStore.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public interface ILeafStore {
```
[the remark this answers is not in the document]
sudiptoguha: fixed

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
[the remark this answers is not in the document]
jotok: https://github.com/aws/random-cut-forest-by-aws/issues/82

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoundingBox.java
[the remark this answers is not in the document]
jotok: * As discussed, we need a read-only BoundingBox interface that can be returned by an INodeView (in order to make visitor implementations data structure agnostic). I'm not sure if that should be this interface, or if we need another `IBoundingBox extends IBoundingBoxView` dichotomy.
* Does the mutable bounding box interface need to explicitly define two separate methods for merging in-place versus returning a merged copy? Can the decision about what to do when merging be left to the implementations?

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoundingBox.java
jotok: https://github.com/aws/random-cut-forest-by-aws/issues/99

jotok: Can you add a TODO comment to remove this method?

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoundingBox.java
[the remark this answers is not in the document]
sudiptoguha: Lets do that now -- this also solves the other issue that we are/were overloading the BoundingBox to be both for inference/postprocessing via the INodeView as well as a building block for internal structure.

I think using AbstractBoundingBox and subclassing to BoundingBoxDouble and BoundingBoxFloat may help remove common code as well and force these assumptions to be explicit. IBoundingBox reduces the number of functions (only those needed by current Visitors, more can be added when we come to it) -- this is the IBoundingBoxView version.

To answer the question re mutable boxes: I see this as a signal in the code that the current box has no further use and therefore can be reclaimed without memory alloc/dealloc. In the other method the current box continues to have relevance and should not be mutated.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoundingBox.java
jotok: #99

jotok (approved):

jotok (approved):

PR #88, opened 2020-12-14
Improvements to sampler package

jotok:
* Standardize field names in CompactSampler
* Update Javadocs
* Add ability to validate the heap property in CompactSampler
* Move common methods to the IStreamSampler interface
* Add `@Data` annotation to Weighted class and remove unused methods
* Improve package test coverage
* Set fields in CompactSamplerState to private


jotok:
The initial commit failed to build due to a javadoc error (namely, using `<p/>` instead of `<p>` in a javadoc comment). Fixed and amended commit.

jotok:
Rebased.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -163,58 +197,75 @@ public static CompactSampler uniformSampler(int sampleSize, long seed, boolean s
     public Optional<Float> acceptSample(long sequenceIndex) {
         evictedPoint = null;
         float weight = computeWeight(sequenceIndex);
-        entriesSeen++;
         if (size < capacity) {
```
sudiptoguha, line 200: Do we want to store the information that acceptSample was called and some weight generated, even if nothing was evicted?

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -163,58 +197,75 @@ public static CompactSampler uniformSampler(int sampleSize, long seed, boolean s
...
+            swapDown(i, validate);
         }
     }
 
     @Override
-    public void addSample(Integer pointRef, float weight, long seqNUm) {
-        checkState(size < capacity, " sampler full");
-        weightArray[size] = weight;
-        referenceArray[size] = pointRef;
+    public void addSample(Integer pointIndex, float weight, long sequenceIndex) {
```
sudiptoguha, line 258: Do we want to check if a acceptSample was called (getEvictedPoint)? And reset that state? That would be true if size = capacity - 1.

Should we check for weights generated earlier in an acceptSample even if nothing was evicted? Probably not. But this does leaf the feature that we can artificially change the weight of a an element that is sampled to have different duration in the sampler.

sudiptoguha (approved):

PR #101, opened 2020-12-17
Remove the Sequential interface

### LINKED ISSUE #80: Remove the Sequential interface
The Sequential interface was introduced in an early 2.0 refactoring. At that point we thought that the input to IUpdatable could be completely generic. We're walking this back now, because the Sequential class was basically a wrapper around a pair of values, and it only existed between the update call and the tree add-point method. Since all update code paths are following the same pattern, we've decided to revert to passing sequence index as a method argument instead of passing it via a short-lived object.

jotok:
Closes #80

This change makes the sequence index an explicit argument to the sampler and tree interfaces. As a result we can remove the Sequential container class. Other changes:

* There is a new interface Sampled that which is used in the return type of `getSamples()` and `getEvictedPoint()`.
* The Weighted container class implements Sampled
* The signatures of `acceptSample` and `addSample` are changed to remove redundancy and to avoid leaking implementation details. In particular, `acceptSample` no longer returns a weight, because the use of weights is a feature of time-decayed sampling, but not a requirement for stream sampling.


Java/core/src/main/java/com/amazon/randomcutforest/executor/SamplerPlusTree.java
```diff
@@ -56,26 +57,30 @@ public SamplerPlusTree(IStreamSampler<P> sampler, ITree<P> tree) {
...
-        Optional<Float> presampleResult = sampler.acceptSample(seqNum);
-        if (presampleResult.isPresent()) {
-            Optional<Sequential<P>> deletedPoint = sampler.getEvictedPoint();
+        if (sampler.acceptSample(sequenceIndex)) {
+            Optional<Sampled<P>> deletedPoint = sampler.getEvictedPoint();
             if (deletedPoint.isPresent()) {
-                tree.deletePoint(deletedPoint.get());
+                Sampled<P> p = deletedPoint.get();
+                tree.deletePoint(p.getValue(), p.getSequenceIndex());
                 deleteRef = deletedPoint.get().getValue();
```
jotok, line 75: This should be `p.getValue()`.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -271,32 +298,30 @@ public void addSample(Integer pointIndex, float weight, long sequenceIndex) {
...
-        for (int i = 0; i < size; i++) {
-            result.add(new Weighted<>(pointIndex[i], weight[i]));
-        }
-        return result;
+        return streamSamples().collect(Collectors.toList());
+    }
+
+    private Stream<Weighted<Integer>> streamSamples() {
+        return IntStream.range(0, size).mapToObj(i -> {
+            long index = sequenceIndex != null ? sequenceIndex[i] : 0L;
```
jotok, line 315: This needs to be documented (the fact that we set sequence index to 0 in the returned values if store-sequence-index is not enabled.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/Sampled.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.amazon.randomcutforest.sampler;
+
+/**
+ * A simple wrapper class representing a point that has been sampled by a
+ * sampler. A sampled point can be added to or removed from a tree.
+ * 
+ * @param <P> The point representation used by this sampled point.
+ */
+public interface Sampled<P> {
```
jotok, line 24: I guess this should be ISampled since we're using the I-prefix on all our new interfaces.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
```diff
@@ -102,121 +114,88 @@ public SimpleStreamSampler(int sampleSize, double lambda, long seed) {
...
+    public Optional<ISampled<P>> getEvictedPoint() {
         return Optional.ofNullable(evictedPoint);
     }
 
     /**
      * @return the list of weighted points currently in the sample. If there is no
      *         sequential information then a dummy variable is placed.
      */
-    @Override
     public List<Weighted<P>> getWeightedSamples() {
```
jotok, line 173: Change to `getWeightedSample` and update comment to reflect the fact that sequence indexes are always stored in SimpleStreamSampler.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
```diff
@@ -102,121 +114,88 @@ public SimpleStreamSampler(int sampleSize, double lambda, long seed) {
...
-            result = new ArrayList<>();
-            weightedSamples.forEach(e -> result.add(new Weighted<>(e.getValue(), e.getWeight())));
-        }
-        return result;
-
+        return new ArrayList<>(samples);
     }
 
     /**
      * @return the list of weighted points currently in the sample. If there is no
```
jotok, line 178: This comment is no longer accurate.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
```diff
@@ -18,71 +18,76 @@
 import java.util.List;
 import java.util.Optional;
 
-import com.amazon.randomcutforest.executor.Sequential;
-
 /**
  * A sampler that samples from an ordered sequence of points.
```
jotok, line 22: Add an explanation for the two-step sampling process.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
```diff
@@ -54,7 +64,7 @@
      * element is the lowest priority point in the sample (or, equivalently, is the
      * point with the greatest weight).
      */
-    private final Queue<Weighted<P>> weightedSamples;
+    private final Queue<Weighted<P>> samples;
```
jotok, line 67: Change name to `sample` to reflect new naming convention.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
```diff
@@ -102,121 +114,88 @@ public SimpleStreamSampler(int sampleSize, double lambda, long seed) {
...
+    @Override
+    public List<ISampled<P>> getSample() {
+        return new ArrayList<>(samples);
     }
 
     /**
      * @return true if this sampler contains enough points to support the anomaly
      *         score computation, false otherwise.
      */
     public boolean isReady() {
```
jotok, line 190: `isReady` and `isFull` have default implementations in the interface now.

Java/core/src/test/java/com/amazon/randomcutforest/sampler/SimpleStreamSamplerTest.java
```diff
@@ -84,35 +89,36 @@ public void testAddSample() {
...
-        assertEquals(0.5f, samples.get(1).getWeight());
+        assertEquals(weight1, samples.get(1).getWeight());
 
         assertArrayEquals(new double[] { -2.2 }, samples.get(2).getValue());
-        assertEquals(1.2f, samples.get(2).getWeight());
+        assertEquals(weight2, samples.get(2).getWeight());
     }
 
     @Test
     public void testAcceptSample() {
```
jotok, line 102: Update test names in sampler package to reflect the new interface method names.

sudiptoguha (approved):
Nice, much cleaner!

Issue #82, opened 2020-12-14
Move validation of builder parameters to constructor


jotok:
Currently some validation is happening in the initCompactFloat() method. We should put all validation in one place and fail fast if a user has selected an incompatible set of options.

jotok:
We should also remove logic from the builder class that overrides user settings.

jotok:
What exception should be thrown when valdiation fails:
* IllegalArgumentException
* UnsupportedOperationException

PR #102, opened 2020-12-17
Updates to forest/builder configuration

### LINKED ISSUE #77: Precision setting should be an enum
We are allowing the user to decide to using single or double precision floating point numbers internally. Currently this is implemented as a boolean flag. I think this setting would make more sense if we defined a enum for precision:

```
public enum Precision {
  SINGLE, DOUBLE;
}
```

```
RandomCutForest forest = RandomCutForest.builder()
  .precision(Precision.SINGLE)
  .build();
```

I think this is more self-explanatory, and it leaves the possibility to support another level of precision (e.g. BigDecimal) if think it would be useful.

jotok:
* Change the singlePrecisionEnabled flag to use a new Precision enum
* Remove logic from the builder that overrides user options. Instead
  we will fail fast in the constructor.

Closes #77 #82


sudiptoguha (approved):
Nice!

PR #108, opened 2020-12-21
removing arbitrary nodeview

sudiptoguha:
*Issue #, if available:* 106

Removes getNodeView.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -90,12 +90,12 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
     }
 
     @Override
-    INode<Integer> getNodeView(Integer node) {
+    protected INode<Integer> getNode(Integer node) {
         return new CompactNodeView(this, node);
     }
 
     @Override
-    void addSequences(Integer nodeRef, long sequenceIndex) {
+    protected void addSequenceIndex(Integer nodeRef, long sequenceIndex) {
         if (sequenceIndexes[nodeRef] == null) {
             sequenceIndexes[nodeRef] = new HashSet<Long>();
         }
@@ -103,7 +103,7 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
     }
 
     @Override
-    void deleteSequences(Integer nodeRef, long sequenceIndex) {
+    protected void deleteSequenceIndex(Integer nodeRef, long sequenceIndex) {
         if (sequenceIndexes[nodeRef] != null) {
             sequenceIndexes[nodeRef].remove(sequenceIndex);
         }
@@ -166,10 +166,10 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
     abstract protected boolean leftOf(Point point, int cutDimension, double cutValue);
 
     @Override
-    abstract boolean checkEqual(Point oldPoint, Point point);
+    abstract protected boolean checkEqual(Point oldPoint, Point point);
 
     @Override
-    abstract String toString(Point point);
+    abstract protected String toString(Point point);
 
     // returns the point based on position in the store
     @Override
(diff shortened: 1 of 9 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -87,10 +86,14 @@ public AbstractCompactRandomCutTree(int maxSize, long seed, LeafStore leafStore,
         this.rootIndex = rootIndex;
         // random = new Random(seed);
         nodeManager = new CompactNodeManager(nodeStore, leafStore, maxSize);
-        nodeView = new CompactNodeView(this, rootIndex);
         this.enableCache = enableCache;
     }
 
+    @Override
+    INode<Integer> getNodeView(Integer node) {
```
jotok, line 93: By not including a visibility modifier (like public or protected), this method will be visible inside the tree package but not visible to any subclasses of AbstractCompactRandomCutTree. Is that intentional?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -541,28 +544,28 @@ protected boolean leftOf(double[] point, NodeReference nodeReference) {
...
+    private <R> void traverseTreeMulti(double[] point, MultiVisitor<R> visitor, NodeReference node, int depthOfNode) {
+        if (isLeaf(node)) {
+            visitor.acceptLeaf(getNodeView(node), depthOfNode);
         } else {
-            if (visitor.trigger(nodeView.getNodeView(currentNode))) {
-                NodeReference leftChild = getLeftChild(currentNode);
-                traverseTreeMulti(point, visitor, leftChild, nodeView.getNodeView(leftChild), depthOfNode + 1);
+            if (visitor.trigger(getNodeView(node))) {
+                traverseTreeMulti(point, visitor, getLeftChild(node), depthOfNode + 1);
                 MultiVisitor<R> newVisitor = visitor.newCopy();
```
jotok, line 557: Is this correct? It seems weird that we make a copy of visitor after traversing the left path, and then we combine with visitor after traversing the right path. It seems like we're including the result of traversing the left path twice.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
jotok, line 93: Shouldn't `getNodeView` return an `INodeView` and not an `INode`?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -87,10 +86,14 @@ public AbstractCompactRandomCutTree(int maxSize, long seed, LeafStore leafStore,
...
+        return new CompactNodeView(this, node);
+    }
+
     @Override
     void addSequences(Integer nodeRef, long sequenceIndex) {
```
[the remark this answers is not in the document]
sudiptoguha: True. Changed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/INode.java
```diff
@@ -17,7 +17,7 @@
 
 public interface INode<NodeReference> extends INodeView<NodeReference> {
```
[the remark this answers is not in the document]
sudiptoguha: No, but lets have this a separate PR.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: No. These methods should be visible to sub-classes, so that other kinds of trees can be built using the same set up.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: This does allow a stronger visitor pattern instead of making a clean copy -- the pattern now also can express a preorder traversal.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: The function should be a getNode(node) because the priviledges correspond to INode and more than just view.

Java/core/src/main/java/com/amazon/randomcutforest/tree/INode.java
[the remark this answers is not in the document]
jotok: https://github.com/aws/random-cut-forest-by-aws/issues/111

jotok (approved):

PR #110, opened 2020-12-22
Change IUpdatable interface to not return nested Optional

### LINKED ISSUE #81: Review IUpdatable interface
This is an open-ended task to review and possibly change the abstractions used in IUpdateCoordinator and IUpdatable. The UpdateReturn class in particular was added as part of a larger restructuring change and didn't get a lot of scrutiny when it was introduced.

jotok:
Rename UpdateReturn to UpdateResult, and add a new method
isStateChange() that can be called to determine if the update changed
the model.

In addition to the above change, this PR adds test coverage to the executor package.

Closes #81


jotok:
Rebase.

wnbts (approved):
looks good! i suggested the same structure a while ago.

PR #118, opened 2020-12-29
Compact rcf

sudiptoguha:
*Issue #, if available:* 117

*Description of changes:*
Fixes bounding box calculations.


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -216,6 +218,33 @@ abstract protected NodeReference addNode(NodeReference parent, NodeReference lef
...
+            }
+        }
+    }
+
+    // manages the bounding boxes and center of mass
+    void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
+        NodeReference tempNode = nodeReference;
+        boolean boxNeedsUpdate = enableCache;
+        while (boxNeedsUpdate && tempNode != null) {
+            boxNeedsUpdate = !modifyBoxAndCheckContains(tempNode, point);
```
jotok, line 240: OK, I see why you used `modify` now, but I think we should either use `update` to be consistent with this method, or else `compute` or `recompute` to indicate that we're computing a deterministic function. And then putting a comment on that function to explain that it's called after something has changed in the subtree would be helpful.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -216,6 +218,33 @@ abstract protected NodeReference addNode(NodeReference parent, NodeReference lef
...
+            tempNode = getParent(tempNode);
+        }
+
+        recursivelyReadjustPointSum(nodeReference, point);
```
jotok, line 244: See, we have "update" and "modify" and "adjust" all in this one class. I'm in favor of using descriptive names, but to me those three are too similar to justify using all three. "Readjust" is better since it indicates that we're doing something again, but I think I prefer "recompute" because it suggests more strongly that there's a correct result.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -276,24 +303,47 @@ private boolean deletePoint(NodeReference nodeReference, Point point, long seque
                 setParent(rootIndex, null);
             } else {
                 replaceNodeBySibling(grandParent, parent, nodeReference);
+                updateAncestorNodesAfterDelete(grandParent, point);
```
jotok, line 306: Similarly, we have  both "recursively readjust" and "update ancestor nodes" which seem to be conveying pretty similar ideas. Can we standardize on something?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -276,24 +303,47 @@ private boolean deletePoint(NodeReference nodeReference, Point point, long seque
...
             }
             delete(nodeReference);
             delete(parent);
-            return false;
+            return;
```
jotok, line 310: minor: unnecessary return statement

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -376,7 +424,8 @@ void resolve(Point point, NodeReference parentIndex,
                 Cut cut = randomCut(random, mergedBox);
                 // the cut is between a leaf node and the new point; it must exist
                 AddPointState<Point, NodeReference, PointReference> newState = new AddPointState<>(pointIndex);
-                newState.initialize(nodeReference, cut.getDimension(), cut.getValue(), sequenceNumber, mergedBox);
+                newState.initialize(nodeReference, cut.getDimension(), cut.getValue(), sequenceNumber, mergedBox,
```
jotok, line 427: I would consider dropping the initialize method, since it's just setting fields, and setting the fields directly in the method instead.

```
AddPointState<Point, NodeReference, PointReference> state = new AddPointState<>(pointIndex);
state.setNodeReference(nodeReference);
state.setCutDimension(cut.getDimensions());
// etc.
```

I guess what `initialize` gets us is that it won't let you forget to set one of the arguments, and it turns 6 or 7 lines of repeated code into a single long method invocation. I don't feel too strongly about this one.

Although, I do wonder: why is `pointIndex` passed to the constructor, but everything else is set by `initialize`?

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeDouble.java
```diff
@@ -111,26 +82,33 @@ protected boolean leftOf(double[] point, int dimension, double val) {
...
-     *
-     * @param nodeReference node in question
-     * @return the bounding box
-     */
-
-    AbstractBoundingBox<double[]> getBoundingBoxReflate(Integer nodeReference) {
-        if (isLeaf(nodeReference)) {
-            return new BoundingBox(getPointFromLeafNode(nodeReference));
+    @Override
+    void readjustPointSum(Integer node, double[] point) {
```
jotok, line 104: Same as above, I kind of like "recompute" better than "readjust" for this.

Java/core/src/main/java/com/amazon/randomcutforest/tree/PointerTree.java
```diff
@@ -58,36 +58,27 @@ protected String toString(double[] doubles) {
         return Arrays.toString(doubles);
     }
 
-    void updateBox(Node nodeReference) {
-        checkState(enableCache, " incorrect invocation");
-        BoundingBox rightBox = nodeReference.getRightChild().getBoundingBox();
-        BoundingBox leftBox = nodeReference.getLeftChild().getBoundingBox();
-        nodeReference.setBoundingBox(leftBox.getMergedBox(rightBox));
+    @Override
+    void setCachedBoxes(Node node, AbstractBoundingBox<double[]> savedBox) {
```
jotok, line 62: Shouldn't this be `setCachedBox` (singular)?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -216,6 +218,33 @@ abstract protected NodeReference addNode(NodeReference parent, NodeReference lef
...
+            while (tempNode != null) {
+                readjustPointSum(tempNode, point);
+                tempNode = getParent(tempNode);
+            }
+        }
+    }
+
+    // manages the bounding boxes and center of mass
+    void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
+        NodeReference tempNode = nodeReference;
```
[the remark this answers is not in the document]
sudiptoguha: The function recursively updates the bounding boxes and bubbles up (after the delete) -- comment added. tempNode would keep changing. We can make nodeReference change.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -216,6 +218,33 @@ abstract protected NodeReference addNode(NodeReference parent, NodeReference lef
 
     abstract protected void deleteSequenceIndex(NodeReference node, long uniqueSequenceNumber);
 
+    abstract boolean modifyBoxAndCheckContains(NodeReference tempNode, Point point);
```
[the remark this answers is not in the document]
sudiptoguha: renamed to reComputeAndCheckContains ... however having done so, perhaps we need a refactor into those two basic functions ... Pointer trees (if they are caching the box needs to update the box).

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: renamed to reComputePointSum

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeDouble.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/tree/PointerTree.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: The else part was/is not present. We do not want the subsequent block evaluated.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: removing the recursively from all names, stnadardizing on updateAncestorX

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: this corresponds to a recompute version of getReflatedBox ... so added a flag to getRelated and reused that, along with contains.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: The case of pointIndex being passed to constructor is for an exact duplicate. We need that state to be returned. The other stuff is basically context. I preferred setting the 4 in a single step because they are interdependent.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
jotok: Oops, I was thinking that this was the last statement in a function call (and not the last statement in an interior code block).

jotok (approved):
I create a new issue (#119) to revisit some method names, but I don't think it needs to block this PR.

PR #134, opened 2021-01-29
fixing size restrictions

sudiptoguha:
*Issue #, if available:* 127

*Description of changes:* Fixes restriction of 16383 for size of compact trees

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -26,6 +26,8 @@ import com.amazon.randomcutforest.store.INodeStore;
 import com.amazon.randomcutforest.store.IPointStoreView;
 import com.amazon.randomcutforest.store.LeafStore;
 import com.amazon.randomcutforest.store.NodeStore;
+import com.amazon.randomcutforest.store.SmallLeafStore;
+import com.amazon.randomcutforest.store.SmallNodeStore;
 
 /**
  * A Compact Random Cut Tree is a tree data structure whose leaves represent
@@ -52,6 +54,8 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
      */
     public static final int NULL = -1;
 
+    public static final int switchNodeStore = 16383;
+
     protected int maxSize;
     protected final ILeafStore leafStore;
     protected final INodeStore nodeStore;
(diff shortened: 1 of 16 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -52,6 +54,8 @@
      */
     public static final int NULL = -1;
 
+    public static final int switchNodeStore = 16383;
```
jotok, line 57: I think this would be more self-explanatory if it was a constant in the SmallNodeStore class: `SmallNodeStore.MAX_TREE_SIZE` or something like that.

Java/serialization-json/src/test/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDeTests.java
```diff
@@ -31,7 +31,7 @@
 
     private static final int dimensions = 4;
     private static final int numberOfTrees = 30;
-    private static final int sampleSize = 128;
+    private static final int sampleSize = 28000;
```
jotok, line 34: Did you mean to leave this in? It's not super important in the long run since we plan to remove this class.

Java/core/src/test/java/com/amazon/randomcutforest/store/LeafStoreTest.java
```diff
@@ -24,12 +24,12 @@
 public class LeafStoreTest {
```
jotok, line 24: We should rewrite this test and `NodeStoreTest` to cover both small and regular stores. We can create an issue to do that in a follow up PR if you don't want to add it here.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done

Java/serialization-json/src/test/java/com/amazon/randomcutforest/serialize/RandomCutForestSerDeTests.java
sudiptoguha: Yes, this is a test that checks that the state variables are handled for the larger NodeStore/LeafStore (hence the 28000). This test will likely move to somewhere else - but for now this seemed to be a good place to have it.

Java/core/src/test/java/com/amazon/randomcutforest/store/LeafStoreTest.java
sudiptoguha: Yes, lets keep that separate. Since multiple things have changed, we would perhaps have to look at overall test coverage as well.

jotok (approved):

Issue #128, opened 2021-01-27
Remove serialization-json package


jotok:
With our new state classes, we no longer require special logic for serializing to JSON (or any other format).

jotok:
Don't forget to update the README to remove references to the SerDe benchmark (which will also have to be removed).

jotok:
Closed by #137

PR #135, opened 2021-01-29
Add new examples module

### LINKED ISSUE #129: Add examples package
Create a set of code examples

* Density estimation
* Forecasting
* Serialization with Jackson and protostuff

I suggest we give each example it's own subdirectory and build file:

```
/Java/Examples/DensityEstimation/pom.xml
```

jotok:
Add new module with serialization examples and runner.

We are planning to remove the serialization-json module from v2 (#128) because the new state classes are POJOs that shouldn't require special logic to work with any Java serialization library. However, we think it will be useful for users to see code examples of serialization, as well as applications of different RCF algorithms. This PR creates a new Maven module for examples that can be invoked by a simple CLI runner. Since these are primarily intended to be code examples and not  functional utilities, the design does not include support for passing arguments to the examples.

Closes #129 

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/runner/ArgumentParser.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/runner/ArgumentParser.java
@@ -28,7 +28,7 @@ import java.util.function.Function;
  */
 public class ArgumentParser {
 
-    public static final String ARCHIVE_NAME = "target/random-cut-forest-1.0.jar";
+    public static final String ARCHIVE_NAME = "randomcutforest-core-1.0.jar";
     private final String runnerClass;
     private final String runnerDescription;
     private final Map<String, Argument<?>> shortFlags;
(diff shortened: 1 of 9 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/runner/ArgumentParser.java
```diff
@@ -28,7 +28,7 @@
  */
 public class ArgumentParser {
 
-    public static final String ARCHIVE_NAME = "target/random-cut-forest-1.0.jar";
+    public static final String ARCHIVE_NAME = "randomcutforest-core-1.0.jar";
```
sudiptoguha, line 31: Good change!

Java/examples/src/main/java/com/amazon/randomcutforest/examples/serialization/JsonExample.java
```diff
@@ -0,0 +1,111 @@
...
+        return "json";
+    }
+
+    @Override
+    public String description() {
+        return "serialize a Random Cut Forest as a JSON string";
+    }
+
+    @Override
+    public void run() throws Exception {
```
sudiptoguha, line 42: Can this be run in IDE?

sudiptoguha (approved):

Java/examples/src/main/java/com/amazon/randomcutforest/examples/serialization/JsonExample.java
jotok: I added a main method to each example so that the example code can be easily invoked from an IDE.

PR #136, opened 2021-01-30
dynamic caching of bounding boxes

sudiptoguha:
*Description of changes:* adds capability of dynamically adjusting the fraction of bounding boxes to saved. The remainder of boxes are recomputed as needed. Setting a low number reduces the in memory footprint but reduces throughput.


Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
```diff
@@ -19,4 +19,5 @@
 import com.amazon.randomcutforest.executor.IUpdatable;
 
 public interface IComponentModel<P> extends ITraversable, IUpdatable<P> {
+    void setBoundingBoxCacheFraction(double fraction);
```
jotok, line 22: I think the other option we could implement for this would be to add a `getTree()` method either to this interface or to `IUpdatable`. Then in the forest and the update executor we add a method:
```
public <R> List<R> mapTrees(Function<ITree<?>, R> function)
```
Then we could implement `setBoundingBoxCacheFraction` in `RandomCutForest` like this:
```
public void setBoundingBoxCacheFraction(double cacheFraction) {
  updateExecutor.mapTrees(t -> t.setBoundingBoxCacheFraction(cacheFraction));
}

sudiptoguha: Done!

jotok (approved):

PR #141, opened 2021-02-02
dynamic samplers

sudiptoguha:
*Issue #, if available:* example of dynamic throughput and sampling adjustments

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -160,7 +160,7 @@ public class RandomCutForest {
     /**
      * The decay factor (lambda value) used by stream samplers in this forest.
      */
-    protected final double lambda;
+    protected double lambda;
     /**
      * Store the time information
      */
@@ -491,7 +491,8 @@ public class RandomCutForest {
      */
     public void setLambda(double lambda) {
         checkArgument(0 <= lambda, String.format("lambda cannot be negative"));
-        updateExecutor.forEachSampler(t -> t.setLambda(lambda));
+        this.lambda = lambda;
+        updateExecutor.forEachSampler(t -> t.setTimeDecay(lambda));
     }
 
     /**
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -489,7 +489,6 @@ public class RandomCutForest {
      * 
      * @param lambda new value of sampling rate
      */
-
     public void setLambda(double lambda) {
         checkArgument(0 <= lambda, String.format("lambda cannot be negative"));
         updateExecutor.forEachSampler(t -> t.setLambda(lambda));
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -485,6 +485,17 @@ public class RandomCutForest {
     }
 
     /**
+     * changes the setting of time dependent sampling on the fly
+     * 
+     * @param lambda new value of sampling rate
+     */
+
+    public void setLambda(double lambda) {
+        checkArgument(0 <= lambda, String.format("lambda cannot be negative"));
+        updateExecutor.forEachSampler(t -> t.setLambda(lambda));
+    }
+
+    /**
      * Visit each of the trees in the forest and combine the individual results into
      * an aggregate result. A visitor is constructed for each tree using the visitor
      * factory, and then submitted to
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
@@ -97,7 +97,19 @@ public class CompactSampler implements IStreamSampler<Integer> {
      * The decay factor used for generating the weight of the point. For greater
      * values of lambda we become more biased in favor of recent points.
      */
-    private final double lambda;
+    private double lambda;
+    /**
+     * The last timestamp when lambda was changed
+     */
+    private long lastUpdateOflambda = 0;
+    /**
+     * most recent timestamp
+     */
+    private long mostRecentTimeStamp = 0;
+    /**
+     * The accumulated sum of lambda before the last update
+     */
+    private float accumulatedLambda = 0;
     /**
      * The random number generator used in sampling.
      */
@@ -372,8 +384,21 @@ public class CompactSampler implements IStreamSampler<Integer> {
         while (randomNumber == 0d) {
             randomNumber = random.nextDouble();
         }
+        mostRecentTimeStamp = (mostRecentTimeStamp < sequenceIndex) ? sequenceIndex : mostRecentTimeStamp;
+        return (float) (-(sequenceIndex - lastUpdateOflambda) * lambda + accumulatedLambda
+                + Math.log(-Math.log(randomNumber)));
+    }
 
-        return (float) (-sequenceIndex * lambda + Math.log(-Math.log(randomNumber)));
+    /**
+     * sets the lambda on the fly
+     * 
+     * @param newLambda the new sampling rate
+     */
+    @Override
+    public void setLambda(double newLambda) {
+        accumulatedLambda += lastUpdateOflambda * lambda;
+        lambda = newLambda;
+        lastUpdateOflambda = mostRecentTimeStamp;
     }
 
     /**
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
@@ -76,49 +76,60 @@ public class CompactSampler implements IStreamSampler<Integer> {
      * point with the greatest weight).
      */
     protected final float[] weight;
+
     /**
      * Index values identifying the points in the sample. See
      * {@link com.amazon.randomcutforest.store.IPointStore}.
      */
     protected final int[] pointIndex;
+
     /**
      * Sequence indexes of the points in the sample.
      */
     protected final long[] sequenceIndex;
+
     /**
      * The number of points in the sample when full.
      */
     protected final int capacity;
+
     /**
      * The number of points currently in the sample.
      */
     protected int size;
+
     /**
      * The decay factor used for generating the weight of the point. For greater
      * values of lambda we become more biased in favor of recent points.
      */
     private double lambda;
+
     /**
      * The last timestamp when lambda was changed
      */
     private long lastUpdateOflambda = 0;
+
     /**
-     * most recent timestamp
+     * most recent timestamp, used to determine lastUpdateOfLambda
      */
     private long mostRecentTimeStamp = 0;
+
     /**
      * The accumulated sum of lambda before the last update
      */
-    private float accumulatedLambda = 0;
+    private double accumulatedLambda = 0;
+
     /**
      * The random number generator used in sampling.
      */
     private final Random random;
+
     /**
      * The point evicted by the last call to {@link #update}, or null if the new
      * point was not accepted by the sampler.
      */
     private transient ISampled<Integer> evictedPoint;
+
     /**
      * This field is used to temporarily store the result from a call to
      * {@link #acceptPoint} for use in the subsequent call to {@link #addPoint}.
@@ -126,6 +137,7 @@ public class CompactSampler implements IStreamSampler<Integer> {
      * Visible for testing.
      */
     protected AcceptPointState acceptPointState;
+
     /**
      * If true, then the sampler will store sequence indexes along with the sampled
      * points.
@@ -209,7 +221,8 @@ public class CompactSampler implements IStreamSampler<Integer> {
      *                      satisfy the heap property.
      */
     public CompactSampler(int sampleSize, int size, double lambda, Random random, float[] weight, int[] pointIndex,
-            long[] sequenceIndex, boolean validateHeap) {
+            long[] sequenceIndex, boolean validateHeap, long mostRecentTimeStamp, long lastUpdateOflambda,
+            double accumulatedLambda) {
 
         checkNotNull(weight, "weight must not be null");
         checkNotNull(pointIndex, "pointIndex must not be null");
@@ -222,6 +235,9 @@ public class CompactSampler implements IStreamSampler<Integer> {
         this.weight = weight;
         this.pointIndex = pointIndex;
         this.sequenceIndex = sequenceIndex;
+        this.mostRecentTimeStamp = mostRecentTimeStamp;
+        this.lastUpdateOflambda = lastUpdateOflambda;
+        this.accumulatedLambda = accumulatedLambda;
 
         reheap(validateHeap);
     }
@@ -384,19 +400,23 @@ public class CompactSampler implements IStreamSampler<Integer> {
         while (randomNumber == 0d) {
             randomNumber = random.nextDouble();
         }
+        // mostRecentTimeStamp corresponds to the maximum sequenceIndex seen
         mostRecentTimeStamp = (mostRecentTimeStamp < sequenceIndex) ? sequenceIndex : mostRecentTimeStamp;
         return (float) (-(sequenceIndex - lastUpdateOflambda) * lambda - accumulatedLambda
                 + Math.log(-Math.log(randomNumber)));
     }
 
     /**
-     * sets the lambda on the fly
+     * sets the lambda on the fly. Note that the assumption is that the times stamps
+     * corresponding to changes to lambda and sequenceIndexes are non-decreasing --
+     * the sequenceIndexes can be out of order among themselves within two different
+     * times when lambda was changed.
      * 
      * @param newLambda the new sampling rate
      */
     @Override
     public void setLambda(double newLambda) {
-        accumulatedLambda += lastUpdateOflambda * lambda;
+        accumulatedLambda += (mostRecentTimeStamp - lastUpdateOflambda) * lambda;
         lambda = newLambda;
         lastUpdateOflambda = mostRecentTimeStamp;
     }
@@ -437,6 +457,18 @@ public class CompactSampler implements IStreamSampler<Integer> {
      */
     public double getLambda() {
         return lambda;
+    }
+
+    public double getAccumulatedLambda() {
+        return accumulatedLambda;
+    }
+
+    public long getLastUpdateOflambda() {
+        return lastUpdateOflambda;
+    }
+
+    public long getMostRecentTimeStamp() {
+        return mostRecentTimeStamp;
     }
 
     public boolean isStoreSequenceIndexesEnabled() {
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
@@ -15,15 +15,15 @@
 
 package com.amazon.randomcutforest.sampler;
 
-import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
-import static com.amazon.randomcutforest.CommonUtils.checkState;
-
 import java.util.List;
 import java.util.Optional;
 import java.util.Random;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
+
+import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
+import static com.amazon.randomcutforest.CommonUtils.checkState;
 
 /**
  * <p>
@@ -385,7 +385,7 @@ public class CompactSampler implements IStreamSampler<Integer> {
             randomNumber = random.nextDouble();
         }
         mostRecentTimeStamp = (mostRecentTimeStamp < sequenceIndex) ? sequenceIndex : mostRecentTimeStamp;
-        return (float) (-(sequenceIndex - lastUpdateOflambda) * lambda + accumulatedLambda
+        return (float) (-(sequenceIndex - lastUpdateOflambda) * lambda - accumulatedLambda
                 + Math.log(-Math.log(randomNumber)));
     }
 
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
@@ -59,7 +59,7 @@ import java.util.stream.Stream;
  * implementation uses less runtime memory.
  * </p>
  */
-public class CompactSampler implements IStreamSampler<Integer> {
+public class CompactSampler extends AbstractStreamSampler<Integer> {
 
     /**
      * When creating a {@code CompactSampler}, the user has the option to disable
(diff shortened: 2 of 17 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -484,6 +484,17 @@ public void setBoundingBoxCacheFraction(double cacheFraction) {
         updateExecutor.forEachTree(t -> t.setBoundingBoxCacheFraction(cacheFraction));
     }
 
+    /**
+     * changes the setting of time dependent sampling on the fly
+     * 
+     * @param lambda new value of sampling rate
+     */
+
```
wnbts, line 492: minor. no need for blank line

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -97,7 +97,19 @@
      * The decay factor used for generating the weight of the point. For greater
      * values of lambda we become more biased in favor of recent points.
      */
-    private final double lambda;
+    private double lambda;
+    /**
+     * The last timestamp when lambda was changed
+     */
+    private long lastUpdateOflambda = 0;
+    /**
```
wnbts, line 105: minor. a blank line separates variables.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -97,7 +97,19 @@
...
+     * most recent timestamp
+     */
+    private long mostRecentTimeStamp = 0;
```
wnbts, line 108: minor. the purpose of this variable is unclear. most recent time of sampling?

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -97,7 +97,19 @@
...
+    /**
+     * The accumulated sum of lambda before the last update
+     */
+    private float accumulatedLambda = 0;
     /**
```
wnbts, line 113: minor. a blank line separate variables.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -97,7 +97,19 @@
...
+     */
+    private long lastUpdateOflambda = 0;
+    /**
+     * most recent timestamp
+     */
+    private long mostRecentTimeStamp = 0;
+    /**
+     * The accumulated sum of lambda before the last update
+     */
+    private float accumulatedLambda = 0;
```
wnbts, line 112: question. is expected to reset the variable to 0 for deserialized models?

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -372,8 +384,21 @@ protected float computeWeight(long sequenceIndex) {
         while (randomNumber == 0d) {
             randomNumber = random.nextDouble();
         }
+        mostRecentTimeStamp = (mostRecentTimeStamp < sequenceIndex) ? sequenceIndex : mostRecentTimeStamp;
```
wnbts, line 387: question. when does mostrecentts increase to be greater than sequence index?

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
```diff
@@ -135,4 +135,11 @@ default boolean isFull() {
      * @return the number of points currently contained by the sampler.
      */
     int size();
+
+    /**
+     * changes the time dependent asmpling on the fly
```
wnbts, line 140: minor. typo

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
```diff
@@ -202,7 +214,21 @@ protected float computeWeight(long sequenceIndex) {
             randomNumber = random.nextDouble();
         }
 
-        return (float) (-sequenceIndex * lambda + Math.log(-Math.log(randomNumber)));
+        mostRecentTimeStamp = (mostRecentTimeStamp < sequenceIndex) ? sequenceIndex : mostRecentTimeStamp;
```
wnbts, line 217: minor. duplicated code.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: No - this is a bug! Fixing.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: mostRecentTimestamp corresponds to the maximum sequence index seen. The hidden assumption is that the sequence-indices and timestamps where lambda is changed are monotonically (non-decreasing) ordered. The sequence indices can be out-of-order among themselves. The assumption is that once lambda is updated, we just need to know the maximum sequence index seen so far. But a good callout - will add the comment.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
sudiptoguha: fixed!

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
sudiptoguha: Yes. These two classes should be refactored out of a common base class. Issue 139.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: The last change of lambda, which corresponds to maintaining the most recent time of sampling. Comments added.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: fixed

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: fixed

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
jotok: To put it another way, we want to be able to support a batch training mode where we don't want a strict requirement that points are submitted in timestamp order. However, for the computation for the accumulated lambda value used in the sampling weight to be correct, we need the following:
* it has to be computed using the maximum seen timestamp (not necessarily the most recently submitted timestamp)
* after the lambda value changes (i.e., after a call to `setLambda`, we can no longer accept points with a timestamp smaller than the current value of `mostRecentTimestamp`

With that in mind, I think we should add a validation to `acceptSample` to check that submitted timestamps are > `lastUpdateOfLambda`.

sudiptoguha: Done!

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
wnbts: the change is not in the current version

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
wnbts: is the difference of timestamp and sequenceindex necessary? or can they be named consistently? like sequenceIndex and largestSequenceIndex?

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/ArraySamplersToCompactStateConverter.java
```diff
@@ -83,6 +83,8 @@ public void addSampler(SimpleStreamSampler<double[]> sampler) {
         samplerState.setPointIndex(pointIndex);
         samplerState.setWeight(weight);
         samplerState.setSequenceIndex(sequenceIndex);
+        samplerState.setAccumulatedLambda(sampler.getAccumulatedLambda());
+        samplerState.setLastUpdateOfLambda(sampler.getLastUpdateOflambda());
```
wnbts, line 87: one field is missing.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
```diff
@@ -78,6 +79,9 @@ public CompactSamplerState toState(CompactSampler model) {
         state.setSize(model.size());
         state.setCapacity(model.getCapacity());
         state.setLambda(model.getLambda());
+        state.setLastUpdateOfLambda(model.getLastUpdateOflambda());
+        state.setAccumulatedLambda(model.getAccumulatedLambda());
+        state.setMostRecentTimeStamp(model.getMostRecentTimeStamp());
```
wnbts, line 84: discussion. For some clients, this feature might not be actually used and the fields could be unnecessary overhead from unused features. We might be able to design a way out for them. Or, such feature/enhancement-specific data will bloat the core model.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: largestSequenceIndexSeen makes sense. Will change.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
sudiptoguha: True, but the main point is that a versatile model is more useful. What is the core model of a sketch of data :) This is three numbers per sampler -- say 150 numbers overall and the gain is that the samplers can perform dynamically adjusted sampling. For time series application, that's a plus. One way to design would be to create version flags and serialization based on those version flags -- but that would also be a trap for users to get stuck with serialized models.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/ArraySamplersToCompactStateConverter.java
sudiptoguha: Fixing.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Should be fixed now.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
sudiptoguha: For example, we are only using samplers with the same parameters -- but there is no reason to do so! Maybe some trees focus on weekly variations and some focus on daily variations. 
Consider the version 1.0 ser/De where the timestamps were present (an option in the current version) -- the fact that those timestamps were present meant one could perform an insertion based on time stamps in deserialization. Sure, in this revamp when the trees can be serialized as well that information (of timestamps) is less critical. But that was handy in 1.0 -- because they were available.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -76,44 +76,68 @@
...
+
+    /**
+     * The last timestamp when lambda was changed
+     */
+    private long lastUpdateOflambda = 0;
+
+    /**
+     * most recent timestamp, used to determine lastUpdateOfLambda
+     */
+    private long largestSequenceIndexSeen = 0;
```
jotok, line 115: I might simplify this further and just go with `maxSequenceIndex`.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -76,44 +76,68 @@
...
      * The decay factor used for generating the weight of the point. For greater
      * values of lambda we become more biased in favor of recent points.
      */
-    private final double lambda;
+    private double lambda;
+
+    /**
+     * The last timestamp when lambda was changed
+     */
+    private long lastUpdateOflambda = 0;
```
jotok, line 110: How about `lambdaUpdateSequenceIndex`.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
```diff
@@ -69,7 +69,8 @@ public CompactSampler toModel(CompactSamplerState state, long seed) {
         }
 
         return new CompactSampler(state.getCapacity(), state.getSize(), state.getLambda(), new Random(seed), weight,
-                pointIndex, sequenceIndex, validateHeap);
+                pointIndex, sequenceIndex, validateHeap, state.getLargestSequenceIndexSeen(),
```
jotok, line 72: What if, as part of the mapping, we adjusted the sampler weights by the accumulated lambda value? Then I think we would only need to serialize the maximum sequence index field. When a new sampler is created from the state, last update time is set to maximum sequence index and accumulated lambda is set to 0.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
```diff
@@ -74,7 +74,19 @@
...
-    private final double lambda;
+    private double lambda;
+    /**
+     * The last timestamp when lambda was changed
+     */
+    private long lastUpdateOflambda = 0;
+    /**
+     * most recent timestamp
+     */
+    private long largestSequenceIndexSeen = 0;
```
jotok, line 85: This is hard-coding the assumption that sequence indexes will start at 0 and increase from there. That's consistent with what we do in the forest now, but we may need to think through this some more if we want to support something like using clock time as the index value.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
sudiptoguha: refactored.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: I think the most-recent should be there ... but modified

Java/core/src/main/java/com/amazon/randomcutforest/sampler/SimpleStreamSampler.java
sudiptoguha: Clock time would be a natural next step. I was thinking of fixing the restriction in the HashSet for the sequenceIndices first before going to milliseconds/etc. Probably a subsequent PR.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
```diff
@@ -135,4 +135,16 @@ default boolean isFull() {
...
+    /**
+     * changes the time dependent sampling on the fly
+     * 
+     * @param lambda the rate of sampling
+     */
+    void setLambda(double lambda);
+
+    void setMaxSequenceIndex(long maxSequenceIndex);
+
+    void setSequenceIndexOfMostRecentLambdaUpdate(long index);
```
jotok, line 148: I don't think `setLambda` and `setSequenceIndexOfMostRecentLambdaUpdate` should be in the interface, because lambda is a property of a particular sampling method.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/serialization/ProtostuffExampleWithDynamicLambda.java
```diff
@@ -0,0 +1,151 @@
...
+
+        double saveLambda = forest.getLambda();
+        forest.setLambda(10 * forest.getLambda());
+        forest2.setLambda(10 * forest2.getLambda());
+
+        for (int i = 0; i < numberOfTrees; i++) {
+            if (((CompactSampler) forest2.getComponents().get(i).getSampler())
+                    .getMaxSequenceIndex() != ((CompactSampler) forest.getComponents().get(i).getSampler())
+                            .getMaxSequenceIndex()) {
+                throw new IllegalStateException("Incorrect sampler state");
```
jotok, line 103: This looks like something that should be checked in a unit test. I think if we want to show something in an example, it should be something like:
* sample for a while with one lambda value
* show that certain points are anomalous
* "flush" the sampler by increasing the Lambda
* update the model with a new distribution of points
* show that the point that was considered an anomaly before changing lambda is no longer an anomaly

Java/testutils/src/main/java/com/amazon/randomcutforest/testutils/NormalMixtureTestData.java
```diff
@@ -47,6 +47,10 @@ public NormalMixtureTestData() {
         this(0.0, 1.0, 4.0, 2.0, 0.01, 0.3);
     }
 
+    public NormalMixtureTestData(double baseMu) {
+        this(baseMu, 1.0, 4.0, 2.0, 0.01, 0.3);
```
jotok, line 51: It seems a little weird to give the option to only change one of the mean values in the mixture distribution.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/dynamicconfiguration/DynamicSampling.java
```diff
@@ -0,0 +1,116 @@
...
+
+public class DynamicSampling implements Example {
+
+    public static void main(String[] args) throws Exception {
+        new DynamicSampling().run();
+    }
+
+    @Override
+    public String command() {
+        return "dynamic sampling";
```
jotok, line 32: this string can be used to activate the example from the command-line, so it shouldn't contain spaces.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -338,42 +326,34 @@ public void addPoint(Integer pointIndex) {
...
 
     /**
-     * @return the point evicted by the most recent call to {@link #update}, or null
-     *         if no point was evicted.
+     * removes the adjustments to weight in accumulated lambda and resets the
+     * updates to lambda
      */
-    public Optional<ISampled<Integer>> getEvictedPoint() {
-        return Optional.ofNullable(evictedPoint);
+    private void reset_weights() {
```
jotok, line 340: If you reset weights, shouldn't you also set the lambda update time to max sequence index?

jotok (approved):

Java/testutils/src/main/java/com/amazon/randomcutforest/testutils/NormalMixtureTestData.java
sudiptoguha: True :) Will update.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/serialization/ProtostuffExampleWithDynamicLambda.java
sudiptoguha: Maybe keep as a separate PR/issue? With the new refactoring of the samplers, all of the tests may need to be looked at.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
sudiptoguha: Ok. That makes sense. We may have to force object conversion in mappers - but let me try

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: That would face a check in update() and then move the "clock" to maxSequence index. It is possible that we want that behavior, but maybe have a more focused PR on just that change (along with potentially out of order updates).

Java/examples/src/main/java/com/amazon/randomcutforest/examples/dynamicconfiguration/DynamicSampling.java
sudiptoguha: fixed. Also fixed two other files where I overlooked the issue.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
sudiptoguha: The forest needs access to setLambda. Made it a default function to be overridden.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
wnbts: for design, serde could rely on some config to control which data to preserve. by default, the config should preserve the core model, which should also be the default forest config.

for example,
rcf = new rcf() // default
rcf_state = new mapper().map(rcf) // default data

rcf = new rcf().enableX().enableY()
rcf_state_with_x_y = new mapper().map(rcf) // state w/ because rcf config has x and y enabled
rcf_state_w/0_x_y = new mapper().disableX().disableY().map(rcf) // state w/o x and y because mapper has x and y explicitly disabled

The goal is to protect the model state from having to take on any data from any possibly used feature, many of which may just be some default values from never being used.

For this one, there may be enableDynamicLambda config at forest level. The default could be false, as the current behavior. If a client explicitly enables this, this operation will be allowed and the relevant data, used or not, will be mapped to state by (mapper) default. Otherwise, the state fields will be null-valued that serializer can skip.

sudiptoguha: Makes sense, I'm open to it. But I thought the issue was that the fields had to be always present (say for protostuff, which needed a schema - but I can be wrong). And if the field is present, then the issue is null vs default value. But yes, these can add up. Josh was looking into making the mapper efficient (copy of arrays) so maybe we have this for consideration there?

jotok: I'm open to adding configuration options to mappers (like `saveTreeState` and `saveExecutorContext`) but I'm not sure if the savings in this case is worth the additional logic. For 50 trees and sample size 256 we're currently getting JSON files around 1MB. In the latest revision, we're adding 2 long fields per sampler, so 800 bytes in the 50 tree case. If we want a mapper to be able to output a state class without those long fields, we would need two sampler state classes (one which includes them and one which doesn't), and we'd have to thread that logic through the RandomCutForestMapper, which is already a bit tangled from dealing with compact enabled/disabled, precision, optional tree state, etc.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/serialization/ProtostuffExampleWithDynamicLambda.java
jotok: We can revise this later, but I think examples need to be clear about what they're an example of. Giving an example where dynamically changing the lambda produces an interesting result is good. I know that I started this with my serialization examples, but now I'm questioning whether we should include assertion-like statements.

jotok (approved):

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
wnbts: since lambda is a public parameter core to rcf and there is a getter, a matching setter and its overheads are justifiable.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/IStreamSampler.java
```diff
@@ -135,4 +135,16 @@ default boolean isFull() {
...
      */
     int size();
+
+    /**
+     * changes the time dependent sampling on the fly. Lambda is the decay rate of
+     * every non-recent point
+     * 
+     * @param lambda the rate of sampling
+     */
+    default void setTimeDecay(double lambda) {
```
wnbts, line 145: minor. the names should be named timeDecay for consistency

wnbts (approved):

PR #143, opened 2021-02-12
smart copy

sudiptoguha:
*Description of changes:* Smarter copy, does not copy unused information in serialization, hence faster and smaller.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
@@ -17,6 +17,8 @@ package com.amazon.randomcutforest.store;
 
 import static com.amazon.randomcutforest.CommonUtils.checkState;
 
+import java.util.Arrays;
+
 /**
  * PointStore is a fixed size repository of points, where each point is a float
  * array of a specified length. A PointStore counts references to points that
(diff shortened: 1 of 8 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -141,7 +143,14 @@ protected void checkValidIndex(int index) {
         checkState(refCount[index] > 0, "ref count at occupied index is 0");
     }
 
+    public int getValidPrefix() {
+        int prefix = capacity;
+        while (prefix > 0 && !occupied.get(prefix - 1))
```
jotok, line 148: nit: use braces.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreFloat.java
```diff
@@ -135,7 +135,7 @@ public String toString(int index) {
     }
 
     public float[] getStore() {
-        return store;
+        return Arrays.copyOf(store, getValidPrefix() * dimensions);
```
jotok, line 138: I think it would be better to perform the truncation in the mapper class:
* most of the time, people don't expect business logic in getters
* since the logic to inflate the array appears in the mapper, it makes sense for the logic to truncate the array to be there too.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreFloat.java
sudiptoguha: Done

jotok (approved):

PR #144, opened 2021-02-12
Use -1 to represent missing root index in AbstractCompactRandomCutTree

### LINKED ISSUE #142: Serializing new tree causes NullPointerException when saveTreeState enabled
On the compact-rcf branch, if you try to map a newly created forest using a mapper with saveTreeState enabled, it will cause a NullPointerException

Example test:
```
    @Test
    public void npe() {
        int dimensions = 4;
        int numberOfTrees = 50;
        int sampleSize = 256;
        Precision precision = Precision.DOUBLE;

        RandomCutForest forest = RandomCutForest.builder().compactEnabled(true).dimensions(dimensions)
            .numberOfTrees(numberOfTrees).sampleSize(sampleSize).precision(precision).build();

        RandomCutForestMapper mapper = new RandomCutForestMapper();
        mapper.setSaveExecutorContext(true);
        mapper.setCopy(true);
        mapper.setSaveTreeState(true);

        RandomCutForestState state = mapper.toState(forest);
    }
```

Stacktrace:
```
java.lang.NullPointerException
	at com.amazon.randomcutforest.tree.AbstractCompactRandomCutTree.getRootIndex(AbstractCompactRandomCutTree.java:205)
	at com.amazon.randomcutforest.state.tree.CompactRandomCutTreeDoubleMapper.toState(CompactRandomCutTreeDoubleMapper.java:70)
	at com.amazon.randomcutforest.state.RandomCutForestMapper.lambda$toState$1(RandomCutForestMapper.java:169)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1384)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
	at com.amazon.randomcutforest.state.RandomCutForestMapper.toState(RandomCutForestMapper.java:169)
```

jotok:
Previously we were using null, which resulted in a NPE when getting the root index. After this change, an empty compact tree will have its root value set to -1, which breaks existing logic in `AbstractRandomCutTree`. Specifically, that class assumed that an empty tree would have a null root value. I added a method `isEmpty` to `AbstractRandomCutTree` so that this assumption can overridden.

I also changed the variable name `rootIndex` in `AbstractRandomCutTree` to `root`, since the "index" part only made sense for compact trees. I moved the getter method `getRoot` to `AbstractRandomCutTree` (it was previously in `AbstractCompactRandomCutTree`.

Closes #142

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeDoubleMapper.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeDoubleMapper.java
@@ -60,14 +60,14 @@ public class CompactRandomCutTreeDoubleMapper implements
         }
 
         return new CompactRandomCutTreeDouble(context.getMaxSize(), seed, (PointStoreDouble) context.getPointStore(),
-                leafStore, nodeStore, state.getRootIndex(), boundingBoxCacheEnabled);
+                leafStore, nodeStore, state.getRoot(), boundingBoxCacheEnabled);
 
     }
 
     @Override
     public CompactRandomCutTreeState toState(CompactRandomCutTreeDouble model) {
         CompactRandomCutTreeState state = new CompactRandomCutTreeState();
-        state.setRootIndex(model.getRootIndex());
+        state.setRoot(model.getRoot());
 
         if (model.getMaxSize() < SmallNodeStore.MAX_TREE_SIZE) {
             SmallLeafStoreMapper leafStoreMapper = new SmallLeafStoreMapper();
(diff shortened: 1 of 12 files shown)


jotok:
Thinking it over some more, I don't think this is a good fix to the issue. We're still using `null` in other places to represent missing values in compact trees. I think we need to be consistent about this.

PR #149, opened 2021-02-18
Compact rcf

sudiptoguha:
*Issue #, if available:* 148 (and exposing conditional fields)

*Description of changes:* removes the bug in bounding box computation after deserialization -- all of the bounding boxes are not computed for efficiency reasons.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -817,7 +817,8 @@ public class RandomCutForest {
      *                              missing values.
      * @return A point with the missing values imputed.
      */
-    public ArrayList<?> getSimpleConditionalField(double[] point, int numberOfMissingValues, int[] missingIndexes) {
+    public ArrayList<double[]> getSimpleConditionalField(double[] point, int numberOfMissingValues,
+            int[] missingIndexes) {
         checkArgument(numberOfMissingValues > 0, "numberOfMissingValues must be greater than or equal to 0");
 
         // We check this condition in traverseForest, but we need to check it here as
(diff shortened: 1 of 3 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -815,24 +817,20 @@ public DensityOutput getSimpleDensity(double[] point) {
      *                              missing values.
      * @return A point with the missing values imputed.
      */
-    public double[] imputeMissingValues(double[] point, int numberOfMissingValues, int[] missingIndexes) {
-        checkArgument(numberOfMissingValues >= 0, "numberOfMissingValues must be greater than or equal to 0");
+    public ArrayList<?> getSimpleConditionalField(double[] point, int numberOfMissingValues, int[] missingIndexes) {
```
jotok, line 820: I think it's a little ugly to have `ArrayList<?>` as the return type in this public API, requiring the user to cast the result in order to use it. We could have this method return `ArrayList<double[]>`, where the arrays would be single-element arrays when number of missing values is 1.

Java/core/src/test/java/com/amazon/randomcutforest/CompactRandomCutForestFunctionalTest.java
```diff
@@ -866,4 +868,30 @@ public void getTotalUpdates_returnExpectedSize() {
         }
         return shingledPoint;
     }
+
+    @Test
+    public void reproduce() {
```
jotok, line 873: Let's call this something like `testUpdateAfterRoundTrip()`.

sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -815,6 +817,36 @@ public DensityOutput getSimpleDensity(double[] point) {
      *                              missing values.
      * @return A point with the missing values imputed.
      */
+    public ArrayList<double[]> getSimpleConditionalField(double[] point, int numberOfMissingValues,
```
jotok, line 820: I think we need a better name for this method, but we can do it in another PR.

#151

jotok (approved):

PR #152, opened 2021-02-18
updated pointstore and shingle aware RCF

sudiptoguha:
*Issue #, if available:* addresses 138 and 150

*Description of changes:* updates RCF to be aware of shingling, and the pointstore uses that information to compress itself. The state/mapper classes also avoid null pointer based logic.

### DIFF
--- a/Java/benchmark/src/main/java/com/amazon/randomcutforest/StateMapperShingledBenchmark.java
+++ b/Java/benchmark/src/main/java/com/amazon/randomcutforest/StateMapperShingledBenchmark.java
@@ -1,0 +1,239 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest;
+
+import static java.lang.Math.PI;
+import static java.lang.Math.cos;
+
+import java.util.Random;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+import com.amazon.randomcutforest.config.Precision;
+import com.amazon.randomcutforest.profilers.OutputSizeProfiler;
+import com.amazon.randomcutforest.state.RandomCutForestMapper;
+import com.amazon.randomcutforest.state.RandomCutForestState;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.protostuff.LinkedBuffer;
+import io.protostuff.ProtostuffIOUtil;
+import io.protostuff.Schema;
+import io.protostuff.runtime.RuntimeSchema;
+
+@Warmup(iterations = 2)
+@Measurement(iterations = 5)
+@Fork(value = 1)
+@State(Scope.Benchmark)
+public class StateMapperShingledBenchmark {
+    public static final int NUM_TRAIN_SAMPLES = 2048;
+    public static final int NUM_TEST_SAMPLES = 50;
+
+    @State(Scope.Thread)
+    public static class BenchmarkState {
+        @Param({ "10" })
+        int dimensions;
+
+        @Param({ "50" })
+        int numberOfTrees;
+
+        @Param({ "256" })
+        int sampleSize;
+
+        @Param({ "false", "true" })
+        boolean saveTreeState;
+
+        @Param({ "SINGLE", "DOUBLE" })
+        Precision precision;
+
+        double[][] trainingData;
+        double[][] testData;
+        RandomCutForestState forestState;
+        String json;
+        byte[] protostuff;
+
+        @Setup(Level.Trial)
+        public void setUpData() {
+            trainingData = genShingledData(NUM_TRAIN_SAMPLES, dimensions, 0);
+            testData = genShingledData(NUM_TEST_SAMPLES, dimensions, 1);
+        }
+
+        @Setup(Level.Invocation)
+        public void setUpForest() throws JsonProcessingException {
+            RandomCutForest forest = RandomCutForest.builder().compactEnabled(true).dimensions(dimensions)
+                    .numberOfTrees(numberOfTrees).sampleSize(sampleSize).precision(precision).shingleSize(dimensions)
+                    .build();
+
+            for (int i = 0; i < NUM_TRAIN_SAMPLES; i++) {
+                forest.update(trainingData[i]);
+            }
+
+            RandomCutForestMapper mapper = new RandomCutForestMapper();
+            mapper.setSaveExecutorContext(true);
+            mapper.setSaveTreeState(saveTreeState);
+            forestState = mapper.toState(forest);
+
+            ObjectMapper jsonMapper = new ObjectMapper();
+            json = jsonMapper.writeValueAsString(forestState);
+
+            Schema<RandomCutForestState> schema = RuntimeSchema.getSchema(RandomCutForestState.class);
+            LinkedBuffer buffer = LinkedBuffer.allocate(512);
+            try {
+                protostuff = ProtostuffIOUtil.toByteArray(forestState, schema, buffer);
+            } finally {
+                buffer.clear();
+            }
+        }
+    }
+
+    private byte[] bytes;
+
+    @TearDown(Level.Iteration)
+    public void tearDown() {
+        OutputSizeProfiler.setTestArray(bytes);
+    }
+
+    @Benchmark
+    @OperationsPerInvocation(NUM_TEST_SAMPLES)
+    public RandomCutForestState roundTripFromState(BenchmarkState state, Blackhole blackhole) {
+        RandomCutForestState forestState = state.forestState;
+        double[][] testData = state.testData;
+
+        for (int i = 0; i < NUM_TEST_SAMPLES; i++) {
+            RandomCutForestMapper mapper = new RandomCutForestMapper();
+            mapper.setSaveExecutorContext(true);
+            mapper.setSaveTreeState(state.saveTreeState);
+            RandomCutForest forest = mapper.toModel(forestState);
+            double score = forest.getAnomalyScore(testData[i]);
+            blackhole.consume(score);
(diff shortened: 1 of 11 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -28,44 +28,224 @@
...
+                for (int i = 0; i < capacity; i++) {
+                    locationList[i] = i * dimensions;
+                }
+            }
+        } else {
+            baseDimension = dimensions;
+        }
+    }
+
     public PointStore(int dimensions, int capacity) {
```
jotok, line 116: Could you re-use the other constructor for this. E.g., is this the same as
```
public PointStore(int dimensions, int capacity) {
  this(dimensions, 1, capacity, false, true);
}
```

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -28,44 +28,224 @@
  * point values and increment and decrement reference counts. Valid index values
  * are between 0 (inclusive) and capacity (exclusive).
  */
-public abstract class PointStore<Point> extends IndexManager implements IPointStore<Point> {
+public abstract class PointStore<Store, P> extends IndexManager implements IPointStore<P> {
```
jotok, line 31: Why did we need to add this generic parameter? In the only existing subclasses, the types of Store and P are the same. What's the motivation to make these separate type parameters/

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -149,7 +343,84 @@ public int getValidPrefix() {
...
+    abstract void copyTo(int current, int running);
+
+    /**
+     * The following function eliminates redundant information that builds up in the
+     * pointstore and shrinks the pointstore
+     */
+
+    public void compact() {
+        checkArgument(!directLocationMap, "incorrect call; should not be used for direct location maps");
+        int DEFAULT_EMPTY = 0; // indicates location not in use
```
jotok, line 355: Should we make thisa class constant?

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -28,44 +28,224 @@
...
+            }
+            int nextIndex = takeIndex(); // no more compactions
+            locationList[nextIndex] = startOfFreeSegment;
+            copyPoint(point, 0, startOfFreeSegment, dimensions);
+            startOfFreeSegment += dimensions;
+            refCount[nextIndex] = 1; // has to be after compactions
+            return nextIndex;
+        }
+        /**
+         * the following corresponds to shingleSize=1 or the no-overlap case. The
```
jotok, line 237: This is more a matter of personal taste, but I might implement this method as
```
if (shingleSize > 1 && shingleAwareOverlapping) {
  addShingleAwareOverlapping(point);
} else {
  addSimple(point); // or addDirect maybe?
}
```
where `addShingleAwareOverlapping` and `addSimple` are private or protected methods.

Java/benchmark/src/main/java/com/amazon/randomcutforest/StateMapperShingledBenchmark.java
```diff
@@ -0,0 +1,239 @@
...
+        int i = 0;
+        for (int j = 0; j < shingleLength; ++j) {
+            double point = recentPointsSeen[(j + indexOfOldestPoint) % shingleLength];
+            shingledPoint[i++] = point;
+
+        }
+        return shingledPoint;
+    }
+
+    private static double[] getDataD(int num, double amplitude, double noise, long seed) {
```
jotok, line 229: What does this method name mean?

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
```diff
@@ -46,6 +60,34 @@
      * An array of reference counts for each stored point.
      */
     private short[] refCount;
+    /**
+     * is direct mapping enabled
+     */
+    @Getter
```
jotok, line 66: I think this is redundant since the class is annotated with `@Data`

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
```diff
@@ -46,6 +60,34 @@
...
+     */
+    private int[] locationList;
+    /**
+     * reverse location data to be usable in future
+     */
+    private int[] reverseLocationList;
+    /**
+     * flag to avoid null issues in the future
+     */
+    @Getter
```
jotok, line 79: Remove unnecessary annotation.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
```diff
@@ -46,6 +60,34 @@
...
+    private boolean reverseAvailable;
+    /**
+     * boolean indicating use of overlapping shingles; need not be used in certain
+     * cases
+     */
+    @Getter
```
jotok, line 85: Remove unnecessary annotation.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
```diff
@@ -46,6 +60,34 @@
...
+    private int[] locationList;
+    /**
+     * reverse location data to be usable in future
+     */
+    private int[] reverseLocationList;
+    /**
+     * flag to avoid null issues in the future
+     */
+    @Getter
+    private boolean reverseAvailable;
```
jotok, line 80: Is this flag to solve the problem with null fields in protostuff? I think that's something that we'll want to handle in a consistent way in all our state classes.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreFloat.java
```diff
@@ -115,27 +97,23 @@ public boolean pointEquals(int index, float[] point) {
...
-     */
     @Override
     public String toString(int index) {
         return Arrays.toString(get(index));
     }
 
-    public float[] getStore() {
-        return store;
+    @Override
+    void copyTo(int current, int running) {
```
jotok, line 115: nit: maybe change the formal method arguments to make the direction of the copy clear:
```
void copyTo(int dest, int source)
```

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -126,6 +126,11 @@
      */
     public static final boolean DEFAULT_CENTER_OF_MASS_ENABLED = false;
 
+    /**
+     * By default RCF is unaware of shingle size
+     */
+    public static final int DEFAULT_KNOWN_SHINGLESIZE = 1;
```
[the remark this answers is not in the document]
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -231,14 +240,20 @@ public RandomCutForest(Builder<?> builder) {
...
+                (shingleSize == 1));
+
         IUpdateCoordinator<Integer> updateCoordinator = new PointStoreCoordinator(tempStore);
         ComponentList<Integer> components = new ComponentList<>(numberOfTrees);
         for (int i = 0; i < numberOfTrees; i++) {
             ITree<Integer> tree = new CompactRandomCutTreeDouble(sampleSize, rng.nextLong(), tempStore,
                     boundingBoxCachingEnabled, centerOfMassEnabled, storeSequenceIndexesEnabled);
             IStreamSampler<Integer> sampler = new CompactSampler(sampleSize, lambda, rng.nextLong(),
                     storeSequenceIndexesEnabled);
+            // if (i==34)
```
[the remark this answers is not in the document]
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -231,14 +240,20 @@ public RandomCutForest(Builder<?> builder) {
     }
 
     private void initCompactDouble() {
-        PointStoreDouble tempStore = new PointStoreDouble(dimensions, sampleSize * numberOfTrees + 1);
+        int maxupdate = 1;
+        int compactReserve = sampleSize / 10;
+        int storeSize = numberOfTrees * sampleSize + maxupdate + compactReserve;
+        PointStoreDouble tempStore = new PointStoreDouble(dimensions, shingleSize, storeSize, (shingleSize > 1),
```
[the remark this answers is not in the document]
sudiptoguha: From the perspective of the forest, this is indeed odd. From the perspective of the pointstore, the shingleSize is used only if overlapping feature is to be used. This is the shingleSize>1, we are enforcing "use shingle size based compression wherever possible" -- we can separate this, but that would be a new flag/state variable -- we can do that in a separate PR. The shingleSize == 1 corresponds to use of directMap (do trees know the location of the point in the pointstore), we are hardcoding "yes". It is conceivable that we may want to use indirect map (if one wants to make changes to pointstore at runtime without disturbing the trees).

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: Currently we are using double[] or float[] as examples of the generic Point. We were using double[] and float[] for the store class with duplication of code. That duplication is avoided if Store is parametrized as well, that's the proximate reason. Otherwise this could be double[][], or P[] which would lose space efficiency.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
sudiptoguha: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreState.java
sudiptoguha: Yes. Looking ahead when the reverse array will be used. The nodeStore/LeafStore do not need these flags because they branch on the value of the capacity field. This may be relevant for the samplers.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreFloat.java
sudiptoguha: Done. Changed it to copy as a block.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreDouble.java
```diff
@@ -31,48 +30,40 @@
...
+    public PointStoreDouble(int dimensions, int capacity) {
+        super(dimensions, 1, capacity, false, true);
+        store = new double[capacity * dimensions];
+    }
+
+    @Override
+    boolean checkShingleAlignment(int location, double[] point) {
+        boolean test = true;
+        for (int i = 0; i < dimensions - baseDimension; i++) {
+            test = test && (point[i] == store[location - dimensions + baseDimension + i]);
```
jotok, line 59: We can short-circuit this logic.

sudiptoguha: Fixed.

jotok (approved):

PR #154, opened 2021-03-12
exposing sequence information to PointStore internally

sudiptoguha:
*Description of changes:* Add time sequence information to update pipeline and can enable out of order updates if desired.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -500,6 +500,22 @@ public class RandomCutForest {
     }
 
     /**
+     * Update the forest with the given point and a timestamp. The point is
+     * submitted to each sampler in the forest as if that timestamp was the correct
+     * stamp. storeSequenceIndexes must be false since the algorithm will not verify
+     * the correctness of the timestamp.
+     *
+     * @param point       The point used to update the forest.
+     * @param sequenceNum The timestamp of the corresponding point
+     */
+    public void update(double[] point, long sequenceNum) {
+        checkNotNull(point, "point must not be null");
+        checkArgument(point.length == dimensions, String.format("point.length must equal %d", dimensions));
+        checkArgument(!storeSequenceIndexesEnabled, "infeasible operation");
+        updateExecutor.update(point, sequenceNum);
+    }
+
+    /**
      * Update the forest such that each tree caches a fraction of the bounding
      * boxes. This allows for a tradeoff between speed and storage.
      *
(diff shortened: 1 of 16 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -499,6 +499,22 @@ public void update(double[] point) {
...
+     * stamp. storeSequenceIndexes must be false since the algorithm will not verify
+     * the correctness of the timestamp.
+     *
+     * @param point       The point used to update the forest.
+     * @param sequenceNum The timestamp of the corresponding point
+     */
+    public void update(double[] point, long sequenceNum) {
+        checkNotNull(point, "point must not be null");
+        checkArgument(point.length == dimensions, String.format("point.length must equal %d", dimensions));
+        checkArgument(!storeSequenceIndexesEnabled, "infeasible operation");
```
jotok, line 514: Is this because we haven't implemented a solution to the problem of duplicate timestamps?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -499,6 +499,22 @@ public void update(double[] point) {
...
+    /**
+     * Update the forest with the given point and a timestamp. The point is
+     * submitted to each sampler in the forest as if that timestamp was the correct
+     * stamp. storeSequenceIndexes must be false since the algorithm will not verify
+     * the correctness of the timestamp.
+     *
+     * @param point       The point used to update the forest.
+     * @param sequenceNum The timestamp of the corresponding point
+     */
+    public void update(double[] point, long sequenceNum) {
```
jotok, line 511: This solution will let us support out-of-order updates, but it's not extensible to uses like passing a label in with the point. Do we want to consider a more generic update method?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Yes. This restriction should go away once HashSet -> HashMap is done. However, another restriction may appear in its place if we enable internal shingling (which we should do to ease use). If we enable both internal shingling and out of order updates then a single update of a data value may spawn multiple shingled updates -- which the executors are currently not set up to do.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Yes - but perhaps not in this PR. Passing a label can be performed via constructing the label outside the forest and flattening everything into a vector. Also internal shingling may be a better (from broader usability point of view) first step for usability before categorical input.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
jotok: Let's give it a better error message (in this PR or a subsequent one). Maybe "This operation is not supported when sequence indexes are enabled".

jotok (approved):

PR #155, opened 2021-03-15
cleanup of point sum and allowing sequenceIndices to be non-unique

sudiptoguha:
*Issue #, if available:* 85 and 116

*Description of changes:* cleans up center of mass calculation for abstract trees and allows sequence indices to be non-unique

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -439,7 +439,7 @@ public class RandomCutForest {
      * reduce the memory size of the model at the cost of requiring double/float
      * conversions.
      *
-     * @return the desired precision to use internal to store points.
+     * @return the desired precision to use internally to store points.
      */
     public Precision getPrecision() {
         return precision;
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -240,9 +240,9 @@ public class RandomCutForest {
     }
 
     private void initCompactDouble() {
-        int maxupdate = 1;
+        int maxUpdate = 1;
         int compactReserve = sampleSize / 10;
-        int storeSize = numberOfTrees * sampleSize + maxupdate + compactReserve;
+        int storeSize = numberOfTrees * sampleSize + maxUpdate + compactReserve;
         PointStoreDouble tempStore = new PointStoreDouble(dimensions, shingleSize, storeSize, (shingleSize > 1),
                 (shingleSize == 1));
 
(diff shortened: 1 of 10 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
```diff
@@ -75,7 +76,7 @@
      * RandomCutTree, this set stores the indexes corresponding to times when the
      * given leaf point was added to the tree.
      */
-    private Set<Long> sequenceIndexes;
+    private HashMap<Long, Integer> sequenceIndexes;
```
jotok, line 79: minor: We can make the field type `Map` instead of the concrete type `HashMap`.

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
```diff
@@ -480,11 +490,12 @@ BoundingBox constructBoxInPlace(BoundingBox currentBox) {
         }
     }
 
-    void reComputePointSum(double[] point) {
+    void reComputePointSum() {
```
jotok, line 493: Let's write `recompute` instead of `reCompute` to be consistent.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -439,7 +439,7 @@ public boolean isStoreSequenceIndexesEnabled() {
      * reduce the memory size of the model at the cost of requiring double/float
      * conversions.
      *
-     * @return the desired precision to use internall to store points.
+     * @return the desired precision to use internal to store points.
```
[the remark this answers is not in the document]
sudiptoguha: makes sense

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -240,9 +240,9 @@ public RandomCutForest(Builder<?> builder) {
     }
 
     private void initCompactDouble() {
-        int maxupdate = 1;
+        int maxUpdate = 1;
```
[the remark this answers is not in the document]
sudiptoguha: Right now we are performing one update at a time. Thus we need exactly one additional space in point store. If for some reason we perform multiple ones (out of order shingled data, for example) then this would have to be changed. In general the three parts show how the capacity was arrived at. If we want to perform dynamic resizing, we may want viz into these different pieces.

Java/core/src/main/java/com/amazon/randomcutforest/store/IPointStoreView.java
```diff
@@ -27,6 +27,18 @@
...
+    /**
+     * useful for managing points, convex combinations, etc., e.g. needed for center
+     * of mass
+     * 
+     * @param index  identifier of the point
+     * @param factor multiplier
+     * @return the new point; or raises an exception if such an object cannot be
+     *         defined
+     */
+
```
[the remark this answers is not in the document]
sudiptoguha: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/tree/Node.java
sudiptoguha: Done.

PR #157, opened 2021-03-20
type fixes

sudiptoguha:
*Description of changes:* Fixes for types

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
@@ -20,8 +20,15 @@ import com.amazon.randomcutforest.executor.IUpdatable;
 import com.amazon.randomcutforest.sampler.IStreamSampler;
 import com.amazon.randomcutforest.tree.ITree;
 
-public interface IComponentModel<P> extends ITraversable, IUpdatable<P> {
-    ITree<P> getTree();
+/**
+ *
+ * @param <P> The internal point representation expected by the component models
+ *            in this list.
+ * @param <Q> The explicit data type of points being passed
+ */
+
+public interface IComponentModel<P, Q> extends ITraversable, IUpdatable<P> {
+    ITree<P, Q> getTree();
 
     IStreamSampler<P> getSampler();
 }
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -200,8 +200,8 @@ public class RandomCutForest {
      */
     protected final int threadPoolSize;
 
-    protected IUpdateCoordinator<?> updateCoordinator;
-    protected ComponentList<?> components;
+    protected IUpdateCoordinator<?, ?> updateCoordinator;
+    protected ComponentList<?, ?> components;
 
     /**
      * An implementation of forest traversal algorithms.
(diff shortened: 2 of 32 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
```diff
@@ -20,8 +20,15 @@
...
-public interface IComponentModel<P> extends ITraversable, IUpdatable<P> {
-    ITree<P> getTree();
+/**
+ *
+ * @param <P> The internal point representation expected by the component models
+ *            in this list.
+ * @param <Q> The explicit data type of points being passed
+ */
+
+public interface IComponentModel<P, Q> extends ITraversable, IUpdatable<P> {
```
jotok, line 30: We should be consistent about the type parameters throughout the library. I think either `Point` and `PointRef` everywhere, or `Point` and `PointIndex` (or `PointIx` if we want to be more abbreviated).

Do you remember why we needed to add `getTree` and `getSampler` here? When we created `ITraversable` and `IUpdatable` we wanted to keep the possibility that a traversable (or updatable) object could be something other than a sampler-plus-tree, but we broke that abstraction because it was complicating something else. Now that an `IComponentModel` explicitly defines getters for a tree and a sampler, it seems like the only possible implementation is `SamplerPlusTree`, and it makes me wonder if we need this interface at all. We could get rid of `IComponentModel`, `ITraversable`, and `IUpdatable`, and just use `SamplerPlusTree` everywhere.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -560,7 +560,7 @@ public void setLambda(double lambda) {
      * @return The aggregated and finalized result after sending a visitor through
      *         each tree in the forest.
      */
-    public <R, S> S traverseForest(double[] point, Function<ITree<?>, Visitor<R>> visitorFactory,
+    public <R, S> S traverseForest(double[] point, Function<ITree<?, ?>, Visitor<R>> visitorFactory,
```
jotok, line 563: We could create a new `VisitorFactory<R>` interface to act as a typedef for this:
```
public interface VisitorFactor<R> extends Function<ITree<?, ?>, Visitor<R>> {}
```

Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
sudiptoguha: Makes sense. We are using PointReference (integer/double[]), Point (actual value, double[]/float[]). 

Re memory :) -- I think the ITraversable is separate because it is how inference is done and is oblivious to any internal data structure/format. Every external interface is double[] at this point. Updates became complicated due to the need (and benefit) of coordination. Now the getters/setters are perhaps shortcuts of convenience? Maybe we need to (i) either define what the functions can do on the updateables and traversables separately (ii) eliminate and get to sampler plus tree.

One benefit of separating the traversal and updates is that traversal can be in parallel with no side effect issues; whereas updates are complicated (and need some coordination). It may be germane down the road -- but we can revisit this then as well provided none of this affects the state class (should not).

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: I like that.

Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
jotok: Per our discussion outside of this thread, we will remove `getSampler` and `getTree` from this interface, and come up with a more generic way to configure models dynamically at runtime.

PR #158, opened 2021-03-23
Add support for dynamically configurable models

jotok:
Add support for configuring component models by setting name. This
change allows us to add new configurable settings in a backwards-compatible way.

Different model implementations may support different settings. By supporting the generic `setConfig` method, we avoid adding implementation-specific methods to our base abstractions: `ITree`, `IStreamSampler`, and `IComponentModel`.


sudiptoguha (approved):
Nice!

PR #159, opened 2021-03-23
Compact rcf

sudiptoguha:
*Description of changes:* internal shingling


Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestState.java
```diff
@@ -48,8 +48,22 @@
 
     private boolean compactEnabled;
 
+    private boolean internalShinglingEnabled;
+
+    private boolean internalRotationEnabled;
+
+    private boolean dynamicResizingEnabled;
```
jotok, line 55: These are really properties of the point store state, right? Are we going to validate that the point store state contained in this forest state object has these properties? If not, then what purpose to these fields have?

Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestState.java
```diff
@@ -48,8 +48,22 @@
...
+
     private boolean centerOfMassEnabled;
 
+    private boolean saveTreeState;
+
+    private boolean saveSamplerState;
+
+    private boolean saveCoordinatorState;
+
+    private boolean singlePrecisionSet;
```
jotok, line 65: Why not use the Precision enum here? I don't think any serialization library will have trouble encoding an enum value.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreDouble.java
```diff
@@ -48,8 +49,19 @@ public PointStoreDouble(boolean overlapping, int startOfFreeSegment, int dimensi
...
 
     public PointStoreDouble(int dimensions, int capacity) {
-        super(dimensions, 1, capacity, false, true);
-        store = new double[capacity * dimensions];
+        this(new Builder().dimensions(dimensions).shingleSize(1).capacity(capacity).currentCapacity(capacity));
+    }
+
+    @Override
+    void resizeStore() {
+        int maxCapacity = rotationEnabled ? 2 * capacity : capacity;
```
jotok, line 57: Does this mean that we don't support resizing when rotation is enabled. Would it be better to have a resize operation that fails when resizing is not allowed than to have one that silently does nothing?

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreFloat.java
```diff
@@ -40,8 +41,19 @@ public PointStoreFloat(boolean overlapping, int startOfFreeSegment, int dimensio
...
 
     public PointStoreFloat(int dimensions, int capacity) {
-        super(dimensions, 1, capacity, false, true);
-        store = new float[capacity * dimensions];
+        this(new Builder().dimensions(dimensions).shingleSize(1).capacity(capacity).currentCapacity(capacity));
+    }
+
+    @Override
+    void resizeStore() {
+        int maxCapacity = rotationEnabled ? 2 * capacity : capacity;
```
jotok, line 49: Same question: why have this method silently do nothing when it's already at max capacity?

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallLeafStore.java
```diff
@@ -90,6 +90,12 @@ public int getPointIndex(int index) {
         return pointIndex[index];
     }
 
+    @Override
+    public int setPointIndex(int index, int pointIndex) {
+        this.pointIndex[index] = pointIndex;
+        return mass[index];
```
jotok, line 96: Is this method really supposed to return mass?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -229,6 +229,11 @@ int getPointReference(int index) {
         return leafStore.getPointIndex(getLeafIndexForTreeIndex(index));
     }
 
+    @Override
+    void switchLeafReference(Integer leafNode, Integer pointIndex) {
```
jotok, line 233: I think we can call this `setPointIndex` or `setLeafPointIndex` for consistency.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -270,75 +270,99 @@ void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
         updateAncestorPointSum(nodeReference);
     }
 
+    NodeReference findLeaf(Point point, boolean deleteMass) {
```
jotok, line 273: I wouldn't expect a method called `findFoo` to change an object's state. This method needs documentation and maybe a different name.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -270,75 +270,99 @@ void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
...
+                    + nodeReference + " node " + false + " Inconsistency in trees in delete step here.");
+        }
+        return nodeReference;
+    }
+
     /**
-     * method to delete a point from the tree
+     * the following function returns the number of copies of a point in the tree
+     * and switches the reference to the provided reference. This may be useful for
+     * collating duplicate points across trees. By default this is set to no-op.
```
jotok, line 299: The part in the comment about the default isn't true, because there's no default implementation of the method.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -270,75 +270,99 @@ void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
...
+    public void switchLeafReference(PointReference newRef) {
+        checkNotNull(newRef, " cannot be null ");
+        NodeReference nodeReference = findLeafAndVerify(getPointFromPointReference(newRef), false);
+        switchLeafReference(nodeReference, newRef);
+    }
+
+    public int getCopies(PointReference reference, boolean flagError) {
+        checkNotNull(reference, " cannot be null ");
+        NodeReference nodeReference = findLeaf(getPointFromPointReference(reference), false);
+        checkState(!flagError || getPointReference(nodeReference) == reference, "error");
```
jotok, line 316: We need a better error message.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -270,75 +270,99 @@ void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
...
-    public void deletePoint(PointReference pointReference, long sequenceNumber) {
-        checkState(root != null, "root must not be null");
-        deletePoint(root, getPointFromPointReference(pointReference), sequenceNumber, 0);
+    public void switchLeafReference(PointReference newRef) {
+        checkNotNull(newRef, " cannot be null ");
+        NodeReference nodeReference = findLeafAndVerify(getPointFromPointReference(newRef), false);
+        switchLeafReference(nodeReference, newRef);
+    }
+
+    public int getCopies(PointReference reference, boolean flagError) {
```
jotok, line 313: This method could use documentation explaining what kind of error is being checked for with `flagError`.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -270,75 +270,99 @@ void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
...
+        Point point = getPointFromPointReference(pointReference);
+        NodeReference nodeReference = findLeafAndVerify(point, true);
+        if (enableSequenceIndices) {
+            deleteSequenceIndex(nodeReference, sequenceNumber);
+        }
+        PointReference returnVal = getPointReference(nodeReference);
 
-            NodeReference parent = getParent(nodeReference);
+        // decrease mass for the delete
+        if (decrementMass(nodeReference) > 0) {
```
jotok, line 341: Isn't the mass being decremented in `findLeafAndVerify` now? Given that someone can try to delete a point that doesn't exist, I think it makes sense to do the mass adjustment after finding the point (which I think is what we had before).

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -270,75 +270,99 @@ void updateAncestorNodesAfterDelete(NodeReference nodeReference, Point point) {
...
+                    : getRightChild(nodeReference);
+        }
+        return nodeReference;
+    }
+
+    NodeReference findLeafAndVerify(Point point, boolean deleteMass) {
+        NodeReference nodeReference = findLeaf(point, deleteMass);
+        Point oldPoint = getPointFromLeafNode(nodeReference);
+        if (!equals(oldPoint, point)) {
+            throw new IllegalStateException(toString(point) + " " + toString(getPointFromLeafNode(nodeReference)) + " "
```
jotok, line 290: Won't this leave the tree in an invalid state if `deleteMass` is true?

Java/core/src/main/java/com/amazon/randomcutforest/tree/HyperTree.java
```diff
@@ -123,7 +123,8 @@ private Cut getCut(IBoundingBoxView bb, Random ring, Function<IBoundingBoxView,
     }
 
     @Override
-    public void deletePoint(double[] point, long sequenceIndex) {
+    public double[] deletePoint(double[] point, long sequenceIndex) {
+        return point;
         // () -> /dev/null
```
jotok, line 128: I think we can remove this comment.

Java/core/src/main/java/com/amazon/randomcutforest/IStateCoordinator.java
```diff
@@ -13,20 +13,21 @@
...
- * by the updatable models, and processes the list of deleted points if needed.
- * An IUpdateCoordinator can be used to manage shared state.
+ * An IStateCoordinator coordinates the updates as well as the translation of
+ * the internal representation to explicit points during inference.
  *
  * @param <PointReference> An internal point representation.
  * @param <Point>          Explicit point type
  */
-public interface IUpdateCoordinator<PointReference, Point> {
+public interface IStateCoordinator<PointReference, Point> {
```
jotok, line 30: I'd like to revisit this name change. This interface defines methods  that are invoked before and after individual trees are updated, and it allows the Update Executor to sequence operations in a way that keeps the global state consistent. So I think "Update Coordinator" is not a bad name for this interface, even if we add the `getStore` method. What other operations are we thinking of using this interface for that would make the more general name "State Coordinator" a better fit?

Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestState.java
sudiptoguha: Fair point. The latter two are indeed details of pointStore and will move them there. Internal shingling however requires visibility from the forest level class -- we should decide whether to augment/or nor to, without a roundtrip. Moreover this flag would enable checks at the input level of the forest.

Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestState.java
sudiptoguha: There are only two choices of the precision, we can go either way. Enums can help accommodate future point types -- but we should make is a point type enum?

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreDouble.java
sudiptoguha: No, we do support resizing when rotation is enabled :)  Internally, we allow 2 times the (externally stated) capacity to allow for internally rotated shingling -- that is an artifact of this specific implementation, which allows better compression as a tradeoff. The issue is that a specific tuple cannot be copied to any location without losing the inherent intent in the position.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStoreFloat.java
sudiptoguha: This is just a mirror of the double[] case. We are doing something -- if rotation is enabled, this implementation allows itself to go to twice the stated capacity.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallLeafStore.java
sudiptoguha: It's a useful shortcut which avoids asking for the mass a second time. But not really needed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done.

sudiptoguha: However the intended use is in AbstractRandomCutTree .. so maybe setLeafPointReference (instead of index)

Java/core/src/main/java/com/amazon/randomcutforest/IStateCoordinator.java
sudiptoguha: The main explicit task of the coordinator at the moment is to facilitate updates and that updates are not leaving the tree in an unusable "state".  However the way it is achieving this is by ensuring coordination between a sampler and a tree. However the point store is used in inference as well -- the trees only have a read only interface. One option is to make the point store forest level and have some glue (but any such glue would look this class) or have a state coordinator. Any "change of state" operation would need the point store and coordination (change the sample size parameter, change/reduce the shingle size parameter -- both of which has been asks).  Each such feature would need more coordination between samplers and trees. "Update" may not be the full description of the coordination required. But names other than StateCoordinator is welcome. PointCooordinator? We could also just call it Coordinator. From a constructor perspective, it makes sense to pass: trees, samplers, executioner's. context and a coordinator.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: In the current code, we should never try to delete a point which is not in the tree. But in future, we may want to. Changed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: Not in the current setup - but someone may wish to invoke deletePoint when the point is not in the tree. Changed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/HyperTree.java
sudiptoguha: Done. Removed comment from the corresponding add as well.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: Changed the method to avoid error flags etc.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: Done. Flag and error removed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: fixed

Java/core/src/main/java/com/amazon/randomcutforest/executor/AbstractForestUpdateExecutor.java
```diff
@@ -58,14 +58,15 @@ protected AbstractForestUpdateExecutor(IUpdateCoordinator<PointReference, Point>
...
-        update(point, updateCoordinator.getTotalUpdates());
+        update(point, stateCoordinator.getTotalUpdates());
     }
 
     public void update(double[] point, long sequenceNumber) {
-        double[] pointCopy = cleanCopy(point);
-        PointReference updateInput = updateCoordinator.initUpdate(pointCopy, sequenceNumber);
-        List<UpdateResult<PointReference>> results = update(updateInput, sequenceNumber);
-        updateCoordinator.completeUpdate(results, updateInput);
+        double[] pointCopy = RandomCutForest.cleanCopy(point);
```
jotok, line 65: Given that the `cleanCopy` method is in the forest, why don't we call it in the forest method before passing the point here?

Java/core/src/main/java/com/amazon/randomcutforest/executor/AbstractForestUpdateExecutor.java
```diff
@@ -58,14 +58,15 @@ protected AbstractForestUpdateExecutor(IUpdateCoordinator<PointReference, Point>
...
+        PointReference updateInput = stateCoordinator.initUpdate(pointCopy, sequenceNumber);
+        List<UpdateResult<PointReference>> results = (updateInput == null) ? null : update(updateInput, sequenceNumber);
```
jotok, line 67: When would `updateInput` be null?

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
```diff
@@ -21,31 +21,34 @@
...
 
     @Override
     public void completeUpdate(List<UpdateResult<Integer>> updateResults, Integer updateInput) {
-        updateResults.forEach(result -> {
-            result.getAddedPoint().ifPresent(store::incrementRefCount);
-            result.getDeletedPoint().ifPresent(store::decrementRefCount);
-        });
-        store.decrementRefCount(updateInput);
-        totalUpdates++;
+        if (updateInput != null) {
```
jotok, line 41: Now that we've added a method `getStore` with a null implementation, and we're add support for do-nothing implementations of `initUpdate` and `completeUpdate`, it seems like we need to rethink the entire update-coordinator paradigm. If it's not a universal paradigm that we can use to define different update methods, then it's just adding noise to the library.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -152,53 +186,14 @@ public CompactSampler(int sampleSize, double lambda, Random random, boolean stor
         }
         this.random = random;
         this.lambda = lambda;
+        this.initialAcceptFraction = initialAcceptFraction;
```
jotok, line 189: Can we do something like:
```
double p = 1 - size() / capacity();
if (Math.random() <= p) {
  // accept point
} else {
  // use time-decayed sampling as before
}
```

This seems conceptually nicer to me and it doesn't require an explicit parameter. The potential downside is that every sample will accept the first point that it sees, and will accept the first few points with high probability, but they should start to diverge before the samplers are full.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -284,24 +292,26 @@ public void reheap(boolean validate) {
...
-        }
-        int current = size++;
-        while (current > 0) {
-            int tmp = (current - 1) / 2;
-            if (this.weight[tmp] < this.weight[current]) {
-                swapWeights(current, tmp);
-                current = tmp;
-            } else
-                break;
+        if (pointIndex != null) {
```
jotok, line 295: Why would we try to add null?

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -403,4 +417,132 @@ private void swapWeights(int a, int b) {
...
+        this.storeSequenceIndexesEnabled = builder.storeSequenceIndexesEnabled;
+        this.initialAcceptFraction = builder.initialAcceptFraction;
+        this.random = builder.random;
+        this.lambda = builder.lambda;
+        this.maxSequenceIndex = builder.maxSequenceIndex;
+        this.sequenceIndexOfMostRecentLambdaUpdate = builder.sequenceIndexOfMostRecentLambdaUpdate;
+
+        if (builder.weight != null || builder.pointIndex != null || builder.weight != null
+                || builder.sequenceIndex != null || builder.validateHeap) {
+            checkArgument(builder.weight.length == builder.capacity, " incorrect state");
```
jotok, line 527: These error messages won't be very helpful to users.

Java/core/src/main/java/com/amazon/randomcutforest/executor/AbstractForestUpdateExecutor.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/executor/AbstractForestUpdateExecutor.java
sudiptoguha: For the initial points in internal shingling. Added comments.

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
sudiptoguha: The null check here is enable (i) count the updates only when they are completed (ii) count the updates for all the points in case of the internal shingling -- so that we can check that they are in sequential increasing order, increasing by 1. We can remove the null check if we relax either (i) and (ii). The real issue is that for internal shingling we will not generate a point (hence any non-null point to update) while the initial shingle is being filled.

Java/core/src/main/java/com/amazon/randomcutforest/executor/AbstractForestUpdateExecutor.java
sudiptoguha: null check made explicit here.

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
sudiptoguha: removing this check since now update executor checks for null for shingling issue

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: removed.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallLeafStore.java
sudiptoguha: It's actually helpful to return the older stored index (to audit).

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: This specific notion of p would mean that the sampler is full after capacity * ln (capacity) number of elements in expectation. We can use this formula, but add the constant boost. That preserves the intent, but also ensures that the samplers are full before capacity/initialFracton (say 0.25 would mean 4 * capacity).

sudiptoguha:
Breaking PR into smaller pieces.

PR #161, opened 2021-04-07
updating samplers

sudiptoguha:
*Issue #, if available:*

*Description of changes:* updating sampler behavior

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
@@ -15,18 +15,18 @@
 
 package com.amazon.randomcutforest.sampler;
 
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+import static com.amazon.randomcutforest.CommonUtils.checkState;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_INITIAL_ACCEPT_FRACTION;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_SAMPLE_SIZE;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_STORE_SEQUENCE_INDEXES_ENABLED;
+
 import java.util.List;
 import java.util.Optional;
 import java.util.Random;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
-
-import static com.amazon.randomcutforest.CommonUtils.checkArgument;
-import static com.amazon.randomcutforest.CommonUtils.checkState;
-import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_INITIAL_ACCEPT_FRACTION;
-import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_SAMPLE_SIZE;
-import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_STORE_SEQUENCE_INDEXES_ENABLED;
 
 /**
  * <p>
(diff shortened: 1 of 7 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -403,4 +415,132 @@ private void swapWeights(int a, int b) {
...
+
+        public CompactSampler build() {
+            return new CompactSampler(this);
+        }
+    }
+
+    public CompactSampler(Builder<?> builder) {
+        // int sampleSize, double lambda, Random random, boolean
+        // storeSequenceIndexesEnabled,
+        // double initialAcceptFraction) {
```
jotok, line 510: nit: delete commented code

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -104,6 +107,11 @@
      */
     private final boolean storeSequenceIndexesEnabled;
 
+    /**
+     * controls the behavior of the initial set of points
```
jotok, line 111: We need better documentation for this parameter. Maybe you could just copy the description from the constructor.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -223,21 +218,32 @@ public boolean acceptPoint(long sequenceIndex) {
...
-            return false;
         }
+        return false;
+    }
+
+    /**
+     * evicts the maximum weight point from the sampler. can be used repeatedly to
+     * change the size of the sampler and associated tree
+     */
+
```
jotok, line 236: nit: remove empty line

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: done

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -15,16 +15,19 @@
...
 import java.util.Random;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+import static com.amazon.randomcutforest.CommonUtils.checkState;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_INITIAL_ACCEPT_FRACTION;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_SAMPLE_SIZE;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_STORE_SEQUENCE_INDEXES_ENABLED;
```
jotok: If we need to have default values both here and in RandomCutForest, then I think the dependency should go the other way: RandomCutForest should import that constants from this class. That said, I don't think CompactSampler should have a default sample size, because having a default sample size makes sense in the context of a Random Cut Forest, but not for a sampler in general. Similarly, if a default initial accept fraction is set in this class, then I think it should either be 0 or 1. Any other choice seems like it would be application dependent, and hence doesn't belong in the sampler class.

Alternatively, we don't have to set defaults in this class at all, and we can require that the parameters be set explicitly when creating a sampler.

sudiptoguha: I'm ok either way. Maybe clean up all the builder related options in one go (there will be multiple classes with these kinds of issues).

jotok: Sure, can you create an issue for tracking?

PR #163, opened 2021-04-07
changing small stores

sudiptoguha:
*Description of changes:* changes the implementation of the smaller stores by eliminating the need to explicitly maintain the free indexes.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
@@ -15,6 +15,8 @@
 
 package com.amazon.randomcutforest.store;
 
+import java.util.Map;
+
 /**
  * An interface for describing access to the node ntore for different types of
  * trees.
@@ -78,12 +80,28 @@ public interface INodeStore {
     int getRightIndex(int index);
 
     /**
+     * sets rightChild
+     *
+     * @param index    node
+     * @param newIndex new right child
+     */
+    void setRightIndex(int index, int newIndex);
+
+    /**
      * gets leftChild
      * 
      * @param index node
      * @return index of left child (can be leaf)
      */
     int getLeftIndex(int index);
+
+    /**
+     * sets leftChild
+     *
+     * @param index    node
+     * @param newIndex new left child
+     */
+    void setLeftIndex(int index, int newIndex);
 
     /**
      * increments mass of node by 1 and returns the new value
@@ -123,10 +141,19 @@ public interface INodeStore {
     /**
      * returns the current mass (number of nodes in he subtree under node)
      * 
-     * @param index
-     * @return
+     * @param index node
+     * @return the mass of the node
      */
     int getMass(int index);
+
+    /**
+     * sets the current mass (number of nodes in he subtree under node) useful for
+     * rebuilding a tree
+     *
+     * @param index   node
+     * @param newMass the new mass of the node
+     */
+    void setMass(int index, int newMass);
 
     /**
      * increases the mass of node as well as all its ancestors by 1 note that all
@@ -138,13 +165,31 @@ public interface INodeStore {
     void increaseMassOfAncestorsAndItself(int index);
 
     /**
+     * decreases the mass of node as well as all its ancestors by 1 note that all
+     * these nodes are internal nodes and we eliminate the back and forth in a
+     * single call
+     *
+     * @param index node
+     */
+    void decreaseMassOfAncestorsAndItself(int index);
+
+    /**
      * returns the sibling of node, both of whom save the same parent
      * 
      * @param parent the parent
      * @param node   the node
      * @return the sibling
      */
-
     public int getSibling(int parent, int node);
+
+    /**
+     * the following returns a map of (leaves,correponding parent) pairs to enable
+     * validation in serialization
+     *
+     * @return map of leaf,corresponding pairs in use for the node store
+     */
+    default Map<Integer, Integer> getLeavesAndParents() {
+        return null;
+    }
 
 }
--- a/Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
@@ -162,7 +162,7 @@ public interface INodeStore {
      * 
      * @param index node
      */
-    void increaseMassOfAncestorsAndItself(int index);
+    void increaseMassOfSelfAndAncestors(int index);
 
     /**
      * decreases the mass of node as well as all its ancestors by 1 note that all
@@ -171,7 +171,7 @@ public interface INodeStore {
      *
      * @param index node
      */
-    void decreaseMassOfAncestorsAndItself(int index);
+    void decreaseMassOfSelfAndAncestors(int index);
 
     /**
      * returns the sibling of node, both of whom save the same parent
--- a/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
@@ -30,6 +30,9 @@ import static com.amazon.randomcutforest.CommonUtils.checkArgument;
  */
 public abstract class PointStore<Store, Point> extends IndexManager implements IPointStore<Point> {
 
+    public static int INFEASIBLE_LOCATION = -1;
+
+    public static int INFEASIBLE_INDEX = -1;
     /**
      * generic store class
      */
--- a/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
@@ -30,9 +30,9 @@ import static com.amazon.randomcutforest.CommonUtils.checkArgument;
  */
 public abstract class PointStore<Store, Point> extends IndexManager implements IPointStore<Point> {
 
-    public static int INFEASIBLE_LOCATION = -1;
+    public static int INFEASIBLE_POINTSTORE_LOCATION = -2;
 
-    public static int INFEASIBLE_INDEX = -1;
+    public static int INFEASIBLE_POINTSTORE_INDEX = -1;
     /**
      * generic store class
      */
--- a/Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
@@ -19,36 +19,54 @@ import static com.amazon.randomcutforest.CommonUtils.checkArgument;
 import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
 import static com.amazon.randomcutforest.CommonUtils.checkState;
 
+import java.util.Arrays;
 import java.util.BitSet;
 import java.util.HashSet;
 import java.util.Set;
 
 /**
  * This class defines common functionality for Store classes, including
- * maintaining the stack of free pointers.
+ * maintaining the stack of free pointers. The stack may be implicitly defined.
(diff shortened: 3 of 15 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -30,6 +30,9 @@
  */
 public abstract class PointStore<Store, Point> extends IndexManager implements IPointStore<Point> {
 
+    public static int INFEASIBLE_LOCATION = -1;
+
+    public static int INFEASIBLE_INDEX = -1;
```
jotok, line 35: If we're going to have 2 different constants, should we assign them different values? Or maybe we only need one constant?

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -19,36 +19,54 @@
...
 import java.util.BitSet;
 import java.util.HashSet;
 import java.util.Set;
 
 /**
  * This class defines common functionality for Store classes, including
- * maintaining the stack of free pointers.
+ * maintaining the stack of free pointers. The stack may be implicitly defined.
+ * if freeIndexPointer exceeds freeIndexes.length, then any intermediate
+ * location say i, must contain entry capacity - i - 1
```
jotok, line 31: These are useful implementation details, but I don't think that they will help people understand how to use the class. Maybe we could move this to an internal comment somewhere.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -19,36 +19,54 @@
...
-     */
     public SmallIndexManager(short capacity) {
-        this.capacity = capacity;
-        freeIndexes = new short[capacity];
+        this(capacity, null);
+    }
 
-        for (int j = 0; j < capacity; j++) {
-            freeIndexes[j] = (short) (capacity - j - 1); // reverse order
+    public SmallIndexManager(short capacity, BitSet bits) {
```
jotok, line 44: Needs documentation.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -75,6 +93,39 @@ public SmallIndexManager(short[] freeIndexes, short freeIndexPointer) {
         }
     }
 
+    public SmallIndexManager(int capacity, short[] freeIndexes, short freeIndexPointer) {
```
jotok, line 96: Do we still need this constructor? We should delete unnecessary constructors before they get baked into the public API.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -75,6 +93,39 @@ public SmallIndexManager(short[] freeIndexes, short freeIndexPointer) {
...
+            if (i < freeIndexes.length) {
+                occupied.clear(freeIndexes[i]);
+            } else {
+                occupied.clear(capacity - i - 1);
+            }
+        }
+
+    }
+
+    public SmallIndexManager(int capacity, int oldCapacity) {
```
jotok, line 116: Could we do this in a public static method that invokes `SmallIndexManager(capacity, occupied)`?

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -106,16 +157,22 @@ public short size() {
     }
 
     /**
-     * Take an index from the free index stack.
+     * Take an index from the free index stack. Note that the stack can be implicit.
```
jotok, line 160: I think notes about how free indexes are computed should be inline comments instead of method documentation. The person calling this method shouldn't have to worry about how the next free index is obtained.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -125,8 +182,26 @@ protected short takeIndex() {
...
+            } else {
+                // for most of the applications of this small sampler we would need one extra
+                // entry; but freeIndexPointer can be -1 as well
+                // this can be changed to a larger value to reduce repeated
+                // allocation/deallocation
+                int cap = Math.min(capacity, freeIndexPointer + 2);
+                int oldLength = freeIndexes.length;
+                freeIndexes = Arrays.copyOf(freeIndexes, cap);
+                for (int j = oldLength; j < cap && j < freeIndexPointer + 1; j++) {
+                    freeIndexes[j] = (short) (capacity - j - 1);
```
jotok, line 201: Why are we populating these values? Won't they be overwritten by future calls to `releaseIndex`? When someone calls `releaseIndex` for a given index, does that imply that they already released these other indexes so they should be added to the free index stack?

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
```diff
@@ -125,8 +182,26 @@ protected short takeIndex() {
...
+                ++freeIndexPointer;
+                return;
+            } else {
+                // for most of the applications of this small sampler we would need one extra
+                // entry; but freeIndexPointer can be -1 as well
+                // this can be changed to a larger value to reduce repeated
+                // allocation/deallocation
+                int cap = Math.min(capacity, freeIndexPointer + 2);
+                int oldLength = freeIndexes.length;
+                freeIndexes = Arrays.copyOf(freeIndexes, cap);
```
jotok, line 199: Now that we're doing dynamic array resizing, I wonder if it would make sense to just use a Java List or something as a stack, and then we don't have to write the bookkeeping code.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallLeafStore.java
```diff
@@ -46,23 +50,30 @@ public SmallLeafStore(short capacity) {
...
-        checkArgument(parentIndex.length == capacity && mass.length == capacity && freeIndexes.length == capacity,
+        checkArgument(pointIndex.length == capacity && parentIndex.length == capacity && mass.length == capacity,
                 "all array arguments must have the same length");
 
         this.pointIndex = pointIndex;
         this.parentIndex = parentIndex;
         this.mass = mass;
     }
 
+    public SmallLeafStore(int capacity, BitSet leafbits) {
```
jotok, line 68: Needs documentation.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallNodeStore.java
```diff
@@ -153,15 +168,71 @@ public int getMass(int index) {
...
     public int getSibling(int parent, int node) {
         return leftIndex[parent] == (short) node ? rightIndex[parent] : leftIndex[parent];
     }
 
+    @Override
+    public Map<Integer, Integer> getLeavesAndParents() {
+        HashMap<Integer, Integer> newMap = new HashMap<>();
+        for (int i = 0; i < capacity; i++) {
+            if (occupied.get(i) && leftIndex[i] >= capacity) {
+                newMap.put((int) leftIndex[i], i);
```
jotok, line 202: Did Java complain about this? I thought it would promote a short to an int automatically.

Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
```diff
@@ -137,14 +164,32 @@
...
     public int getSibling(int parent, int node);
 
+    /**
+     * the following returns a map of (leaves,correponding parent) pairs to enable
+     * validation in serialization
+     *
+     * @return map of leaf,corresponding pairs in use for the node store
+     */
+    default Map<Integer, Integer> getLeavesAndParents() {
+        return null;
```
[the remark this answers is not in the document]
sudiptoguha: It temporary - will go away once we update the other nodestore classes.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: Two constants (with different negative values) sound like a good idea. Using the same constant becomes problematic in sense of type checking.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
```diff
@@ -137,14 +164,32 @@
...
     void increaseMassOfAncestorsAndItself(int index);
 
+    /**
+     * decreases the mass of node as well as all its ancestors by 1 note that all
+     * these nodes are internal nodes and we eliminate the back and forth in a
+     * single call
+     *
+     * @param index node
+     */
+    void decreaseMassOfAncestorsAndItself(int index);
```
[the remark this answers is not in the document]
sudiptoguha: Names changed. Let's defer the default implementation for now -- there are two classes that use this, primarily separated due to the base data type of short[] versus int []. But this is definitely feasible.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: This should be limited to the mapper classes. The bitset constructor makes sense when the external entity does not have information about the stack. The stack and the pointer still saves over the bit set in the serialized model. But it may be worth looking at this option once the entirety of changes to the mappers are in.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: Absolutely. Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: possible. Lets switch once the other changes are in.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallLeafStore.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallNodeStore.java
sudiptoguha: Yes, it suggests the cast.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallIndexManager.java
sudiptoguha: We are populating them because the newly added "index" breaks the implicit guarantee -- and if we add the break to the pattern then we need to ensure that the intermediate values are filled in as well. We are adding the index to the stack at the last line of the method.

jotok (approved):

PR #167, opened 2021-04-14
reducing serialized space

sudiptoguha:
*Description of changes:* avoiding storing information which is redundant in the forest


Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestState.java
```diff
@@ -48,8 +50,18 @@
...
+
     private boolean centerOfMassEnabled;
 
+    private boolean saveTreeState;
+
+    private boolean saveSamplerState;
+
+    private boolean saveCoordinatorState;
+
+    private boolean singlePrecisionSet;
```
jotok, line 63: I would either use the Precision enum or a String value here.

Java/core/src/main/java/com/amazon/randomcutforest/store/SmallLeafStore.java
```diff
@@ -37,20 +37,17 @@
  * The reference to the nodes will be offset by an amount that this ihe
  * capacity.
  */
-public class SmallLeafStore extends SmallIndexManager implements ILeafStore {
+public class SmallLeafStore extends IndexManager implements ILeafStore {
```
jotok, line 40: Did we want to update these classes as well to has-an IndexManager from is-an Index manager?

sudiptoguha: Great question -- we probably should. Given that we had to do that for PointStore and is-an became incredibly cumbersome from a constructor point of view, it is likely that down the road we will have to do the same for NodeStore.
LeafStores may not be needed at all even.

sudiptoguha:
Refactoring and removing leaf store as well as multiple files managing the leafstore.

PR #168, opened 2021-04-14
Compact rcf

sudiptoguha:
*Description of changes:* removes leafstore, simplifies mappings and saves space


Java/core/src/main/java/com/amazon/randomcutforest/state/store/NodeStoreMapper.java
```diff
@@ -33,34 +37,132 @@
...
+
         return state;
     }
+
+    boolean reduceToBits(int size, int[] leftIndex, int[] rightIndex) {
+        boolean check = leftIndex.length == rightIndex.length;
+        int leafCounter = leftIndex.length;
+        int nodeCounter = 1;
+        for (int i = 0; i < size && check; i++) {
+            if (leftIndex[i] != NULL) {
```
jotok, line 123: What if `leftIndex[i] == null` but `rightIndex[i] != null` or if the left index is a leaf but the right one isn't? Should we check those cases?

Java/core/src/main/java/com/amazon/randomcutforest/state/store/NodeStoreMapper.java
```diff
@@ -33,34 +37,132 @@
...
+        state.setLeafFreeIndexPointer(model.getLeafFreeIndexPointer());
+        if (!samplerNeeded) {
+            state.setLeafPointIndex(ArrayPacking.pack(model.getLeafPointIndex(), compress));
+            state.setLeafmass(ArrayPacking.pack(model.getLeafMass(), compress));
+        }
+
         return state;
     }
+
+    boolean reduceToBits(int size, int[] leftIndex, int[] rightIndex) {
```
jotok, line 118: It looks like this method is doing two things: encoding the left index and right index arrays to indicate whether a given node has children or not, and performing some correctness on the arrays. Currently I think certain kinds of bad inputs could still pass the checks in this method. I think we should either make this method fully validate the arrays or else we should remove the existing validation. The partial validation we're currently doing seems like it lead to trouble later.

My suggestion would be to add a `isCanonical()` or `isCanonicalized()` method to NodeStore and remove the validation from here. It will make creating the bit arrays simpler (although we'll end up doing two scans of the index arrays).

Java/core/src/main/java/com/amazon/randomcutforest/state/store/NodeStoreState.java
```diff
@@ -19,24 +19,24 @@
...
-    public short[] smallFreeIndexes;
-    /*
-     * the following are for NodeStore
-     */
-    public int[] parentIndex;
-    public int[] leftIndex;
-    public int[] rightIndex;
-    public int[] mass;
-    public int[] freeIndexes;
+    private int capacity;
```
jotok, line 22: Nice, I think we meant to do this a while ago.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/NodeStoreState.java
```diff
@@ -19,24 +19,24 @@
...
+    private boolean compressed;
+    private int[] cutDimension;
+    public double[] cutValueDouble;
```
jotok, line 25: missed one?

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeDoubleMapper.java
```diff
@@ -19,70 +19,57 @@
...
-import com.amazon.randomcutforest.store.SmallNodeStore;
 import com.amazon.randomcutforest.tree.CompactRandomCutTreeDouble;
 
 @Getter
 @Setter
 public class CompactRandomCutTreeDoubleMapper implements
         IContextualStateMapper<CompactRandomCutTreeDouble, CompactRandomCutTreeState, CompactRandomCutTreeContext> {
 
-    private boolean boundingBoxCacheEnabled;
+    private boolean samplersNeeded = false;
```
jotok, line 34: Can we change this flag name to indicate what's being stored or omitted?

Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
```diff
@@ -182,14 +209,12 @@
...
-     * @return map of leaf,corresponding pairs in use for the node store
-     */
-    default Map<Integer, Integer> getLeavesAndParents() {
-        return null;
-    }
+    boolean isLeaf(int index);
+
+    int getCapacity();
+
+    int computeLeafIndex(int index);
```
jotok, line 216: Does this really need to be in the interface? It seems like this should be handled internally in the node store.

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
```diff
@@ -34,40 +39,78 @@
...
+        } else {
+            this.leafPointIndex = new int[capacity + 1];
+            Arrays.fill(this.leafPointIndex, PointStore.INFEASIBLE_POINTSTORE_INDEX);
+        }
+        if (leafMass != null) {
+            for (int i = 0; i < capacity; i++) {
+                if (parentIndex[i] == NULL) {
+                    rebuildMass(i);
+                }
+            }
```
jotok, line 104: Can we move this inside the previous if-block?

Also, since `rebuildMass` is recursive, would it suffice to call `rebuildMass` once on the root node?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -150,9 +103,80 @@ protected void addSequenceIndex(Integer nodeRef, long sequenceIndex) {
...
+
+    public void renormalize() {
+        NodeStore result = new NodeStore(maxSize - 1);
+        if (root != null) {
+            if (!isLeaf(root)) {
+                int nodeCounter = 0;
+                int currentNode = 0;
+                int leafcounter = maxSize - 1;
+                int[] map = new int[2 * maxSize - 1];
+                ArrayBlockingQueue<Integer> nodeQueue = new ArrayBlockingQueue<>(maxSize - 1);
```
jotok, line 120: Do we really need a blocking queue? If not, we could just use a LInkedList here.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -150,9 +103,80 @@ protected void addSequenceIndex(Integer nodeRef, long sequenceIndex) {
         sequenceIndexes[leafRef].put(sequenceIndex, num + 1);
     }
 
+    public double getBoundingBoxCacheFraction() {
+        return boundingBoxCacheFraction;
+    }
+
+    abstract void swapCaches(int[] map);
+
+    public void renormalize() {
```
jotok, line 112: Let's stick to either "canonical" or "normalized" throughout.

Should we add a method to NodeStore to "canonicalize" in place? It could return the index map:

```
public int[] canonicalize(PointRef root);
```

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -150,9 +103,80 @@ protected void addSequenceIndex(Integer nodeRef, long sequenceIndex) {
...
+                root = 0;
+            } else {
+                root = result.addLeaf(NULL, getPointReference(root), getMass(root));
+            }
+        }
+        nodeStore = result;
+    }
+
     @Override
     protected void deleteSequenceIndex(Integer nodeRef, long sequenceIndex) {
```
jotok, line 178: Should we move sequence indexes into the node store while we're at it? Then `computeLeafIndex` wouldn't have to be a public method.

Java/core/src/test/java/com/amazon/randomcutforest/state/store/NodeStoreMapperTest.java
```diff
@@ -22,35 +22,33 @@
...
     @BeforeEach
     public void setUp() {
-        mapper = new SmallNodeStoreMapper();
+        mapper = new NodeStoreMapper();
     }
 
     @Test
     public void testRoundTrip() {
-        SmallNodeStore store = new SmallNodeStore((short) 10);
+        NodeStore store = new NodeStore((short) 10);
```
jotok, line 37: I think we don't have to cast to `short` any more.

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
```diff
@@ -80,26 +73,22 @@ public CompactSampler toModel(CompactSamplerState state, long seed) {
...
-            }
-        } else {
-            state.setWeight(model.getWeightArray());
-            state.setPointIndex(model.getPointIndexArray());
-            if (model.isStoreSequenceIndexesEnabled()) {
-                state.setSequenceIndex(model.getSequenceIndexArray());
-            }
+        state.setWeight(Arrays.copyOf(model.getWeightArray(), model.size()));
+        state.setPointIndex(
+                ArrayPacking.pack(Arrays.copyOf(model.getPointIndexArray(), model.size()), state.isCompressed()));
```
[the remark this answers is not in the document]
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/NodeStoreMapper.java
sudiptoguha: In that first case the tree is malformed in the first place. But the left can be a leaf and the right can be non-leaf.
But the point about validation in nodestore makes sense.

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
sudiptoguha: The node store does not know the root :) But that is an issue we perhaps should fix ...

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: We need the strong FIFO property -- but we can use a linked list.

Java/core/src/test/java/com/amazon/randomcutforest/state/store/NodeStoreMapperTest.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Making the root a part of the nodestore will help this ...

Java/core/src/main/java/com/amazon/randomcutforest/state/store/NodeStoreMapper.java
sudiptoguha: Done. Maybe there can be a better reorganization of code down the road -- but perhaps not in this PR.

Java/core/src/main/java/com/amazon/randomcutforest/store/NodeStore.java
sudiptoguha: Kept this the same for now, maybe this would be resolved in a reorganization of code.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Solved partially; the canonicalAndNotALeaf() checks if the root = 0 in the nodestore

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: True, but we have to keep PointSum and the boxes in the tree (precision issues), and thus any reconfiguration has to be exposed to the tree anyways -- the NodeStore now just has information about internal nodes and leaves. It may be easier to get a different implementation of NodeStore if needed (if we keep the sequenceIndex separate, at least for now). This decision would not affect the state in any way.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/NodeStoreMapper.java
```diff
@@ -33,34 +37,132 @@
      */
     private boolean copy = true;
 
+    private boolean feasibleCanonical = true;
```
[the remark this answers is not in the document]
sudiptoguha: Fixed

Java/core/src/main/java/com/amazon/randomcutforest/executor/PointStoreCoordinator.java
```diff
@@ -38,16 +39,19 @@ public PointStoreCoordinator(IPointStore<Point> store) {
...
     }
 
     @Override
     public void completeUpdate(List<UpdateResult<Integer>> updateResults, Integer updateInput) {
-        updateResults.forEach(result -> {
-            result.getAddedPoint().ifPresent(store::incrementRefCount);
-            result.getDeletedPoint().ifPresent(store::decrementRefCount);
-        });
-        store.decrementRefCount(updateInput);
+        if (updateInput != null) { // can be null for initial shingling
```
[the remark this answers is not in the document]
sudiptoguha: True. Next PR? Where code will only be deleted!

Java/core/src/main/java/com/amazon/randomcutforest/state/sampler/CompactSamplerMapper.java
```diff
@@ -37,35 +38,27 @@
...
 
     /**
-     * If true, then model data will be copied (i.e., the state class will not share
-     * any data with the model). If false, some model data may be shared with the
-     * state class. Copying is enabled by default.
+     * used to compress data, can be set to false for debug
      */
+    private boolean compress = true;
+
+    // the following is for tests
```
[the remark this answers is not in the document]
sudiptoguha: The tests use the copy version -- did not want to remove the copy here and trigger 4 more files changed. Will remove in next PR which will just delete lines.

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeFloatMapper.java
```diff
@@ -19,67 +19,55 @@
...
-import com.amazon.randomcutforest.store.SmallNodeStore;
 import com.amazon.randomcutforest.tree.CompactRandomCutTreeFloat;
 
 @Getter
 @Setter
 public class CompactRandomCutTreeFloatMapper implements
         IContextualStateMapper<CompactRandomCutTreeFloat, CompactRandomCutTreeState, CompactRandomCutTreeContext> {
 
-    private boolean boundingBoxCacheEnabled;
+    private boolean partialTreeInUse = false;
```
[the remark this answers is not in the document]
sudiptoguha: A new version of samplersNeeded. The mappers need to know that at serialization time. Deserialization is based on state.

Java/core/src/main/java/com/amazon/randomcutforest/store/INodeStore.java
```diff
@@ -182,14 +209,16 @@
...
-        return null;
+    boolean isLeaf(int index);
+
+    int getCapacity();
+
+    int computeLeafIndex(int index);
+
+    int size();
+
+    default boolean isCanonicalAndNotALeaf() {
```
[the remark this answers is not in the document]
sudiptoguha: Fair. Another thing to remove in a remove only PR.

PR #172, opened 2021-04-22
Rust CI Workflow

cswiercz:
Create a workflow specification for Rust pull requests. Builds package and runs tests.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,25 @@
+name: Rust CI
+
+on:
+  pull_request:
+    branches: [ main ]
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build Rust
+      run: cargo build --verbose
```
jotok, line 23: The default `build` target doesn't run tests?

PR #173, opened 2021-04-22
[rust] Random Cut Trees

cswiercz:
Implement random cut trees using the basic building blocks of #169 . The additions in this PR consist of three files:

* `tree.rs` -- the implementation of a `Tree<T>` where `T` is a float type (either `f32` or `f64`)
* `point_adder.rs` -- for ease of reading and to avoid making `Tree` to "fat" of a type, I factored out the point addition code into a separate type, `PointAdder`. The only data `PointAdder` has is a mutable reference to a tree and the input point.
* `point_deleter.rs` -- (same as `PointAdder` but for point deletion)

Another design choice is to make the point store an `Rc<RefCell<PointStore>`. Because both the `Tree` and the upcoming `SampledTree` need to mutate the point store I didn't want to give `Tree` sole ownership.

Note that all contributions thus far contain tests in the documentation, which are run when invoking `cargo test`. This PR also include separate (non-doc) tests for validating the point addition and deletion algorithms.

Building:

    cargo build

Testing:

    cargo test

Documentation:

    cargo docs --no-deps --open


Rust/src/tree/point_adder.rs
```diff
@@ -0,0 +1,308 @@
...
+
+
+/// A mechanism for adding points to trees.
+/// 
+/// Initialized with a mutable reference to a tree, this struct contains all of
+/// the instructions necessary to add points to trees. This code is factored
+/// out from `Tree` for improved readability.
+pub struct PointAdder<'a, T> {
+    tree: &'a mut Tree<T>,
+}
```
jotok, line 18: Is it possible to have multiple `impl` blocks for the same struct in different files? If so, you could keep your code organization without having to introduce wrapper structs.

Rust/src/tree/point_adder.rs
```diff
@@ -0,0 +1,308 @@
...
+                };
+                let node = Node::new_leaf(point_key);
+                let node_key = self.tree.node_store_mut().insert(node);
+                self.tree.set_root_node(Some(node_key));
+                AddResult::AddedPoint(point_key)
+            }
+        }
+    }
+
+    /// Main recursive point addition algorithm givena new point and a current
```
jotok, line 45: minor: typo in "givena".

jotok (approved):

Rust/src/tree/point_adder.rs
cswiercz: Great idea. I initially misread this as "move everything into the same `impl` block" but got it right the second time. The only downside is that private functions cannot be shared between different `impl` blocks. That's probably good by design, but it means that there are a handful of helper functions that I needed to make public as I cannot create two private copies,  one for each block.

cswiercz:
@jotok  Now that #172 is implemented, I see that that I ended up using a feature that is only available on Rust nightly builds. Because of this, it’s causing a build failure in the CI.

Two options:
1. Figure out a way to use nightly in the CI and communicate to users that this is a requirement.
2. Rewrite code to manually use the trait bounds.

More information: I use the unstable feature, [`trait_alias`](https://doc.rust-lang.org/unstable-book/language-features/trait-alias.html) so that I don't have to repeat the same trait bounds on a large number of types:

```rust
/// Floating point type for RCF representing either f32 or f64.
///
/// This type represents either a f32 or f64. In order to make the algorithm
/// generic across floats we need to specify some of the necessary traits for
/// various operations. Since this is used everywhere in the library we
/// encapsulate all of these traits into a single trait alias.
pub trait RCFFloat = Float + One + PartialOrd + Sum + Zero;
```

Taking a look at this features corresponding [tracking issue](https://github.com/rust-lang/rust/issues/41517), it seems like most of the TODOs are done but there are still some unresolved questions. However, the last comment on this issue was on Jan 1, 2021.

Thoughts? My opinion is that it may be worth the added mess of trait bounds if the user doesn't have to use the nightly build for a single convenience feature.

jotok:
I would check the following:
- is there a planned release date for the feature? (sounds like no)
- is it possible to specify in the build file that we want to use nightly?

If the answer to both questions is now, then I think it's better for usability if we don't rely on this feature for now. We can create an issue to re-do this work if/when the feature is launched.

PR #174, opened 2021-04-26
Store point data as bytes in PointStoreState

jotok:
Update `PointStoreState` to use a `byte[]` array field for point data, instead of separate `float[]` and `double[]` fields to handle the different cases.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/util/ArrayPacking.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/util/ArrayPacking.java
@@ -170,7 +170,7 @@ public class ArrayPacking {
             int count = 0;
             for (int i = 3; i < packedArray.length; i++) {
                 long code = packedArray[i];
-                for (int j = 0; j < packNum && count < min(output.length, length); j++) {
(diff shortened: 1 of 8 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/util/ArrayPacking.java
```diff
@@ -78,37 +113,63 @@ static int logMax(long base) {
...
                 }
             }
         } else {
             long base = ((long) max - min + 1);
             int packNum = logMax(base);
             int count = 0;
-            for (int i = 3; i < inputArray.length; i++) {
-                long code = inputArray[i];
+            for (int i = 3; i < packedArray.length; i++) {
+                long code = packedArray[i];
```
sudiptoguha, line 172: The next line should be min(packedArray[2],output.length).  But we can fix in the next PR as well.

sudiptoguha (approved):

Java/core/src/main/java/com/amazon/randomcutforest/util/ArrayPacking.java
jotok: Fixed.

PR #175, opened 2021-04-27
single precision cuts

sudiptoguha:
*Issue #, if available:* use single precision random cuts when single precision is set.

*Description of changes:* Reduces the precision of stored cuts, saving space. Also refactors the tests so that shingling based tests are standalone. Adds version string to state classes.


Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestState.java
```diff
@@ -30,6 +30,8 @@
 @Data
 public class RandomCutForestState {
 
+    private String version = "compact";
```
jotok, line 33: I think we should use number versions, and we should store them either as String constants or as an enum so that we use a consistent scheme in all the different files. E.g.
```
package com.amazon.randomcutforest.state;

public class Version {
  public static final String V2_0 = "2.0";
}
```

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -146,6 +146,18 @@ static Cut randomCut(Random random, AbstractBoundingBox<?> box) {
...
 
+    /**
+     * A local version of randomcut that can be overridden for lower precision
+     * requirments
+     * 
+     * @param random random number generator
+     * @param box    bounding box corresponding to the generation of the cut
+     * @return a cut corresponding to the box
+     */
+    Cut treeCut(Random random, AbstractBoundingBox<?> box) {
```
jotok, line 157: We already have a method named `randomCut` that can be overridden. Why do we need a new method with a new name here? I think the name `treeCut` is misleading if this does the same thing as `randomCut`. It sounds like this would prune branches off the tree.

This should be `protected` if we want it to be overridden by subclasses.

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeFloat.java
```diff
@@ -118,4 +119,43 @@ public CompactRandomCutTreeFloat build() {
...
+        for (int i = 0; i < box.getDimensions(); i++) {
+            double range = box.getRange(i);
+            if (breakPoint <= range) {
+                float cutValue = (float) (box.getMinValue(i) + breakPoint);
+
+                // Random cuts have to take a value in the half-open interval [minValue,
+                // maxValue) to ensure that a
+                // Node has a valid left child and right child.
+                if ((cutValue == box.getMaxValue(i)) && (box.getMinValue(i) < box.getMaxValue(i))) {
+                    cutValue = Math.nextAfter(cutValue, box.getMinValue(i));
```
jotok, line 148: If min-values and max-values are both double-precision, then isn't it possible that taking a single-precision step will result in a cut value that's smaller than the min value?

Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestState.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: RandomCut a static function. We can make that protected ...

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactRandomCutTreeFloat.java
sudiptoguha: Yes it can. But minValues and maxValues are both float (as would be guaranteed for single precision) then there should be no problem. We can force an explicit conversion to make that point. However the bigger issue is that if max - min is a double and 0 as a float then the range calculations etc will all have problems.

PR #183, opened 2021-05-05
refactoring caching logic

sudiptoguha:
*Description of changes:* Refactoring caching logic to allow for null during deserialization as well as tune for continuous runs

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -60,7 +60,8 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
 
     protected INodeStore nodeStore;
     protected IPointStoreView<Point> pointStore;
-    protected AbstractBoundingBox<Point>[] cachedBoxes;
+    protected IBoxCache<Point> cacheManager;
+    // protected AbstractBoundingBox<Point>[] cachedBoxes;
     protected Point[] pointSum;
     protected HashMap<Long, Integer>[] sequenceIndexes;
 
@@ -115,8 +116,6 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
     public double getBoundingBoxCacheFraction() {
         return boundingBoxCacheFraction;
     }
-
-    abstract void swapCaches(int[] map);
 
     /**
      * The following function reorders the nodes stored in the tree in a breadth
@@ -200,9 +199,8 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
                 }
                 validateInternalState(currentNode == nodeStore.size(), "incorrect state");
 
-                if (cachedBoxes != null) {
-                    swapCaches(map);
-                }
+                cacheManager.swapCaches(map);
+
                 if (enableSequenceIndices) {
                     HashMap<Long, Integer>[] newSequence = new HashMap[maxSize];
                     for (int i = 0; i < maxSize; i++) { // iterate over leaves
@@ -259,16 +257,17 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
     protected AbstractBoundingBox<Point> constructBoxInPlace(Integer nodeReference) {
         if (isLeaf(nodeReference)) {
             return getMutableLeafBoxFromLeafNode(nodeReference);
-        } else if (cachedBoxes != null && cachedBoxes[nodeReference] != null) {
-            return cachedBoxes[nodeReference].copy();
+        } else if (boundingBoxCacheFraction == 0) {
+            return constructBoxInPlace(constructBoxInPlace(getLeftChild(nodeReference)), getRightChild(nodeReference));
         } else {
+            AbstractBoundingBox<Point> oldBox = (cacheManager.inUse(nodeReference)) ? cacheManager.getBox(nodeReference)
+                    : null;
+            if (oldBox != null) { // note cachemanager can have a null box, after deserialization
+                return oldBox.copy();
+            }
             AbstractBoundingBox<Point> currentBox = constructBoxInPlace(
                     constructBoxInPlace(getLeftChild(nodeReference)), getRightChild(nodeReference));
-            // the following is useful in cases where the forest is deserialized and
-            // even though caches may be set, the individual boxes are null
-            if (cachedBoxes != null && cacheRandom.nextDouble() <= boundingBoxCacheFraction) {
-                cachedBoxes[nodeReference] = currentBox.copy();
-            }
+            cacheManager.setBox(nodeReference, currentBox.copy());
             return currentBox;
         }
     }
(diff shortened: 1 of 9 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -60,7 +60,8 @@
 
     protected INodeStore nodeStore;
     protected IPointStoreView<Point> pointStore;
-    protected AbstractBoundingBox<Point>[] cachedBoxes;
+    protected IBoxCache<Point> cacheManager;
+    // protected AbstractBoundingBox<Point>[] cachedBoxes;
```
jotok, line 64: Delete commented out code.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -60,7 +60,8 @@
 
     protected INodeStore nodeStore;
     protected IPointStoreView<Point> pointStore;
-    protected AbstractBoundingBox<Point>[] cachedBoxes;
+    protected IBoxCache<Point> cacheManager;
```
jotok, line 63: Could we just call this `boxCache` to match the type?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -259,16 +257,17 @@ protected void deleteSequenceIndex(Integer nodeRef, long sequenceIndex) {
     protected AbstractBoundingBox<Point> constructBoxInPlace(Integer nodeReference) {
         if (isLeaf(nodeReference)) {
             return getMutableLeafBoxFromLeafNode(nodeReference);
-        } else if (cachedBoxes != null && cachedBoxes[nodeReference] != null) {
-            return cachedBoxes[nodeReference].copy();
+        } else if (boundingBoxCacheFraction == 0) {
```
jotok, line 260: Let's define a method
```
public boolean isBoundingBoxCacheEnabled() {
  return boudningBoxCacheFraction > 0.0;
}
```

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -259,16 +257,17 @@ protected void deleteSequenceIndex(Integer nodeRef, long sequenceIndex) {
...
+            return constructBoxInPlace(constructBoxInPlace(getLeftChild(nodeReference)), getRightChild(nodeReference));
         } else {
+            AbstractBoundingBox<Point> oldBox = (cacheManager.inUse(nodeReference)) ? cacheManager.getBox(nodeReference)
```
jotok, line 263: I'm assuming this is written this way because `getBox` will throw an exception if a particular nodeReference is not in use. I think our other choices are to have `getBox` return null (which would be consistent with the Map interface), or to have `getBox` return an optional.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCache.java
```diff
@@ -0,0 +1,97 @@
...
+    int maxSize;
+
+    protected BoxCache(long seed, double cacheFraction, int maxSize) {
+        randomSeed = seed;
+        this.cacheFraction = cacheFraction;
+        this.maxSize = maxSize;
+    }
+
+    @Override
+    public boolean inUse(int index) {
```
jotok, line 37: I would either call this `isOccupied` to match our use of `occupied` in the store classes, or `containsKey` to match the Map interface.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCache.java
```diff
@@ -0,0 +1,97 @@
...
+        if (cacheFraction == 0.0) {
+            return false;
+        } else if (cacheFraction == 1.0) {
+            return true;
+        } else if (cacheFraction >= 0.5) {
```
jotok, line 42: I would extract this check into a private or protected method, `isDirectMap` or something like that.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCacheDouble.java
```diff
@@ -0,0 +1,51 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest.tree;
+
+import java.util.HashMap;
+import java.util.Random;
+
+public class BoxCacheDouble extends BoxCache<double[]> {
```
jotok, line 21: I wonder if we should make the generic `BoxCache<T>` an abstract class, because it looks like it looks like it's not usable without subclassing. We could make `initiateMaps` a protected abstract method.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: fixed, returns null. Which was already there in the instance class

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCache.java
sudiptoguha: containsKey ... this is different from store class where everything is likely to be used/occupied

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCacheDouble.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public BoxCacheDouble(long seed, double cacheFraction, int maxSize) {
+        super(seed, cacheFraction, maxSize);
+        initiateMaps();
+    }
+
+    void initiateMaps() {
```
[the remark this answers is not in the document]
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCache.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.HashMap;
+import java.util.Random;
+
+public class BoxCache<Point> implements IBoxCache<Point> {
+
+    double cacheFraction;
+    Random cacheRandom;
+    long randomSeed;
+    AbstractBoundingBox<Point>[] cachedBoxes;
+    HashMap<Integer, Integer> cacheMap;
```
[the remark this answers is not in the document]
sudiptoguha: in the concrete classes; fixed with the abstract class

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCacheDouble.java
sudiptoguha: Done, abstract class.

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCache.java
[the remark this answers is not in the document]
sudiptoguha: Subclasses need some access at this moment for initialization and remapping. Maybe make an issue and come back when the methods have settled and we have viz into longer term needs?

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoxCache.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest.tree;
+
+public interface IBoxCache<Point> {
```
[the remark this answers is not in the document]
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/BoxCache.java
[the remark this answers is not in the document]
jotok: For subclass visibility you want `protected`. No modifier means "package-private" visibility, meaning the field is visible only to other classes defined in the same package.

PR #184, opened 2021-05-05
Update the `isOutputReady()` method to check component models

### LINKED ISSUE #181: [Java] The `isOutputReady` method should query component models
In the current implementation, the `isOutputReady` method in `RandomCutForest` only checks to see how many points have been submitted to the forest. The changes in #161 made it so that samplers could fill up at different rates. This issue is to implement the following behavior:

- Add the method `isOutputReady` to the `IComponentModel` interface. The expected behavior is that after the first time this method returns `true`, it will remain `true` unless some action is taken by the user (e.g., if we were to add support for flushing a sample).
- The forest-level `isOutputReady` method should be equivalent to asking if every component model is ready.

jotok:
Add a new isOutputReady method to the ITraversable interface. The forest
will now check all component models for readiness, instead of waiting
for a fixed number of updates.

Closes #181

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
@@ -80,11 +80,18 @@ public abstract class AbstractRandomCutTree<Point, NodeReference, PointReference
         this.enableSequenceIndices = enableSequenceIndices;
     }
 
-    public AbstractRandomCutTree(AbstractRandomCutTree.Builder builder) {
-        this.random = new Random(builder.randomSeed);
+    public AbstractRandomCutTree(AbstractRandomCutTree.Builder<?> builder) {
+        if (builder.random != null) {
+            this.random = builder.random;
+        } else {
+            this.random = new Random(builder.randomSeed);
+        }
         this.enableCenterOfMass = builder.centerOfMassEnabled;
         this.enableSequenceIndices = builder.storeSequenceIndexesEnabled;
         this.boundingBoxCacheFraction = builder.boundingBoxCacheFraction;
+
+        // This should be set to an appropriate value in a subclass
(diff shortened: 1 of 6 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/RandomCutTree.java
```diff
@@ -343,41 +330,9 @@ protected void deleteSequenceIndex(Node node, long uniqueSequenceNumber) {
...
-            this.random = new Random(randomSeed);
-            return (T) this;
-        }
-
-        public T random(Random random) {
-            this.random = random;
-            return (T) this;
-        }
-
+    public static class Builder<T extends Builder<T>> extends AbstractRandomCutTree.Builder<T> {
```
sudiptoguha, line 333: Nice!

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -80,11 +80,18 @@ public AbstractRandomCutTree(Random random, boolean enableCache, boolean enableC
...
+            this.random = builder.random;
+        } else {
+            this.random = new Random(builder.randomSeed);
+        }
         this.enableCenterOfMass = builder.centerOfMassEnabled;
         this.enableSequenceIndices = builder.storeSequenceIndexesEnabled;
         this.boundingBoxCacheFraction = builder.boundingBoxCacheFraction;
+
+        // This should be set to an appropriate value in a subclass
+        outputAfter = Integer.MAX_VALUE;
```
sudiptoguha, line 94: Makes sense. Do we set this in AbstractCompactRandomCutTree? Samplesize/4 ??

sudiptoguha: Nevermind. Already present.

sudiptoguha (approved):

PR #185, opened 2021-05-06
fixes and cleanup

sudiptoguha:
*Description of changes:* fixes for CompactNodeView

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactNodeView.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactNodeView.java
@@ -15,31 +15,48 @@
 
 package com.amazon.randomcutforest.tree;
 
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+
 import java.util.Arrays;
 import java.util.Set;
 
-public class CompactNodeView implements INode<Integer> {
-    final AbstractCompactRandomCutTree<?> tree;
-    int currentNodeOffset;
+import com.amazon.randomcutforest.store.INodeStore;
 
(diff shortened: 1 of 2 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactNodeView.java
```diff
@@ -15,31 +15,48 @@
...
     int currentNodeOffset;
+    IBoxCache<Point> boxCache;
+    INodeStore nodeStore;
 
-    public CompactNodeView(AbstractCompactRandomCutTree<?> tree, int initialNodeIndex) {
+    public CompactNodeView(AbstractCompactRandomCutTree<Point> tree, int initialNodeIndex) {
         this.tree = tree;
         this.currentNodeOffset = initialNodeIndex;
+        boxCache = tree.boxCache;
+        nodeStore = tree.nodeStore;
```
jotok, line 35: Did this make a different in throughput?

Java/core/src/main/java/com/amazon/randomcutforest/tree/CompactNodeView.java
```diff
@@ -15,31 +15,48 @@
...
+    IBoundingBoxView getBox(int node) {
+        if (!nodeStore.isLeaf(node)) {
+            IBoundingBoxView box = boxCache.getBox(node);
+            if (box != null)
+                return box;
+        }
+        return tree.getBoundingBox(node);
     }
 
     public IBoundingBoxView getSiblingBoundingBox(double[] point) {
```
jotok, line 55: We don't have to change it here, but this name always trips me up because I expect it to return the sibling of the current node. I wonder if there's another name that would be clearer.

jotok (approved):

PR #188, opened 2021-05-07
[rust] Implement the random cut forest type

cswiercz:
This commit only implements `update()` on random cut forests. We also implement a random cut forest builder type.

Only one new file: `random_cut_forest.rs`. (Plus adding the types to `lib.rs`.) Aside from initializers and getters, the only method implemented in `RandomCutForest` for now is `update()`. Basic anomaly scoring will be implemented in the subsequent PR.

To review/examine documentation:

    cargo docs --no-deps --open

This is part four of a stack of PRs for implementing an initial Rust version of RCF:

1. #169 
2. #173 
3. #178

### DIFF
--- a/Rust/src/random_cut_forest.rs
+++ b/Rust/src/random_cut_forest.rs
@@ -42,7 +42,6 @@ use std::iter::Sum;
 pub struct RandomCutForest<T> {
     dimension: usize,
     num_observations: usize,
-    num_trees: usize,
     sample_size: usize,
     time_decay: f32,
     trees: Vec<SampledTree<T>>,
@@ -101,7 +100,7 @@ impl<T> RandomCutForest<T>
     pub fn num_observations(&self) -> usize { self.num_observations }
 
     /// Return the number of trees in this forest.
-    pub fn num_trees(&self) -> usize { self.num_trees }
+    pub fn num_trees(&self) -> usize { self.trees.len() }
 
     /// Return the number of samples/observations stored in each tree.
     pub fn sample_size(&self) -> usize { self.sample_size }
@@ -113,7 +112,7 @@ impl<T> RandomCutForest<T>
 
 /// Convenient mechanism for creating [`RandomCutForest`]s.
 ///
-/// Random cut forests are highly configurable and come with large number of
+/// Random cut forests are highly configurable and come with a large number of
 /// parameters, many of which have reasonable default values. This builder
 /// makes it easier to construct a random cut forest model
 ///
@@ -212,7 +211,6 @@ impl<T> RandomCutForestBuilder<T>
 
         RandomCutForest {
             dimension: self.dimension,
-            num_trees: self.num_trees,
             sample_size: self.sample_size,
             time_decay: self.time_decay,
             trees: trees,
--- a/Rust/src/random_cut_forest.rs
+++ b/Rust/src/random_cut_forest.rs
@@ -1,0 +1,250 @@
+extern crate num_traits;
+use num_traits::Float;
+
+use crate::SampledTree;
+
+use std::marker::PhantomData;
+use std::iter::Sum;
+
+/// A random cut forest model.
+///
+/// Random cut forests are model-free data structures for sketching data
+/// streams. This type is the main interface for training a random cut forest
+/// and using the model for various scoring operations such as anomaly detection.
+///
+/// A random cut forest is a collection of random cut trees; specifically a
+/// collection of [`SampledTree`] structs. An update to a random cut forest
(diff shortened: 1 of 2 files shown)


Rust/src/random_cut_forest.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// //forest.update(vec![2.0, 3.0, 4.0]);
+    /// ```
+    pub fn update(&mut self, point: Vec<T>) {
+        assert_eq!(point.len(), self.dimension,
+            "Dimension mismatch. Expected {}-dimensional input.",
+            self.dimension);
+
+        self.num_observations += 1;
+        for tree in self.trees.iter_mut() {
+            tree.update(point.clone(), self.num_observations)
```
jmazanec15, line 90: If each tree has the same number of observations, does the tree need to store this information or can it just be stored in the rcf? I believe the sampler will store num_observations as well.

Rust/src/random_cut_forest.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub fn sample_size(&self) -> usize { self.sample_size }
+
+    /// Return a vector of references to the trees of the forest.
+    pub fn trees(&self) -> &Vec<SampledTree<T>> { &self.trees }
+}
+
+
+/// Convenient mechanism for creating [`RandomCutForest`]s.
+///
+/// Random cut forests are highly configurable and come with large number of
```
jmazanec15, line 116: nit: a large

Rust/src/random_cut_forest.rs
```diff
@@ -0,0 +1,250 @@
...
+/// parameters, many of which have reasonable default values. This builder
+/// makes it easier to construct a random cut forest model
+///
+/// The builder has the following required parameters for initialization:
+///
+/// * `dimension`
+///
+/// The builder uses the following defaults for the remaining parameters:
+///
+/// * `num_trees = 50`
```
jmazanec15, line 126: Out of curiosity, is there any way in Rust where these wouldnt need to be hardcoded?

Rust/src/random_cut_forest.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    /// Set the number of samples retained by each tree in the random cut forest.
+    pub fn sample_size(mut self, sample_size: usize) -> RandomCutForestBuilder<T> {
+        self.sample_size = sample_size;
+        self
+    }
+
+    /// Set the random sampling decay factor of the random cut forest.
+    pub fn time_decay(mut self, time_decay: f32) -> RandomCutForestBuilder<T> {
+        self.time_decay = time_decay;
```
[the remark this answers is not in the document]
cswiercz: It's a non-negative real number. I can throw in checks for each of the model parameters in the builder.

On that note, I don't know what is the Rust convention for testing validity of configuration. Type constructors aren't "standard", in the sense that I could always create a new RCF type like so:

```rust
let mut forest = RandomCutForest { sample_size: 256, num_trees: 50, ... };
```

By doing so I can skip any checks performed in some `::new()` function. Thoughts?

Rust/src/random_cut_forest.rs
jotok: I believe these trees are "sampled trees", meaning this is the operation submitting the point to the sampler. We want the sampler to receive (point, index) pairs for data debugging, and we want to take the index as a parameter (instead of using a counter in the sampler) for flexibility. In the future we may want to support out-of-order updates (e.g., when processing data in batch mode).

jotok (approved):

Rust/src/random_cut_forest.rs
jmazanec15: Makes sense, thanks

Rust/src/random_cut_forest.rs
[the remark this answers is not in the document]
jmazanec15: Could RandomCutForest struct be made private to the module? That way type construction could not be used outside of the module and validation could be handled in new function.

Related stack overflow: https://stackoverflow.com/questions/45017183/can-you-disable-constructor-syntax-for-a-type

cswiercz (commented):
(I don't know why my comments were turned into a "review", whatever that means. I can't "approve" of these comments or whatever because I'm the author of this PR. I guess I need to read about this on the GitHub docs...)

cswiercz:
Responding to a comment by @jmazanec15 , I don't think I can make `RandomCutForest` private because it needs to be exposed to the user. That is, I don't think you can return private types to the customer.

That said, because the fields are not public I don't think you can create a `RandomCutForest` with the usual `{ ... }` syntax outside of the module.

PR #192, opened 2021-05-13
[rust] Basic visitor-based anomaly score

cswiercz:
Provides a basic implementation of visitor-based anomaly scoring. The algorithm is implemented in, `RandomCutForest::update()` and `AnomalyScoreVisitor`.

This PR also includes the first major regression test where we confirm that a data point several standard deviations away from a Gaussian blob has a larger score than any other training score. We also test on the double Gaussian/"barbell" example from the original paper.

Review documentation:

    cargo doc --no-deps --open

This is part five of a stack of PRs for implementing an initial Rust version of RCF:

1. #169 
2. #173 
3. #178 
4. #188

### TODO

- [x] Compare anomaly scores between Java and Rust versions.

### DIFF
--- a/Rust/src/anomaly_score_visitor.rs
+++ b/Rust/src/anomaly_score_visitor.rs
@@ -7,7 +7,7 @@ use crate::tree::{BoundingBox, Internal, Leaf, Tree};
 
 
 /// A visitor on nodes used to compute an anomaly score.
-/// 
+///
 /// Given a tree and a point to score, the tree returns a branch of nodes using
 /// [`Tree::traverse()`]. This branch is a path from the leaf node in the tree
 /// closest to the point to score (in the L1 norm) to the root node of the tree.
@@ -15,7 +15,7 @@ use crate::tree::{BoundingBox, Internal, Leaf, Tree};
 /// then updates the score using the *separation probability* at each internal
 /// node; the probability that a random cut separates the point to score from
 /// that node's bounding box.
-/// 
+///
 /// ```text
 ///    ______________   ___
 ///   |             P|   |
@@ -24,11 +24,11 @@ use crate::tree::{BoundingBox, Internal, Leaf, Tree};
 ///   |         |    |   |
 ///   |    B    |    |   | l2
 ///   |_________|____|  _|_
-/// 
-///    |--------||---| 
+///
+///    |--------||---|
 ///        l1     d1
 /// ```
-/// 
+///
 /// The final anomaly score is retrieved using [`Self::get_result`].
 pub struct AnomalyScoreVisitor<'a, T> {
     // A tree on which an anomaly score will be computed
@@ -55,8 +55,8 @@ where
 {
 
     /// Initialize an anomaly score visitor with a tree and a point to score.
-    /// 
-    /// The anomaly score of this visitor is initialized to zero. 
+    ///
+    /// The anomaly score of this visitor is initialized to zero.
     pub fn new(
         tree: &'a Tree<T>,
         point_to_score: &'a Vec<T>,
@@ -71,7 +71,7 @@ where
     }
 
     /// Initialize the anomaly score from a leaf node.
-    /// 
+    ///
     /// A leaf node is the first node visited in the anomaly scoring process.
     /// The initial anomaly score depends whether the point at the leaf node
     /// is equal to the point to score.
@@ -80,7 +80,7 @@ where
         let point = point_store.get(leaf.point()).unwrap();
         if *self.point_to_score == *point {
             self.point_inside_box = true;
-            self.anomaly_score = damp::<T>(leaf.mass(), self.tree.mass()) * 
+            self.anomaly_score = damp::<T>(leaf.mass(), self.tree.mass()) *
                 score_seen(depth, leaf.mass());
         } else {
             self.anomaly_score = score_unseen(depth);
@@ -88,8 +88,8 @@ where
     }
 
     /// Update the anomaly score from an internal node.
-    /// 
-    /// After visiting a leaf node in [`Self::accept_leaf`] the remaining nodes 
+    ///
+    /// After visiting a leaf node in [`Self::accept_leaf`] the remaining nodes
     /// encountered while traversing up the tree to the root are internal
     /// nodes. The main piece of information used by this algorithm is the
     /// bounding box at these nodes.
@@ -108,9 +108,9 @@ where
     }
 
     /// Normalize and return the anomaly score computed by the visitor.
-    /// 
-    /// The anomaly score computed by the visiting process, via 
-    /// [`Self::accept_leaf()`] and [`Self::accept()`], is normalized using the 
+    ///
+    /// The anomaly score computed by the visiting process, via
+    /// [`Self::accept_leaf()`] and [`Self::accept()`], is normalized using the
     /// mass of the tree before returning. This is so that the resulting anomaly
     /// score is independent of the number of samples in the tree.
     pub fn get_result(&self) -> T {
@@ -158,7 +158,7 @@ where
 
         range_diff_sum / new_range_sum
     }
-    
+
 }
 
 #[inline(always)]
--- a/Rust/src/anomaly_score_visitor.rs
+++ b/Rust/src/anomaly_score_visitor.rs
@@ -1,0 +1,196 @@
+extern crate num_traits;
+use num_traits::{Float, One, Zero};
+
+use std::iter::Sum;
+
+use crate::tree::{BoundingBox, Internal, Leaf, Tree};
+
+
+/// A visitor on nodes used to compute an anomaly score.
+/// 
+/// Given a tree and a point to score, the tree returns a branch of nodes using
+/// [`Tree::traverse()`]. This branch is a path from the leaf node in the tree
+/// closest to the point to score (in the L1 norm) to the root node of the tree.
+/// The anomaly score visitor initializes an anomaly score at the leaf node and
+/// then updates the score using the *separation probability* at each internal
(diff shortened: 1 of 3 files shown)


Rust/src/random_cut_forest.rs
```diff
@@ -90,6 +92,60 @@ impl<T> RandomCutForest<T>
...
+            .sum();
+        anomaly_score / T::from(self.num_trees()).unwrap()
+    }
+
+    /// Compute the anomaly score from a particular tree. Used by
+    /// ['Self::anomaly_score'].
+    fn tree_anomaly_score(&self, point: &Vec<T>, tree: &Tree<T>) -> T {
+        // traverse nodes from the root to the leaf closest to the input point
+        let nodes: Vec<&Node<T>> = tree.traverse(point).collect();
+        let mut visitor = AnomalyScoreVisitor::new(tree, point);
```
jmazanec15, line 135: Instead of having an AnomalyScoreVisitor, couldn't the forest have an AnomalyScorer that implements a `score` function? This way, the scoring functionality RCF uses could be configurable at construction.

Rust/src/anomaly_score_visitor.rs
```diff
@@ -0,0 +1,196 @@
...
+                    new_range_sum = new_range_sum + old_range;
+                    self.coordinate_inside_box[i] = true;
+                    continue;
+                }
+
+                let new_range = max_value - min_value;
+                new_range_sum = new_range_sum + new_range;
+                range_diff_sum = range_diff_sum + new_range - old_range;
+            } else {
+                new_range_sum = new_range_sum + old_range;
```
jmazanec15, line 151: Any reason not to use new_range_sum += old_range?

Rust/src/anomaly_score_visitor.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    /// Update the anomaly score from an internal node.
+    ///
+    /// After visiting a leaf node in [`Self::accept_leaf`] the remaining nodes
+    /// encountered while traversing up the tree to the root are internal
+    /// nodes. The main piece of information used by this algorithm is the
+    /// bounding box at these nodes.
+    pub fn accept(&mut self, node: &Internal<T>, depth: T) {
```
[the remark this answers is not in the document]
cswiercz: Standard nomenclature from GoF.

Rust/src/random_cut_forest.rs
cswiercz: Good thinking about future scoring functions. (As there are several known scoring functions and many many more unknown ones.) I'm not sure how best to go about this in Rust. Feel tree to cut a tracking ticket where we can start a discussion. Would be good to get people's ideas.

Initial idea: `Visitor` trait with [associated output type](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html). `RandomCutForest` would have a single generic scoring function that accepts a point and a type that implements `Visitor`. To implement `anomaly_score()` you just create an instance of the visitor and give it to this generic scoring function. I dunno. Again, best to start an issue thread.

Rust/src/anomaly_score_visitor.rs
cswiercz: The `Float` trait, which I use to genericize over `f32` and `f64` only implements `Add` and not `AddAssign`. I wanted to minimize the number of dependent traits for the input type and managed to reduce it to `Float + Sum`. The `Sum` trait means you can call `.sum()` on a collection of that type. This requirement mostly is due to making `BoundingBox` easier to implement.

In Rust nightly you can define `trait aliases`. See [this comment](https://github.com/aws/random-cut-forest-by-aws/pull/173#issuecomment-830291529) for context and info.

cswiercz: Anyway, the solution to this comment is either:

1. keep the "verbose" accumulation
2. include `AddAssign` in the trait bounds and use `+=`

I don't know what the cons are to 2. Are compilers smart enough to accumulate, here?

Rust/src/anomaly_score_visitor.rs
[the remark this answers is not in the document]
jmazanec15: Makes sense. Thanks

Rust/src/random_cut_forest.rs
cswiercz: #206

Rust/src/anomaly_score_visitor.rs
jmazanec15: I think the verbose accumulation is fine

jmazanec15 (approved):

cswiercz:
errors

Plot of score difference between Rust and Java on the 12-dimensional "I02" dataset in the [St. Petersburg INCART](https://physionet.org/content/incartdb/1.0.0/) database of EKG signals. The first standard deviation of relative error is approximately 4%. Other experiments also show approximately 5% score similarity between the algorithms. Mean relative error is within 1% of zero.

cswiercz:
errors

Same data but with 500 trees. 2% rel error stddev and 1% rel. error mean.

cswiercz:
@jotok Based on the experiments above, I'm convinced by algorithm correctness. I'd appreciate comment on the design...modulo architecture and generalization improvements that will be made in V2.

Rust/src/anomaly_score_visitor.rs
```diff
@@ -0,0 +1,196 @@
...
+            (one - separation_probability) * self.anomaly_score;
+    }
+
+    /// Normalize and return the anomaly score computed by the visitor.
+    ///
+    /// The anomaly score computed by the visiting process, via
+    /// [`Self::accept_leaf()`] and [`Self::accept()`], is normalized using the
+    /// mass of the tree before returning. This is so that the resulting anomaly
+    /// score is independent of the number of samples in the tree.
+    pub fn get_result(&self) -> T {
```
jotok, line 116: minor: it seems like we're mostly not using "get" in our getter method names. Should we change this to just `result()`?

Rust/src/random_cut_forest.rs
```diff
@@ -90,6 +92,60 @@ impl<T> RandomCutForest<T>
...
+    pub fn anomaly_score(&self, point: &Vec<T>) -> T {
+        let anomaly_score: T = self.trees.iter()
+            .map(|t| self.tree_anomaly_score(point, t.tree()))
+            .sum();
+        anomaly_score / T::from(self.num_trees()).unwrap()
+    }
+
+    /// Compute the anomaly score from a particular tree. Used by
+    /// ['Self::anomaly_score'].
+    fn tree_anomaly_score(&self, point: &Vec<T>, tree: &Tree<T>) -> T {
```
jotok, line 132: What's the rationale for putting this here as opposed to the tree implementation?

jotok (approved):

Rust/src/anomaly_score_visitor.rs
cswiercz: I guess I saw this `get_result()` not so much as a getter but as a "finalize" like thing. After all, `get_result()` usually does more than just return a value; here it does some normalization. From my reading, I don't see a standard w.r.t. the visitor pattern.

More generally, I chose to omit the `get_`s in the type getters just for brevity. Also, I've seen a number of Rust devs use `field()` and `field_mut()` instead of `get_field()` and `set_field()`. So the question is, do we try to match the Java implementation of RCF or do we write things in the way Rust developers expect.

Anyway, that's my perspective. Given that, willing to adjust in the nomenclature based on your thoughts.

Rust/src/random_cut_forest.rs
cswiercz: Great question. When I wrote this I put it here without much thought. Of course, it could go into `anomaly_score(), itself. Just wanted to break it out.

That said, there's a bigger problem here: for each "score"/visitor there is an operation on each of the trees and then some kind of combination logic. In the Java version, `traverseForest()` accepts the point along with a visitor (factory), accumulator, and finisher. The former is the visitor, as usual. The latter two are the post-processing.

As we add more scoring algorithms, do we want to add a corresponding function to *both* `Forest` and `Tree`? I detect a code smell in this case. The `traverseForest()` approach seems cleaner, but I wonder if a `ScoringAlgorithm` type would be more appropriate. The `ScoringAlgorithm` (name pending) would include each of the things used by `traverseForest()` plus more. 

At this point, it may be worth forking this discussion out to a tracking issue. I think it's an important problem because I want the addition of new scoring algorithms to be as easy as possible to implement for the developer.

cswiercz: Actually, #206 already exists. Will include some of this discussion, there.

Rust/src/anomaly_score_visitor.rs
jotok: My vote is to follow rust conventions when they conflict with what's in the Java library.

Rust/src/random_cut_forest.rs
jotok: Right, so far we've been able to get by with just generic traversal methods defined on the tree class. Methods like "get anomaly score" are just UX conveniences for invoking the visitor pattern plus aggregation. Adding new algorithms shouldn't require adding new methods to the tree class.

cswiercz: > Adding new algorithms shouldn't require adding new methods to the tree class.

Agreed.

Rust/src/anomaly_score_visitor.rs
cswiercz: Right on. In that case I'll leave `get_result()` as is. There is an opportunity to revisit in #206 .

PR #193, opened 2021-05-13
Normalize names in the state package

jotok:
- Match names in the corresponding model classes.
- Use the "Enabled" suffix for boolean flags where it improves
  readability.
- Use "compressed" in state classes versus "isCompressionEnabled" in
  mapper classes.
- Use "isPartialTreeState" in state classes versus
  "isPartialTreeStateEnabled" in mapper classes.
- Store precision as a String instead of a boolean.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestMapper.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestMapper.java
@@ -71,40 +71,40 @@ public class RandomCutForestMapper
      * {@link RandomCutForestState} object produced by the mapper. Tree state is not
      * saved by default.
      */
-    private boolean saveTreeState = false;
+    private boolean saveTreeStateEnabled = false;
 
     /**
      * A flag indicating whether the point store should be included in the
      * {@link RandomCutForestState} object produced by the mapper. This is saved by
      * default for compact trees
      */
-    private boolean saveCoordinatorState = true;
+    private boolean saveCoordinatorStateEnabled = true;
 
     /**
      * A flag indicating whether the samplers should be included in the
      * {@link RandomCutForestState} object produced by the mapper. This is saved by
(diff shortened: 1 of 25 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestMapper.java
```diff
@@ -123,64 +123,65 @@
...
         }
 
         if (forest.isCompact()) {
-            if (saveCoordinatorState) {
-                PointStoreCoordinator pointStoreCoordinator = (PointStoreCoordinator) forest.getUpdateCoordinator();
+            if (saveCoordinatorStateEnabled) {
+                PointStoreCoordinator<?> pointStoreCoordinator = (PointStoreCoordinator<?>) forest
+                        .getUpdateCoordinator();
                 PointStoreState pointStoreState;
                 if (forest.getPrecision() == Precision.FLOAT_32) {
```
sudiptoguha, line 163: forest.getPrecision().name()?? It's probably ok in the model -> state direction.

sudiptoguha (approved):

Java/core/src/main/java/com/amazon/randomcutforest/state/RandomCutForestMapper.java
jotok: `precision` in the forest class has the enum type `Precision`, so comparing to the enum value `FLOAT_32` is correct here. The `.name()` returns the string representation of the enum constant. In the state classes, I made the `precision` field a String. My reasoning was that I didn't want to break protostuff if we added new enum values later (I don't know if that would break it or not, but it didn't seem worth taking the chance).

sudiptoguha: makes sense.

PR #196, opened 2021-05-13
Change Spotless configuration to use Google AOSP style

jotok:
We were previously maintaining our own list of formatting rules, which were based on a previously existing checkstyle file. Rather than keep our own set of custom formatting rules, we can follow Google AOSP style (as defined by the google-java-format plugin), and that will make our codebase consistent with any other project using this built-in default.


jotok:
We don't want to remove support for building with Java 1.8 at this time.

PR #203, opened 2021-05-17
reference test in compact trees

sudiptoguha:
*Description of changes:* tests point equality via references for deletes.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -15,18 +15,18 @@
 
 package com.amazon.randomcutforest.tree;
 
-import static com.amazon.randomcutforest.CommonUtils.checkArgument;
-import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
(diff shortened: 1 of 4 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -510,6 +510,11 @@ public int getRootIndex() {
         return intValue(root);
     }
 
+    @Override
+    protected boolean referenceEquals(Integer oldPointRef, Integer pointRef) {
+        return oldPointRef.equals(pointRef);
```
jotok, line 515: Would there be any value to adding short-circuiting here? I.e., will the old ref ever be the same as the new ref?
```
return oldPointRef == pointRef || oldPointRef.equals(pointRef);
```

jotok (approved):

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: There could be some value for classes which are more complicated than Integer. Perhaps we can add it to those classes?

PR #217, opened 2021-05-25
[Rust] Create a Visitor trait to abstract out scoring algorithms

### LINKED ISSUE #206: [Rust] Design pattern for multiple scoring methods
Anomaly scoring is just one use of the data sketch offered by RCF. We want to develop some kind of generic framework for easily adding new visitors. There are several components and stages of the process to consider:

* a `Visitor` is applied to each tree, resulting in some output type `T`
* the output across every tree is processed in some way
  * `anomaly_score()` simply averages the results
  * `impute_missing_values()` returns the median result across all trees.
  * `nearest_neighbor()` does something else

Observations from the Java version: `getAnomalyScore()` and `getNearNeighborsInSample()` follow the `traverseForest()` structure pretty well but `imputeMissingValues()` breaks the pattern. Something more general might be a solution? Also, each algorithm is spread out between what's written in `RandomCutForest`, `traverseForest()`, and in the case of nearest neighbors `Neighbor.collector()`. What if we just put all of these instructions into a new type or separate `impl`?

While we can draw some inspiration from the Java implementation, Rust is not an object-oriented language. Trying to verbatim copy an inheritance pattern using traits is not always possible. (See [this article](https://blog.logrocket.com/what-you-cant-do-in-rust-and-what-to-do-instead/).) Finally, macros may make adding some kind of `ScoringAlgorithm` type to the RCF easier.

cswiercz:
*Issue #, if available:*
Closes #206 

*Description of changes:*

Most scoring algorithms make use of a visitor design pattern. The `Visitor` trait makes it easy to perform node traversals with new algorithms. We also add a `Tree::traverse()` method that makes use of any type that implements `Visitor`. This addition makes the implementation of `anomaly_score()`much cleaner.

Note that in #206 there is some discussion about a general `Algorithm` type/trait with idea that such a type would simplify algorithm development. This new `Visitor` trait alone seems to simplify things enough, at least for now. Therefore, this PR can potentially close #206.


cswiercz:
In this redesign I also address Issue #219 . Instead of reversing tree iterators, which made it more difficult and more expensive to track node depth, we have two separate `Tree` functions:

* `iter()`: returns a forward iterator on nodes. Useful for debugging.
* `traverse()`: applies a visitor in reverse iteration on nodes. (Starting at the nearest leaf.)

The documentation is updated to explain this difference.

Rust/src/algorithm/visitor.rs
```diff
@@ -0,0 +1,38 @@
...
+/// [`Tree::traverse()`](crate::tree::Tree::traverse).
+///
+pub trait Visitor<T> {
+
+    /// The output type of the visitor.
+    type Output;
+
+    /// Initialize visitor calculation at the leaf node nearest a query point.
+    ///
+    /// This function is called on the leaf node nearest to an input query
```
jmazanec15, line 17: Because the leaf is passed in, the description shouldn't allude to which leaf is passed in. I do not think it necessary here to make assumptions on that leaf.

Rust/src/lib.rs
```diff
@@ -28,15 +28,20 @@
 //!     .collect();
 //! ```
 //!
+//! ## Scoring Algorithms
+//!
+//! The [`algorithm`] module contains tools and components for creating new
```
jmazanec15, line 33: algorithm seems like it could refer to a lot in the RCF library. Why not call the module scoring or scorer?

Rust/src/tree/tree.rs
```diff
@@ -198,22 +202,63 @@ impl<T> Tree<T>
...
-    pub fn traverse<'a>(&'a self, point: &'a Vec<T>) -> NodeTraverser<'a, T> {
-        NodeTraverser::new(self, point)
+    pub fn iter<'a>(&'a self, point: &'a Vec<T>) -> NodeIterator<'a, T> {
+        NodeIterator::new(self, point)
+    }
+
+    /// Apply a visitor to a tree traversal with a query point.
+    ///
+    /// Given a query point and a visitor, we apply the visitor to the nodes
+    /// of the tree starting at the leaf nearest to the query and traversing up
```
jmazanec15, line 226: A little bit confused. Where is the logic to start at the nearest leaf? This appears to start at the root.

Rust/src/algorithm/visitor.rs
cswiercz: Good call. I agree that the visitor is agnostic to the traversal algorithm.

Rust/src/lib.rs
cswiercz: I went with `algorithm` since forecasting and KNN are not exactly producing "scores". However, I didn't spend all that much time thinking of a name other than "algorithm". Given my comment about scores I'm open to other suggestions. Good names are half the battle.


I see, though, that I write "Scoring Algorithms" in this section of the documentation. I will make the wording more consistent.

cswiercz: I would especially like @jotok 's opinion on names. My thought is that the `algorithm` (or whatever) submodule contains all the "good stuff" for researchers.

Rust/src/tree/tree.rs
cswiercz: Great question. If you look at `traverse_helper()` you'll see the recursion step:

```rust
self.traverse_helper(point, visitor, next_node_key, depth + One::one());
visitor.accept(&node, depth);
```

This means that we recurse down to the nearest leaf node and execute `accept_leaf()` first before following the stack back up to execute `accept()` at each of the intermediate internal nodes.

It's different than `Tree::iter()`, I know. I kept `iter()` mostly because it helps with some tests and I couldn't figure out a *good way* to return a node iteration in the same order as `traverse()`.

jmazanec15: Ah I see makes sense.

jmazanec15: Okay, I think I am missing something. Why cant the scoring start from the root and compute as it goes down to the nearest leaf in order to avoid recursion?

cswiercz: Really good question. I was writing up a response but realized that I it was (a) getting really long and should be turned into a manuscript of some kind and (b) not all that convincing that one direction is better than another. Still, a good exercise to carry out.

@sudiptoguha Do you have a one-sentence argument for this?

sudiptoguha: Probably not a one-sentence :) The discussion is perhaps top-down versus bottom up? In general bottom up is more powerful and more expressive. One example is that in the current Java version, the bounding boxes are recreated on the fly and once a point is inside a bounding box then there is no change to score/density/most of envisioned visitor applications (forecasting is a different discussion altogether). If you start from the top, you need the bounding box at the top -- continuing the argument, you need bounding boxes everywhere. If you go bottom up -- you only need to only ever build the bounding boxes you need. For N leaves, the number of possible internal bounding boxes is N-1 and the bounding boxes are 2D size (max, min for each of the D dimensions). So the bounding boxes themselves are almost 2ND in size -- note the amount of basic information is ND (for the points, ignoring the shingling). So there you have 66% saving (ND versus ND + 2ND) savings in space if you go bottom up and build boxes on the fly. And you can adjust the space/throughput tradeoff by storing some of the boxes (as the Java version does).

jotok: The anomaly score algorithm is defined recursively at level n by:
```
(1 - p) score_n + p score_{n + 1}
```
where p is the probably that the query point (i.e., the point being scored) would be separated from the bounding box at level n on the query path by a random cut. The original implementation of RCFs didn't have a generic traversal function, it just had a recursive anomaly score function that mirrored the score computation. We kept the leaf-to-root logic of the anomaly score method when we refactored the library to use the visitor pattern. So the answer is partly historical, but algorithms that we added to the library later (like density estimation and imputation) followed the same pattern of computing a value at the leaf node first and then updating that value by traversing back to the root. That said, it's not a requirement that the implementation of the traversal function be recursive, and there's no reason we couldn't also define a traversal method that visits nodes in root to leaf order.

jotok: I would also add that so far we haven't seen much downside in the recursive implementation, since the recursion depth will be equal to the tree depth, which is on the order of log-sample-size. I did one test a long time ago (when the library looked a lot different than it does now) where I rewrote the traversal method to be iterative instead of recursive, but it was actually slower than the recursive version. So from my perspective I'm open to an iterative implementation, changing this part of the library hasn't been a high priority.

Rust/src/lib.rs
jotok: I'm sympathetic Chris's point of view, because in my mind I think of RCF as a data structure on which different scoring algorithms are implemented, but I think Jack's right that this will probably be a point of confusion to new users of the library. What about `visitors.rs`? Barring that, I think some version of "scoring" is fine, if we take an expansive view of what a "score" can be.

Rust/src/tree/tree.rs
cswiercz: @sudiptoguha Thanks for the response and for bringing up the dynamic bounding boxes as an example. Very clear reason to use the bottom up approach.

@jotok Glad that you experience with the scoring methods sort of validates that (at least specific to *anomaly* scoring) traversal direction is somewhat of a choice rather than a requirement. Of course this doesn't apply to some of the other scoring algorithms. Interesting bit about the performance difference.

Rust/src/lib.rs
cswiercz: I like `visitors` as a submodule; `scoring` has the risk of making a narrow first impression of what these things mean. My only issue is that the visitor is only one component of an algorithm implementation. However, I'm still fuzzy on this distinction so for now a visitors sub-module is sufficient.

Rust/src/visitor/visitor.rs
```diff
@@ -0,0 +1,38 @@
+use crate::{Internal, Leaf};
+
+/// Visitors on random cut trees.
+///
+/// Many random cut forest scoring algorithms use a visitor design pattern on
+/// the nodes of a tree. The nodes typically run from the leaf closest to the
+/// input query point along the path to the root node such as in the function
+/// [`Tree::traverse()`](crate::tree::Tree::traverse).
+///
+pub trait Visitor<T> {
```
jotok, line 10: Do we need to define both `T` and `Output` in the trait definition?

cswiercz: I believe so. `T` is (almost) always the float type. You can see it show up in the definition of this trait's `accept()` function:

```rust
fn accept(&mut self, node: &Internal<T>, depth: T);
```

I use a separate associated output type to handle the attribution visitor case. In many situations the compiler can infer `T`. For example, in `RandomCutForest::anomaly_score()` you see the construction line,

```rust
let mut visitor = AnomalyScoreVisitor::new(sampled_tree.tree(), point);
```

Since `sampled_tree` is of type `SampledTree<T>` the `T` part is inferred. That is, I don't need to write `let mut visitor: AnomalyScoreVisitor<T>`. The type of the output is defined in the visitor implementation. For example, in the definition of `AnomalyScoreVisitor` we have,

```rust
impl<'a, T> Visitor<T> for AnomalyScoreVisitor<'a, T> where
    T: Float + One + Sum + Zero
{
    type Output = T;
    // ...
}
```

When I finally create an `AttributionScoreVisitor` I would set `type Output = Vec<T>`. This output type can also be inferred in some cases since `Self::Output` is the output type of `get_result()`.

cswiercz: One follow up question, though, is whether `depth` should be of type `T`. I do so for convenience since depth is used to update the scalar anomaly score which is of type `T`. But it could be of type `u16`, `u32`, or whatever. I'm leaning towards leaving it at `T` but would like to hear your opinion.

jotok (approved):

PR #220, opened 2021-06-02
full reproducibility on serialization

sudiptoguha:
*Description of changes:* enables full reproducibility on state serialization/deserialization. Also fixes minor bugs/omissions.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
@@ -15,12 +15,14 @@
 
 package com.amazon.randomcutforest.sampler;
 
-import static com.amazon.randomcutforest.CommonUtils.checkArgument;
-import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
+import com.amazon.randomcutforest.config.Config;
 
 import java.util.Random;
 
-import com.amazon.randomcutforest.config.Config;
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_INITIAL_ACCEPT_FRACTION;
+import static com.amazon.randomcutforest.RandomCutForest.DEFAULT_SAMPLE_SIZE;
 
 public abstract class AbstractStreamSampler<P> implements IStreamSampler<P> {
     /**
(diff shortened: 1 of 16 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
```diff
@@ -158,4 +184,91 @@ public void setMostRecentTimeDecayUpdate(long index) {
...
+
+        ReplayableRandom(long randomSeed) {
+            this.randomSeed = randomSeed;
+        }
+
+        ReplayableRandom(Random random) {
+            this.testRandom = random;
+        }
+
+        double nextDouble() {
```
jotok, line 216: We should put a note somewhere that the sampler is not concurrency safe. There's already a disclaimer in the README, but we might want to make it more expansive and say something like "assume that classes in this library are not concurrency safe unless explicitly state otherwise in the documentation", and then maybe some words about how we support parallelism by synchronizing access to shared state.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
```diff
@@ -158,4 +184,91 @@ public void setMostRecentTimeDecayUpdate(long index) {
...
+            this.random = random;
+            return (T) this;
+        }
+
+        public T maxSequenceIndex(long maxSequenceIndex) {
+            this.maxSequenceIndex = maxSequenceIndex;
+            return (T) this;
+        }
+
+        public T mostRecentLambdaUpdate(long sequenceIndexOfMostRecentLambdaUpdate) {
```
jotok, line 259: `Lambda` should be `TimeDecay` to match.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -418,104 +290,60 @@ private void swapWeights(int a, int b) {
         }
     }
 
-    public static class Builder<T extends CompactSampler.Builder<T>> {
+    public static class Builder extends AbstractStreamSampler.Builder<Builder> {
```
jotok, line 293: I think this change means that we won't be able to further subclass CompactSampler.Builder and have it work as expected.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -542,4 +370,24 @@ public CompactSampler(Builder<?> builder) {
             }
         }
     }
+
+    public CompactSampler(int sampleSize, double lambda, Random random, boolean storeSequences) {
```
jotok, line 374: Did you mean to delete the old constructors?

Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
sudiptoguha: Agree.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
sudiptoguha: Done. We probably should have a more focused PR to have all lambda -> timeDecay elsewhere.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: yes, but then the number of files became too large and I did not want to change the tests simultaneously. Lets have a separate PR for removing these constructors in the tests (and in the main forest class).

jotok (approved):

PR #230, opened 2021-06-03
Merging development branch into 2.0 branch

jotok:
This is the first step toward creating a 2.0 release candidate. This PR merges the development branch compact-rcf into the 2.0 branch. Features added in the compact-rcf branch:

- New compact implementations of trees and samplers
- Configurable bounding box cache
- Internal shingling
- State classes (POJOs) used to define model state and to facilitate serialization
- Support for dynamic configuration at runtime


Java/benchmark/src/main/java/com/amazon/randomcutforest/ForestUpdateExecutorBenchmark.java
```diff
@@ -30,12 +29,18 @@
...
+import com.amazon.randomcutforest.executor.ParallelForestUpdateExecutor;
+import com.amazon.randomcutforest.executor.PassThroughCoordinator;
+import com.amazon.randomcutforest.executor.PointStoreCoordinator;
+import com.amazon.randomcutforest.executor.SamplerPlusTree;
+import com.amazon.randomcutforest.executor.SequentialForestUpdateExecutor;
+import com.amazon.randomcutforest.sampler.SimpleStreamSampler;
+import com.amazon.randomcutforest.store.PointStoreDouble;
 import com.amazon.randomcutforest.testutils.NormalMixtureTestData;
-import com.amazon.randomcutforest.tree.CompactRandomCutTree;
+import com.amazon.randomcutforest.tree.CompactRandomCutTreeDouble;
```
cswiercz, line 42: I've seen this pattern where a `Class` is defined with generics and then `ClassFloat` and `ClassDouble` are defined as `Class<float>` and `Class<double>`. Though, there are some instances like in `IndexManager` where the float type is not exactly the same as generic: `IndexManager<float[], float[]>`. Nonetheless, Is there a reason why the user can't do this? That is, instead of having three versions of similar functionality why not have one generic one and the user just defines a `RandomCutForest<Double>` in their code? Or do these classes have float type-specific operations?

jotok: This is due to a Java limitation: Java doesn't support classes that are generic over unboxed primitive types. You can have `Class<Double>` with a capital `D` corresponding to the boxed type, but then every `Double` in the class is a pointer to a `double` value, and we take a hit for all the dereferencing. In the places where we created `ClassDouble` and `ClassFloat` versions of classes, it should either be because there was some math we wanted to do with primitive floats and doubles, or else because the class depends on another class that does math with primitive floats and doubles. That said, there was a lot of refactoring that happened after these classes were first created, and it would be good to review all of them to make sure that we really need `Double` and `Float` versions still.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -106,7 +173,7 @@
     /**
      * Random number generator used by the forest.
      */
-    protected final Random rng;
+    protected Random random;
```
cswiercz, line 176: Is removing `final` so you can do full state restoration in serde?

Java/benchmark/src/main/java/com/amazon/randomcutforest/ForestUpdateExecutorBenchmark.java
sudiptoguha: Indeed, this is a plus for Rust and would not be needed there. The pattern is over primarily over [BoundingBox, Tree, PointStore, BoxCache] and then it spills over to the mapper classes. PointStore would need this precision setup (checking equality etc.). Now if the PointStore converts these to some standard (say double[]) to pass to the tree (since the tree coordinates the bounding boxes)  then there is multiple conversions of precision which seems like a bad idea (plus the precision of the random cuts becomes an issue in these conversions). That kind of indicates a split for the trees is needed. The boxes (basic class and the cached) are interesting -- it may be possible to refactor them (i.e., go via PointStore). But the Visitors were using the double precision anyways and we chose to keep the Visitors as intact as possible. So the refactoring would likely involve the Visitors. Right now the precision set up is internal -- the input or the visitors see double[]. Conversions are done exactly at two places (once in input, at PointStore and once in egress, while exposing to the visitor in CompactNodeView).

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: yes (mostly and some no :)). The full state restoration goes via the randomSeed (which is how the forest instantiates the trees). But for the tests -- they mock up a Random() and test the different codepath (which is very useful). So the removal of final is because we keep both the seed and the Random(). That being said - this was happening because there were a thousand constructors, which is now changed thx to Josh's last PR and they all go through the builders. Maybe this is a cleanup we can have post merge.

Java/benchmark/src/main/java/com/amazon/randomcutforest/ForestUpdateExecutorBenchmark.java
cswiercz: (Thanks for the explanation. FWIW, I wasn't trying to flex any Rust muscles nor did I know that this was a Java language limitation.)

@jotok I can help look for the places you mentioned where the two versions might not be needed.

@sudiptoguha It's best to avoid multiple conversations for floating point accuracy and performance reasons, so I'm not advocating in any way to remove the `Float` vs. `Double` distinction. But having to handle both types explicitly is a bit of a scaling problem that, apparently, cannot be avoided.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
cswiercz: Thanks, got it. Not to mention that there may be a use case to reinitializing the rng in an existing model.

sudiptoguha: Yes - it may be useful to rerandomize (or the opposite, sync) across different forest :) The models encapture stochastic process states, so these operations may see use via the dynamic setConfig options in a 2.X

Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ *
+ * @param <PointReference> The internal point representation expected by the
+ *                         component models in this list.
+ * @param <Point>          The explicit data type of points being passed
+ */
+
+public interface IComponentModel<PointReference, Point>
+        extends ITraversable, IUpdatable<PointReference>, IDynamicConfig {
+}
```
cswiercz, line 31: What is the etymology of "`IComponentModel`". I see here that it's something that's traversable, updatable, and configurable but seeing it used elsewhere it's not clear what is the purpose.

Also, why does it need `Point`? `IUpdatable` needs `PointReference` but I don't see anything to which the `Point` type is passed.

cswiercz:
General question: about how many classes on average implement `IThing` for each interface `IThing`?

Java/core/src/main/java/com/amazon/randomcutforest/IComponentModel.java
sudiptoguha: Let me answer the easier one: the trees need Point because that is the explicit datatype exchanged with the pointstore. Now the interfaces on top of tree do not really need to know what Point is -- but there should be type check somewhere. 
For the harder question: it may make sense to break the pattern of one-tree-one-sampler down the road. Bootstrap sampling being the obvious use case. Other example can be grouped trees -- say 10 trees per sampler, and a few samplers. So the IComponent is a place where w can branch out.

sudiptoguha:
> General question: about how many classes on average implement `IThing` for each interface `IThing`? How many functions take in the `IThing` instead of a (slightly more) concrete `AbstractThing`?

AbstractThings are actually the main bulks of code -- the concrete subclasses are basically handling precision (copy, equality, etc.). AbstractThings are specific implementation (making caching decisions, storage decisions) that can affect heap. So if one wants a specific IThing that does something (no pun intended) then they can branch off without touching this code.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -290,6 +572,79 @@ public int getThreadPoolSize() {
         return threadPoolSize;
     }
 
+    // TODO change in next PR with the mappers and tests
```
wnbts, line 575: nothing. just raise this todo in the main class

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -290,6 +572,79 @@ public int getThreadPoolSize() {
...
+    }
+
+    /**
+     * Returns a clean deep copy of the point. Current clean-ups include changing
+     * negative zero -0.0 to positive zero 0.0.
+     *
+     * @param point The original data point.
+     * @return a clean deep copy of the original point.
+     */
+    public static double[] cleanCopy(double[] point) {
```
wnbts, line 591: minor. this public method should not belong to this forest class.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -625,64 +1025,55 @@ public DensityOutput getSimpleDensity(double[] point) {
...
      * @return A point with the missing values imputed.
      */
-    public double[] imputeMissingValues(double[] point, int numberOfMissingValues, int[] missingIndexes) {
-        checkArgument(numberOfMissingValues >= 0, "numberOfMissingValues must be greater than or equal to 0");
-
-        // We check this condition in traverseForest, but we need to check it here s
-        // wellin case we need to copy the
-        // point in the next block
-        checkNotNull(point, "point must not be null");
+    public ArrayList<double[]> getConditionalCentralField(double[] point, int numberOfMissingValues,
```
wnbts, line 1028: the return data type should be abstract. and the comment needs update.

wnbts (commented):
the main rcf class has a large public interface now. given it's likely the main entry point, i think it's better to hide unnecessary methods for internal/implementation/utility from the public so it has a clean api, and avoids unwanted dependencies that might require major version change in the future.

Java/benchmark/src/main/java/com/amazon/randomcutforest/ForestUpdateExecutorBenchmark.java
jotok: I think that in Java we can support 32-bit and 64-bit floating point numbers, and say that if you want more control than that then you'll have to use a different implementation (i.e., Rust). So this bit of ugliness should at least not get any worse than it currently is.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -140,58 +234,210 @@
...
     /**
-     * Number of threads to use in the threadpool if parallel execution is enabled.
+     * Number of threads to use in the thread pool if parallel execution is enabled.
      */
     protected final int threadPoolSize;
+    /**
+     * A string to define an "execution mode" that can be used to set multiple
+     * configuration options. This field is not currently in use.
+     */
+    protected String mode;
```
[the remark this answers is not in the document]
sudiptoguha: issue 237

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: issue 239. However the cleanCopy is needed whenever the point is submitted to the forest (for update or scoring of any sort, we need to check for 0.0 and -0.0 in all of them). But it'll be good to broaden the discussion to other methods as well as protected classification instead of public.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: issue 238. The return datatype is double[] for each tree and we just present a list. Note though that this is not the precision issue -- all input/output is double/double[]. The float[]/double[] is internal.

sudiptoguha:
Merging the large commit and the issues can be picked off one of by one.

PR #236, opened 2021-06-08
[Rust] Implement shared point stores

### LINKED ISSUE #225: [Rust v2.0] Shared point store
A major feature of Rust v2.0 is a shared point store: a single point store that can be referenced from all trees to reduce memory duplication. In addition, this point store should have the option of being "shingle-aware": include additional optimization for overlapping shingles. (Some designs solve both situations simultaneously.)

cswiercz:
*Issue #, if available:*

Closes #225.

*Description of changes:*

We offer a shared point store trait as well as several implementations of efficient shared point stores.

**Note**

This is an in-progress PR. I released it early to start a discussion about shared point store API design and implementation considerations. Do not merge until the work is complete but please comment.

Things that need to happen before a merge:
- [ ] Update `Tree`, `SampledTree`, and `RandomCutForest` to use shared point stores
- [ ] (optional?) Allow users to specify a point store using the `RandomCutForestBuilder`
- [ ] Create benchmarks in a `benches/` directory. This is somewhat standardized with cargo. See [here](https://doc.rust-lang.org/stable/cargo/reference/cargo-targets.html?#benchmarks) for more information.

### DIFF
--- a/Rust/src/lib.rs
+++ b/Rust/src/lib.rs
@@ -47,6 +47,8 @@ pub use sampler::{SamplerResult, StreamSampler, WeightedSample};
 mod sampled_tree;
 pub use sampled_tree::SampledTree;
 
+mod shared_point_store;
+
 mod store;
 pub use store::{NodeStore, PointStore};
 
--- a/Rust/src/shared_point_store.rs
+++ b/Rust/src/shared_point_store.rs
@@ -1,0 +1,147 @@
+use std::collections::HashMap;
+
+/// Trait for data structures representing shared point stores.
+///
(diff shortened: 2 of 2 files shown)


cswiercz:
Summary of changes so far:

* Updated the `PointStore` trait to explicitly output `&[T]` instead of allowing for different output types via the old associated type. Point stores should only output points and that's it.
* Created a shingle-aware points store. Check the docs using `cargo doc --open` or just look at the type-level comments to see an explanation of how it works.

Next up:
* benchmarks: will likely use [Criterion](https://bheisler.github.io/criterion.rs/book/getting_started.html) for measuring performance.

PR #240, opened 2021-06-09
2.0

sudiptoguha:
Issue #, if available: 237, some of 238 (the specific examples), and 231

*Description of changes: General fixes and refactoring of ImputeVisitors with documentation.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -572,7 +572,6 @@ public class RandomCutForest {
         return threadPoolSize;
     }
 
-    // TODO change in next PR with the mappers and tests
     public IStateCoordinator<?, ?> getUpdateCoordinator() {
         return stateCoordinator;
     }
@@ -1023,21 +1022,25 @@ public class RandomCutForest {
      *                              values in the point. The length of the array
      *                              should be greater than or equal to the number of
      *                              missing values.
+     * @param centrality            a parameter that provides a central estimation
+     *                              versus a more random estimation
      * @return A point with the missing values imputed.
      */
-    public ArrayList<double[]> getConditionalCentralField(double[] point, int numberOfMissingValues,
-            int[] missingIndexes) {
+    public ArrayList<double[]> getConditionalField(double[] point, int numberOfMissingValues, int[] missingIndexes,
+            double centrality) {
         checkArgument(numberOfMissingValues > 0, "numberOfMissingValues must be greater than 0");
         checkNotNull(missingIndexes, "missingIndexes must not be null");
         checkArgument(numberOfMissingValues <= missingIndexes.length,
                 "numberOfMissingValues must be less than or equal to missingIndexes.length");
+        checkArgument(centrality >= 0 && centrality <= 1, "centrality needs to be in range [0,1]");
 
         if (!isOutputReady()) {
             return new ArrayList<>();
         }
 
-        IMultiVisitorFactory<double[]> visitorFactory = (tree, y) -> new ImputeVisitor(y, numberOfMissingValues,
-                tree.projectMissingIndices(transformIndices(missingIndexes, point.length)));
+        int[] liftedIndices = transformIndices(missingIndexes, point.length);
+        IMultiVisitorFactory<double[]> visitorFactory = (tree, y) -> new ImputeVisitor(y, tree.projectToTree(y),
+                liftedIndices, tree.projectMissingIndices(liftedIndices), 1.0);
 
         Collector<double[], ArrayList<double[]>, ArrayList<double[]>> collector = Collector.of(ArrayList::new,
                 ArrayList::add, (left, right) -> {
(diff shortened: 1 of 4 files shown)


sudiptoguha:
man package builds without errors locally

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -1023,21 +1022,25 @@ public DensityOutput getSimpleDensity(double[] point) {
...
      */
-    public ArrayList<double[]> getConditionalCentralField(double[] point, int numberOfMissingValues,
-            int[] missingIndexes) {
+    public ArrayList<double[]> getConditionalField(double[] point, int numberOfMissingValues, int[] missingIndexes,
+            double centrality) {
         checkArgument(numberOfMissingValues > 0, "numberOfMissingValues must be greater than 0");
         checkNotNull(missingIndexes, "missingIndexes must not be null");
         checkArgument(numberOfMissingValues <= missingIndexes.length,
                 "numberOfMissingValues must be less than or equal to missingIndexes.length");
+        checkArgument(centrality >= 0 && centrality <= 1, "centrality needs to be in range [0,1]");
```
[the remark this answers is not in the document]
sudiptoguha: A goal/tenent of the Java RCF2.0 version is to set up a framework that is futuristic and extensible. So that there can be multiple improvements in a continuous fashion without having a large version upgrade. So from that spirit, the question flows down to "what features should we have so that future changes would not be blocked". The biggest blocker in this context, if of course the serialized state, and the second biggest blocker is the API used for specific functions. If we do see "obvious next asks" (which may take some time to post, and perhaps await an user based on effort to impact ratio) then we should have that. So that's a long winded way of saying "specific use of the parameter would follow in subsequent PRs, based on need/ask". In terms of likelihood based interpretation, the current imputeVisitor does spell that out without changing the functionality.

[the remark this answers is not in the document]
cswiercz: Got it. Thanks for the clarification.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -32,26 +32,65 @@
...
  */
 public class ImputeVisitor implements MultiVisitor<double[]> {
-    private final boolean[] missing;
-    private final boolean[] liftedMissing;
-    private double[] queryPoint;
-    private double[] liftedPoint;
-    private double rank;
+
+    // default large values for initialization; consider -ve log( 0 )
+    public static double DEFAULT_INIT_VALUE = Double.MAX_VALUE / 4;
```
jotok, line 37: Why this choice specifically? Because we need a big number but not the biggest number?

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -94,14 +131,8 @@ public ImputeVisitor(double[] queryPoint, int numberOfMissingIndices, int[] miss
...
         this.liftedMissing = Arrays.copyOf(original.liftedMissing, original.liftedPoint.length);
-        rank = 10.0;
-    }
-
-    /**
-     * @return the rank of the imputed point in this visitor.
-     */
-    public double getRank() {
-        return rank;
+        anomalyRank = Double.MAX_VALUE / 2;
```
jotok, line 134: Why doesn't this use `DEFAULT_INIT_VALUE`? If the point is that we want to use a value that's strictly larger than in the visitor we're copying, won't that fail if the visitor has to split twice (i.e., because there are 2 missing values in the input point)?

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -179,6 +215,14 @@ public boolean trigger(final INodeView node) {
         return missing[node.getCutDimension()];
     }
 
+    public double getRank() {
```
jotok, line 218: Given that we changed `rank` to `anomalyRank` for clarity, should we change the getter to match?

jotok:
In the file `.github/workflows/maven.yml`, if you change
```
on:
  pull_request:
    branches: 
      - main
      - compact-rcf
```
to
```
on:
  pull_request:
    branches: 
      - '*'
```
then we'll get the CI build for all pull requiests.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
sudiptoguha: Yes. We may make approximate comparisons of rank (say up to factors of 4) and this avoids any potential overflows.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
sudiptoguha: Oversight. Fixed.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
sudiptoguha: Yes, and perhaps make the modifier protected since the main goal foreseen is in eventual subclassing.

sudiptoguha:
maven.yml changed, but needs to be merged.

PR #244, opened 2021-06-11
tree tests

sudiptoguha:
*Description of changes:* Adds multiple new tree tests and fixes of small issues in newer (abstract tree) as well as older trees.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -17,8 +17,8 @@ package com.amazon.randomcutforest.tree;
 
 import static com.amazon.randomcutforest.CommonUtils.checkArgument;
 import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
-import static com.amazon.randomcutforest.CommonUtils.validateInternalState;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.concurrent.ArrayBlockingQueue;
 
@@ -162,6 +162,7 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
                 int currentNode = 0;
                 int leafcounter = maxSize - 1;
                 int[] map = new int[2 * maxSize - 1];
+                Arrays.fill(map, NULL);
                 ArrayBlockingQueue<Integer> nodeQueue = new ArrayBlockingQueue<>(maxSize - 1);
                 nodeQueue.add(root);
                 while (!nodeQueue.isEmpty()) {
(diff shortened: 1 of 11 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -171,7 +172,7 @@ public void reorderNodesInBreadthFirstOrder() {
                     if (isLeaf(leftChild)) {
                         // the parent is the current node and the indices, mass are being copied over
                         map[leftChild] = result.addLeaf(currentNode, getPointReference(leftChild), getMass(leftChild));
-                        validateInternalState(map[leftChild] == leafcounter, "incorrect state");
+                        // validateInternalState(map[leftChild] == leafcounter, "incorrect state");
```
jotok, line 175: Should we delete these instead of commenting them out?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -494,8 +497,8 @@ public PointReference addPoint(PointReference pointReference, long sequenceNumbe
             savedSiblingNode = followReference;
 
             // now iterative proceed up the tree and try to construct a cut
+            // currentUnmergedBox != null
```
jotok, line 500: Did you mean to delete this?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: The best solution would be a debug-flag. If we keep them uncommented, then these assertions are untestable (because they're holding!) and branch/code coverage is low. If we delete them then we lose the insight. We can delete them if we choose to.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: Same as above. This is a validation step.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
jotok: In a future PR, we can replace checks like these with asserts, which must be explicitly enabled at runtime by passing a command-line flag to the JVM.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
jotok: This looks like leftover code, because it doesn't invoke `checkState`, it doesn't throw an exception if the condition is false, etc.

sudiptoguha: Oops - actually the next line which is removed had the checkArgument(). Let's make both of these asserts()

PR #245, opened 2021-06-14
Adding tests and fixes

sudiptoguha:
*Description of changes:* general cleanup and extra tests for validation.


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractBoundingBox.java
```diff
@@ -197,6 +186,4 @@ public boolean contains(int dimension, double value) {
     @Override
     public abstract boolean equals(Object other);
```
jotok, line 187: Do we need to keep this in `AbstractBoundingBox`? the `equals` method is defined on `Object`.

jotok (approved):

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractBoundingBox.java
sudiptoguha: No. Can remove in next PR.

Issue #221, opened 2021-06-03
[Rust v2.0] Complete model configuration


cswiercz:
Complete the following list of static model configuration options. Some of these options may take a while and have linked tracking issues since they require significant work.

- [x] dimensions
- [x] number of trees
- [x] sample size
- [ ] output after
- [ ] shingle size
- [ ] store sequence indexes (boolean)
- [x] floating point precision
- [ ] parallel or sequential execution
- [ ] internal shingling (boolean)
- [x] time decay
- [ ] bounding box cache fraction

PR #246, opened 2021-06-15
[Rust] Add output_after threshold to RCF

jmazanec15:
*Issue #, if available:*
#221 

*Description of changes:*
Initially, the anomaly score that RCF outputs may not have very much meaning. For instance, given two points in the tree, can it really be said that there is enough information to determine if the third point is anomalous? This may cause a lot of false positives at the beginning of the stream.

This PR adds an "output_after" parameter to the RandomCutForest. If the number of observations against this RCF is less than or equal to this value, the RCF will simply return 0. By default, the builder sets this parameter to 0.

### DIFF
--- a/Rust/src/random_cut_forest.rs
+++ b/Rust/src/random_cut_forest.rs
@@ -46,6 +46,7 @@ pub struct RandomCutForest<T> {
     sample_size: usize,
     time_decay: f32,
     trees: Vec<SampledTree<T>>,
+    output_after: usize,
 }
 
 impl<T> RandomCutForest<T>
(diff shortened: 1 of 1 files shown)


cswiercz:
Could you add a test of applying `output_after` to`anomaly_score()`? Such a test is too verbose and particular for a doctest so you can add it to the `tests` module at the end of the file. Use a tiny forest. (The `sample_size` need not be greater than `output_after` to test `output_after`.)

jmazanec15:
> Could you add a test of applying `output_after` to`anomaly_score()`? Such a test is too verbose and particular for a doctest so you can add it to the `tests` module at the end of the file. Use a tiny forest. (The `sample_size` need not be greater than `output_after` to test `output_after`.)

Added test case

PR #247, opened 2021-06-15
PointStore tests and coverage

sudiptoguha:
*Description of changes:* Fixes to point store and tests that also addresses coverage. Plus cleanup of odd code that is not being used.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -119,7 +119,7 @@ public class RandomCutForest {
 
     /**
      * By default, the collection of points stored in the forest will increase from
-     * a small size, as needed to maximum capccity
+     * a small size, as needed to maximum capacity
      */
     public static final boolean DEFAULT_DYNAMIC_RESIZING_ENABLED = true;
 
@@ -187,7 +187,7 @@ public class RandomCutForest {
      */
     protected final int shingleSize;
     /**
-     * The input dimensions for known shinglesize and internal shingling
+     * The input dimensions for known shingle size and internal shingling
      */
     protected final int inputDimensions;
     /**
@@ -253,6 +253,12 @@ public class RandomCutForest {
     private boolean outputReady;
 
     /**
+     * used for initializing the compact forests
+     */
+    private final int initialPointStoreSize;
+    private final int pointStoreCapacity;
+
+    /**
      * An implementation of forest traversal algorithms.
      */
     protected AbstractForestTraversalExecutor traversalExecutor;
@@ -290,7 +296,7 @@ public class RandomCutForest {
 
     private void initCompactDouble(Builder<?> builder) {
         PointStoreDouble tempStore = PointStoreDouble.builder().internalRotationEnabled(builder.internalRotationEnabled)
-                .capacity(numberOfTrees * sampleSize + 1).initialSize(builder.initialPointStoreSize.get())
+                .capacity(pointStoreCapacity).initialSize(initialPointStoreSize)
                 .directLocationEnabled(builder.directLocationMapEnabled)
                 .internalShinglingEnabled(internalShinglingEnabled)
                 .dynamicResizingEnabled(builder.dynamicResizingEnabled).shingleSize(shingleSize).dimensions(dimensions)
@@ -317,7 +323,7 @@ public class RandomCutForest {
 
     private void initCompactFloat(Builder<?> builder) {
         PointStoreFloat tempStore = PointStoreFloat.builder().internalRotationEnabled(builder.internalRotationEnabled)
-                .capacity(numberOfTrees * sampleSize + 1).initialSize(builder.initialPointStoreSize.get())
+                .capacity(pointStoreCapacity).initialSize(initialPointStoreSize)
                 .directLocationEnabled(builder.directLocationMapEnabled)
                 .internalShinglingEnabled(internalShinglingEnabled)
                 .dynamicResizingEnabled(builder.dynamicResizingEnabled).shingleSize(shingleSize).dimensions(dimensions)
@@ -403,8 +409,8 @@ public class RandomCutForest {
         // " need shingle size > 1 for internal shingling");
         if (builder.internalRotationEnabled) {
             checkArgument(builder.internalShinglingEnabled, " enable internal shingling");
+            checkArgument(builder.compact, " option not supported, enable compact trees");
         }
-        checkArgument(!builder.compact || builder.compact, " option not supported, enable compact trees");
         builder.initialPointStoreSize.ifPresent(n -> {
             checkArgument(n > 0, "initial point store must be greater than 0");
             checkArgument(n > builder.sampleSize * builder.numberOfTrees || builder.dynamicResizingEnabled,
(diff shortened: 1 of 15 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -290,8 +290,7 @@ public RandomCutForest(Builder<?> builder) {
 
     private void initCompactDouble(Builder<?> builder) {
         PointStoreDouble tempStore = PointStoreDouble.builder().internalRotationEnabled(builder.internalRotationEnabled)
-                .capacity(numberOfTrees * sampleSize + 1).indexCapacity(builder.initialPointStoreSize.get())
-                .currentStoreCapacity(builder.initialPointStoreSize.get())
+                .capacity(numberOfTrees * sampleSize + 1).initialSize(builder.initialPointStoreSize.get())
```
jotok, line 293: How does `initialPointStoreSize` get initialized if the user doesn't supply a value?

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeState.java
```diff
@@ -38,6 +37,8 @@
     // to be used later
```
jotok, line 37: I think we can remove this comment.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -124,11 +123,13 @@
     @Override
     public int decrementRefCount(int index) {
         indexManager.checkValidIndex(index);
-        checkState(refCount[index] >= 0, " incorrect state ");
         if (refCount[index] == 1) {
             indexManager.releaseIndex(index);
-        }
 
+            // if (shingleSize > 1) {
```
jotok, line 129: Should we remove these lines?

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -160,13 +161,11 @@ public int decrementRefCount(int index) {
      */
     int takeIndex() {
         if (indexManager.isFull()) {
-            if (indexManager.getCapacity() < capacity) {
+            if (indexManager.getCapacity() < capacity && dynamicResizingEnabled) {
```
jotok, line 164: I think the error message in the `else` branch is misleading. The problem is that the point store has reached its capacity, right?

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -211,7 +208,8 @@ public int add(double[] point, long sequenceNum) {
 
         double[] tempPoint = point;
         if (internalShinglingEnabled) {
-            tempPoint = changeShingleInPlace(internalShingle, point);
+            // rotation is supported via the output and inoput is unchanged
```
jotok, line 211: typo in "input"

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -211,7 +208,8 @@ public int add(double[] point, long sequenceNum) {
...
+            tempPoint = changeShingleInPlace(internalShingle, point, false);
```
jotok, line 212: `changeShingleInPlace` is not a very descriptive name. change how?

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -433,25 +431,7 @@ public boolean isDirectLocationMap() {
...
-    }
-
-    /**
-     * used for serialization
-     * 
-     * @return the pointer to the stack of free indices
-     */
-    public int getFreeIndexPointer() {
-        return indexManager.getFreeIndexPointer();
+        return (internalShinglingEnabled) ? copyShingle() : null;
```
jotok, line 434: It seems like this should throw an exception instead of returning null. Would cause problems in other places?
```
checkState(internalShinglingEnabled, "internal shingling is not enabled");
return copyShingle();
```

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -710,6 +689,11 @@ public T capacity(int capacity) {
             return (T) this;
         }
 
+        public T initialSize(int initialPointStoreSize) {
```
jotok, line 692: In other places we use `size` to mean "the number of things in this container". If that's what this value mean, I would just call it "size".

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -710,6 +689,11 @@ public T capacity(int capacity) {
...
+            this.initialPointStoreSize = initialPointStoreSize;
+            return (T) this;
+        }
+
         public T currentStoreCapacity(int currentStoreCapacity) {
```
jotok, line 697: I wonder if we should change `capacity` to `maxCapacity` for the point store, and use `capacity` to mean "current capacity", i.e., number of points that can be stored without resizing. What do you think?

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: It's set to 2*sampleSize earlier. But can be solved cleaner - changing.

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeState.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: renaming to constructShingleInPlace. The goal was to avoid making copies.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: The mappers were using the get() --  but it may be a good change for the mappers to not write nulls. Changing. Will cause the addition of two of the mapper files.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: its initialSize ... if dynamicresizing is allowed, then we can start from a small size an not allocate the worst case storage.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: maxCapacity sounds like "active consideration" -- capacity is already signaling "max". But we can change if desired

jotok (approved):

Issue #223, opened 2021-06-03
[Rust v2.0] Fundamental Algorithms


cswiercz:
The Rust v2.0 should include implementations of the following algorithms. Each algorithm has listed in it the corresponding tracking issues and PRs.

- [x] scalar anomaly score (#192)
- [ ] vector anomaly score
- [ ] approximate anomaly scores (scalar and vector)
- [ ] density estimation
- [ ] imputation and forecasting
- [ ] nearest neighbors

PR #252, opened 2021-06-16
Anomaly attribution

cswiercz:
*Issue #, if available:*

Implements one of the algorithms in #223 . ("vector anomaly score")

*Description of changes:*

We create a new visitor for computing attribution scores and use
this visitor in a new RandomCutForest function. We also factor
out some common sub-functions used to compute anomaly and attribution
scores in a visitor utility function file.

### DIFF
--- a/Rust/src/random_cut_forest.rs
+++ b/Rust/src/random_cut_forest.rs
@@ -2,7 +2,7 @@ extern crate num_traits;
 use num_traits::{Float, Zero};
 
 use crate::SampledTree;
-use crate::visitor::AnomalyScoreVisitor;
+use crate::visitor::{AnomalyScoreVisitor, AttributionVisitor, DiVec};
 
 use std::marker::PhantomData;
 use std::iter::Sum;
@@ -126,7 +126,6 @@ impl<T> RandomCutForest<T>
     /// ```
     pub fn anomaly_score(&self, point: &Vec<T>) -> T {
         let mut anomaly_score: T = Zero::zero();
-
         if self.num_observations <= self.output_after {
             return anomaly_score;
         }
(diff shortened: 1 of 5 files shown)


cswiercz:
Questions for reviewers:

1. Is the name `anomaly_attribution()` okay?
2. Is there a sufficient amount of testing? -- Keep in mind that example code in any documentation is executed on every `cargo test`, in addition to the `#[cfg(test)]` submodules.

Rust/src/random_cut_forest.rs
```diff
@@ -333,27 +392,67 @@ mod tests {
...
+        let mut attribution_one_mean: f32 = 0.0;
+        let anomaly_attribution = forest.anomaly_attribution(&anomaly);
+        let anomaly_sum = anomaly_attribution.total_hi_lo_sum();
+        for i in 0..num_points {
+            let attribution = forest.anomaly_attribution(&points[i]);
+            let relative_error = (attribution.total_hi_lo_sum() - scores[i]).abs() / scores[i];
+
+            attribution_one_mean += attribution.hi_lo_sum(0);
+            assert!(relative_error < 0.01);
+            assert!(anomaly_attribution.hi_lo_sum(0) > attribution.hi_lo_sum(0));
```
cswiercz, line 431: For each point in the forest we test that the sum of it's attribution scores (`total_hi_lo_sum()`) is equal to its anomaly score. To do this "robustly", we check if the relative error between the attribution sum and the score is close to zero; specifically, within 1% relative error.

cswiercz: @cswiercz TODO - because we're traversing the same cuts we should test if the error is close to machine epsilon.

sudiptoguha:
No longer germane.

PR #254, opened 2021-06-17
make transformToShingledPoint private

wnbts:
RandomCutForest::transformToShingledPoint is a helper method for implementation. This PR moves it from public interface to protected implementation.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -608,7 +608,7 @@ public class RandomCutForest {
      * @return a shingled copy or a clean copy
      */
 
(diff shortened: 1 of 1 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -608,7 +608,7 @@ public int getThreadPoolSize() {
      * @return a shingled copy or a clean copy
      */
 
-    public double[] transformToShingledPoint(double[] point) {
+    private double[] transformToShingledPoint(double[] point) {
```
sudiptoguha, line 611: Would "protected" suffice?

wnbts: updated to protected

PR #262, opened 2021-07-16
small node store

sudiptoguha:
*Description of changes:* saves heap by using a smaller node store when sample size, dimensions are low and float_32 is used.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/state/store/SmallNodeStoreMapper.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/state/store/SmallNodeStoreMapper.java
@@ -15,6 +15,7 @@
 
 package com.amazon.randomcutforest.state.store;
 
+import static com.amazon.randomcutforest.CommonUtils.checkState;
 import static com.amazon.randomcutforest.tree.AbstractCompactRandomCutTree.NULL;
 
 import java.util.Arrays;
@@ -25,12 +26,12 @@ import lombok.Setter;
 import com.amazon.randomcutforest.CommonUtils;
 import com.amazon.randomcutforest.config.Precision;
 import com.amazon.randomcutforest.state.IStateMapper;
-import com.amazon.randomcutforest.store.NodeStore;
+import com.amazon.randomcutforest.store.SmallNodeStore;
 import com.amazon.randomcutforest.util.ArrayPacking;
 
 @Getter
 @Setter
-public class NodeStoreMapper implements IStateMapper<NodeStore, NodeStoreState> {
+public class SmallNodeStoreMapper implements IStateMapper<SmallNodeStore, NodeStoreState> {
 
     /**
      * if single precision, then stores the cut information as a float array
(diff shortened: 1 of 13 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/state/store/SmallNodeStoreMapper.java
```diff
@@ -0,0 +1,138 @@
...
+
+@Getter
+@Setter
+public class SmallNodeStoreMapper implements IStateMapper<SmallNodeStore, NodeStoreState> {
+
+    /**
+     * if single precision, then stores the cut information as a float array
+     * (converted to bytes)
+     */
+    private Precision precision = Precision.FLOAT_64;
```
kaituo, line 40: Could you explain why precision is hardcoded?

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeFloatMapper.java
```diff
@@ -40,9 +42,16 @@ public CompactRandomCutTreeFloat toModel(CompactRandomCutTreeState state, Compac
 
         INodeStore nodeStore;
 
-        NodeStoreMapper nodeStoreMapper = new NodeStoreMapper();
-        nodeStoreMapper.setPartialTreeStateEnabled(state.isPartialTreeState());
-        nodeStore = nodeStoreMapper.toModel(state.getNodeStoreState());
+        if (state.getMaxSize() < SmallNodeStore.MAX_SMALLNODESTORE_CAPACITY
```
kaituo, line 45: do we have functional tests covering both branches using SmallNodeStore and NodeStore?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -77,7 +80,12 @@ public AbstractCompactRandomCutTree(
         }
 
         if (builder.root == NULL) {
-            this.nodeStore = new NodeStore(maxSize - 1);
+            if (builder.float32Precision && maxSize < SmallNodeStore.MAX_SMALLNODESTORE_CAPACITY
```
kaituo, line 83: Can we extract a method to check if small node store is needed? Saw this logic used in a couple of places.

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/CompactRandomCutTreeFloatMapper.java
sudiptoguha: Good callout. Adding tests for SmallNodeStore. Let me add functional tests for both branches -- we have not tested with sample size > 16k at too many places, but there is no reason it cannot be used with large sample sizes.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Again, great callout. Changing.

Java/core/src/main/java/com/amazon/randomcutforest/state/store/SmallNodeStoreMapper.java
sudiptoguha: Um ... :) This is leftover from the default being FLOAT_64 and was explicitly overridden for FLOAT_32 (in NodeStore). So it should be changed in NodeStoreMapper. For SmallNodeStore this should be FLOAT_32. This is needed to store the cuts accurately because they are written out into the byte[]. Cleaning up.

sudiptoguha:
> Can you explain why small node store is bound to float 32 precision?

Sure. For float64 precision one would already be spending a lot of space in PointStore, cuts as float64. The savings of int -> short would have a small impact. That saving can be extracted with multiple files -- we can do that in the future if there is need. At the present moment, is there is a need for float64 x short ?

kaituo:
> > Can you explain why small node store is bound to float 32 precision?
> 
> Sure. For float64 precision one would already be spending a lot of space in PointStore, cuts as float64. The savings of int -> short would have a small impact. That saving can be extracted with multiple files -- we can do that in the future if there is need. At the present moment, is there is a need for float64 x short ?

got it.  Thanks for the explanation.

kaituo (approved):

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -90,6 +99,15 @@ public AbstractCompactRandomCutTree(
         }
     }
 
+    public static boolean canUseSmallNodeStore(boolean float32precision, int capacity, int dimensions) {
```
jotok, line 102: Can this take Precision instead of a boolean?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -63,12 +65,15 @@
     protected IBoxCache<Point> boxCache;
     protected Point[] pointSum;
     protected HashMap<Long, Integer>[] sequenceIndexes;
+    protected boolean float32Precision;
```
jotok, line 68: Let's move the Precision field here and out of the concrete classes.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -554,6 +573,11 @@ public T maxSize(int maxSize) {
             return (T) this;
         }
 
+        public T float32Precision(boolean precision) {
```
jotok, line 576: After we change this field to Precision, make this protected since it should be set by child classes and not by users.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done

jotok (approved):

PR #263, opened 2021-07-20
cache and convertor issues

sudiptoguha:
*Issue #, if available:* 261

*Description of changes:* introduces reInitialization of box caches and also introduces precision as a parameter in the conversion of models from 1.0 to 2.0.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -576,7 +576,8 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
     }
 
     @Override
-    protected void changeCacheFraction(double fraction) {
+    public void setBoundingBoxCacheFraction(double fraction) {
         boxCache.reInitialize(fraction);
+        super.setBoundingBoxCacheFraction(fraction);
     }
 }
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -577,7 +577,7 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
 
     @Override
     public void setBoundingBoxCacheFraction(double fraction) {
-        boxCache.reInitialize(fraction);
+        boxCache.setCacheFraction(fraction);
         super.setBoundingBoxCacheFraction(fraction);
     }
 }
--- a/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
@@ -575,4 +575,8 @@ public abstract class AbstractCompactRandomCutTree<Point> extends AbstractRandom
 
     }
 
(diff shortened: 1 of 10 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -575,4 +575,8 @@ public T maxSize(int maxSize) {
 
     }
 
+    @Override
+    protected void changeCacheFraction(double fraction) {
```
jotok, line 579: Do we really need the second method here? Instead of adding `changeCacheFraction` to `RandomCutTree` we could just override `setBoundingBoxCacheFraction`:
```
@Override
public void setBoundingBoxCacheFraction(double fraction) {
  super.setBoundingBoxCacheFraction(fraction);
  boxCache.reInitialize(fraction);
}
```

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
```diff
@@ -100,12 +100,16 @@ protected AbstractRandomCutTree(AbstractRandomCutTree.Builder<?> builder) {
         }
     }
 
+    protected void changeCacheFraction(double fraction) {
```
jotok, line 103: I think we don't need this method (see commend on AbstractCompactRandomCutTree).

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
```diff
@@ -575,4 +575,8 @@ public T maxSize(int maxSize) {
...
+        boxCache.reInitialize(fraction);
```
jotok, line 580: minor: I prefer "reinitialize" to "reInitialize".

Java/core/src/main/java/com/amazon/randomcutforest/tree/IBoxCache.java
```diff
@@ -59,4 +59,12 @@
      */
     void addToBox(int index, Point point);
 
+    /**
+     * changes the fraction of boxes cached dynamically
+     * 
+     * @param fraction new fraction of boxes to be cached
+     */
+
+    void reInitialize(double fraction);
```
jotok, line 68: I think we should call this `setCacheFraction`. Implementations can decide what to do with the fraction value (i.e., if reinitialization is required that's an implementation detail).

sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: redundant with issue below

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractRandomCutTree.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractCompactRandomCutTree.java
sudiptoguha: Done. Oops :)

kaituo (approved):

jotok (approved):

PR #267, opened 2021-08-12
merging models

sudiptoguha:
*Description of changes:* Merges a collection of 1.0 models (with same parameters) into a single forest with number of trees less or equal to the sum of the number of trees across the different models.

### DIFF
--- a/Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
+++ b/Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
@@ -106,36 +106,36 @@ public class V1JsonToV2StateConverter {
         }
 
         public void addSampler(V1SerializedRandomCutForest.Sampler sampler) {
-            V1SerializedRandomCutForest.WeightedSamples[] samples = sampler.getWeightedSamples();
-            int[] pointIndex = new int[samples.length];
-            float[] weight = new float[samples.length];
-            long[] sequenceIndex = new long[samples.length];
-
-            for (int i = 0; i < samples.length; i++) {
-                V1SerializedRandomCutForest.WeightedSamples sample = samples[i];
-                double[] point = sample.getPoint();
-                int index = pointStore.add(point, sample.getSequenceIndex());
-                pointIndex[i] = (Integer) globalTree.addPoint(index, 0L);
-                if (pointIndex[i] != index) {
-                    pointStore.incrementRefCount(pointIndex[i]);
-                    pointStore.decrementRefCount(index);
-                }
-                weight[i] = (float) sample.getWeight();
-                sequenceIndex[i] = sample.getSequenceIndex();
-            }
-
-            CompactSamplerState samplerState = new CompactSamplerState();
-            samplerState.setSize(samples.length);
-            samplerState.setCapacity(sampler.getSampleSize());
-            samplerState.setTimeDecay(sampler.getLambda());
-            samplerState.setPointIndex(pointIndex);
-            samplerState.setWeight(weight);
-            samplerState.setSequenceIndex(sequenceIndex);
-            samplerState.setSequenceIndexOfMostRecentTimeDecayUpdate(0L);
-            samplerState.setMaxSequenceIndex(sampler.getEntriesSeen());
-            samplerState.setInitialAcceptFraction(1.0);
-
             if (compactSamplerStates.size() < maxNumberOfTrees) {
+                V1SerializedRandomCutForest.WeightedSamples[] samples = sampler.getWeightedSamples();
+                int[] pointIndex = new int[samples.length];
+                float[] weight = new float[samples.length];
+                long[] sequenceIndex = new long[samples.length];
+
+                for (int i = 0; i < samples.length; i++) {
+                    V1SerializedRandomCutForest.WeightedSamples sample = samples[i];
+                    double[] point = sample.getPoint();
+                    int index = pointStore.add(point, sample.getSequenceIndex());
+                    pointIndex[i] = (Integer) globalTree.addPoint(index, 0L);
+                    if (pointIndex[i] != index) {
+                        pointStore.incrementRefCount(pointIndex[i]);
+                        pointStore.decrementRefCount(index);
+                    }
+                    weight[i] = (float) sample.getWeight();
+                    sequenceIndex[i] = sample.getSequenceIndex();
+                }
+
+                CompactSamplerState samplerState = new CompactSamplerState();
+                samplerState.setSize(samples.length);
+                samplerState.setCapacity(sampler.getSampleSize());
+                samplerState.setTimeDecay(sampler.getLambda());
+                samplerState.setPointIndex(pointIndex);
+                samplerState.setWeight(weight);
+                samplerState.setSequenceIndex(sequenceIndex);
+                samplerState.setSequenceIndexOfMostRecentTimeDecayUpdate(0L);
+                samplerState.setMaxSequenceIndex(sampler.getEntriesSeen());
+                samplerState.setInitialAcceptFraction(1.0);
+
                 compactSamplerStates.add(samplerState);
             }
         }
--- a/Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
+++ b/Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
@@ -15,13 +15,6 @@
 
 package com.amazon.randomcutforest.serialize.json.v1;
 
-import java.io.IOException;
-import java.io.Reader;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-
 import com.amazon.randomcutforest.config.Precision;
 import com.amazon.randomcutforest.state.ExecutionContext;
 import com.amazon.randomcutforest.state.RandomCutForestState;
@@ -37,6 +30,15 @@ import com.amazon.randomcutforest.tree.CompactRandomCutTreeFloat;
 import com.amazon.randomcutforest.tree.ITree;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import java.io.IOException;
+import java.io.Reader;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+
 public class V1JsonToV2StateConverter {
 
     private final ObjectMapper mapper = new ObjectMapper();
@@ -44,6 +46,14 @@ public class V1JsonToV2StateConverter {
     public RandomCutForestState convert(String json, Precision precision) throws IOException {
         V1SerializedRandomCutForest forest = mapper.readValue(json, V1SerializedRandomCutForest.class);
         return convert(forest, precision);
+    }
+
+    public RandomCutForestState convert(ArrayList<String> jsons, int numberOfTrees, Precision precision) throws IOException {
+        ArrayList<V1SerializedRandomCutForest> forests = new ArrayList<>();
+        for(int i=0; i<jsons.size();i++) {
+                forests.add(mapper.readValue(jsons.get(i), V1SerializedRandomCutForest.class));
+        }
+        return convert(forests, numberOfTrees, precision);
     }
 
     public RandomCutForestState convert(Reader reader, Precision precision) throws IOException {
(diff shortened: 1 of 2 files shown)


Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
```diff
@@ -46,6 +48,15 @@ public RandomCutForestState convert(String json, Precision precision) throws IOE
         return convert(forest, precision);
     }
 
+    public RandomCutForestState convert(ArrayList<String> jsons, int numberOfTrees, Precision precision)
```
wnbts, line 51: why not make a more general version of merge which takes multiple RandomCutForestState objects and outputs a merged RandomCutForestState? that way only one merge for the current version is needed

sudiptoguha: Do not follow ... that is what is happening. The older convert now calls this new convert with an array list of size 1.

wnbts: why not leave this converter do its job, V1SerializedRandomCutForest to RandomCutForestState and have a separate merge function that takes (List<RandomCutForestState) and outputs a merged RandomCutForestState so this merge function can work on any states wherever the state is directly from v2 or indirect converted from v1.

sudiptoguha: :) This library has never recommended that RCF be split. This package exists because the RCF state was split. Split models are inefficient and those considerations may/may not have gone in when the models were split. This RCF 2.0 package builds half a bridge -- the remainder of the bridge has to be built by the calling code.

wnbts: the suggestion is for improving code modularity.

with a bunch of json, the caller can do covert to get a list of states, then pass to merge(list of states). as you can see, existing convert is not unchanged. only a new merge is added to only address the merge need, in comparison to a monolithic convertAndMerge function.

wnbts: that said, i won't block this but to raise it for discussion

sudiptoguha: Ok. It is a good discussion to have. An iterated merge is a possibility -- but that means creating new state classes for the merge. That means new tests, new artifacts -- it can become a standard "wag the dog". This serialization package exists to get out of the business of json management and not to add more weight in this section.  This code is supposed to be run once -- I'm inclined to not overdevelop it.

wnbts: just a last word for this discussion, should the merge function just look like this without any new classes/types? And much of the code here can be moved there.
```
RandomCutForestState merge(List<RandomCutForestState>, int numTrees)
```

wnbts: unrelated minor. ArrayList should be List for general use.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
```diff
@@ -154,7 +135,69 @@ public void addSampler(V1SerializedRandomCutForest.Sampler sampler) {
...
+     */
+    public RandomCutForestState convert(List<V1SerializedRandomCutForest> serializedForests, int numberOfTrees,
+            Precision precision) {
+        checkArgument(serializedForests.size() > 0, "incorrect usage of convert");
+        checkArgument(numberOfTrees > 0, "incorrect parameter");
+        int sum = 0;
+        for (int i = 0; i < serializedForests.size(); i++) {
+            sum += serializedForests.get(i).getNumberOfTrees();
+        }
+        checkArgument(sum >= numberOfTrees, "incorrect parameters");
```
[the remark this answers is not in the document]
sudiptoguha: To answer your first question : Can you stop adding trees when it is larger than or equals numberOfTrees? This code will stop at 30 trees -- take a look at the tests :)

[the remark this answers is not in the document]
sudiptoguha: The coordinator should collect all the different model partitions, decide on how many trees it wants (has to be less than the sum of the trees across the different partitions) and call this. If the number of trees in the existing models do not meet the default threshold then the number of trees should be set to that lower number. Basically the coordinator should be the single repository of "the state of the merge". Otherwise there will be all sorts of issues with consistency.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
sudiptoguha: I see that the meta comment is not being acknowledged. This is not general use. Now the code is available, and some folks want to take responsibility for model decisions and create more state for running merges, and then fail because of distributed consistency issues, they are welcome to do so. Also, Precision should be a parameter, because as far as I understand, precision is being converted as well.

wnbts: I acknowledge this viewpoint. The difference is not a blocker. The discussion should give future readers enough information on this decision.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
```diff
@@ -46,6 +48,15 @@ public RandomCutForestState convert(String json, Precision precision) throws IOE
         return convert(forest, precision);
     }
 
+    public RandomCutForestState convert(ArrayList<String> jsons, int numberOfTrees, Precision precision)
+            throws IOException {
+        ArrayList<V1SerializedRandomCutForest> forests = new ArrayList<>();
```
wnbts, line 53: suggestion. the size of the array is known jsons.size().

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
```diff
@@ -57,50 +68,19 @@ public RandomCutForestState convert(URL url, Precision precision) throws IOExcep
...
-                state.getNumberOfTrees() * state.getSampleSize() + 1, precision);
-
-        Arrays.stream(serializedForest.getExecutor().getExecutor().getTreeUpdaters())
-                .map(V1SerializedRandomCutForest.TreeUpdater::getSampler).forEach(samplerConverter::addSampler);
-
-        state.setPointStoreState(samplerConverter.getPointStoreState(precision));
-        state.setCompactSamplerStates(samplerConverter.compactSamplerStates);
-
-        return state;
+        ArrayList<V1SerializedRandomCutForest> newList = new ArrayList<>();
```
wnbts, line 71: suggestion. for single-item list, there is Collections.singletonList or Arrays.asList.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
```diff
@@ -125,36 +106,98 @@ public PointStoreState getPointStoreState(Precision precision) {
...
-            samplerState.setMaxSequenceIndex(sampler.getEntriesSeen());
-            samplerState.setInitialAcceptFraction(1.0);
-
-            compactSamplerStates.add(samplerState);
+        SamplerConverter samplerConverter = new SamplerConverter(state.getDimensions(),
+                state.getNumberOfTrees() * state.getSampleSize() + 1, precision, numberOfTrees);
+
+        for (int i = 0; i < serializedForests.size(); i++) {
+            Arrays.stream(serializedForests.get(i).getExecutor().getExecutor().getTreeUpdaters())
+                    .map(V1SerializedRandomCutForest.TreeUpdater::getSampler).forEach(samplerConverter::addSampler);
```
wnbts, line 196: suggestion. another way to control the number of operations is to use Stream::limit(n). the way it doesn't iterate through the entire stream with many skips.

wnbts (approved):

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
sudiptoguha: If you suggest the syntax to write, I am happy to change.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
sudiptoguha: If you suggest the syntax I am happy to change.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
sudiptoguha: If you suggest the rewrite, I am happy to change

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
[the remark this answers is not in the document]
kaituo: The caller won’t know the actual number of trees in the json before invoking your method.  I want to avoid the illegalargument exception as exceptions is more expensive than an if else check.

Do you think this is better? I would return earlier if I know the conversion won’t succeed.  Change the number of trees to min(), but it brings complications that we need a threshold.  For example, if the total number of trees in the checkpoint is 3, we may not want to continue.

```
public Optional<RandomCutForestState> convert(ArrayList<String> jsons, int numberOfTrees, Precision precision)
            throws IOException {
        ArrayList<V1SerializedRandomCutForest> forests = new ArrayList<>();
    int sum = 0;
        for (int i = 0; i < jsons.size(); i++) {
       V1SerializedRandomCutForest forest = mapper.readValue(jsons.get(i), V1SerializedRandomCutForest.class)
            forests.add(forest);
      sum += forest.getNumberOfTrees();
        }
    if (sum < numberOfTrees) {
      return Optional.empty();
    }
        return Optional.ofNullable(convert(forests, numberOfTrees, precision));
    }
```

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
wnbts: yes, it's a suggestion for using the arraylist constructor with a known size. or there could be unnecessary overallocation/reallocation.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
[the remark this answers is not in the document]
kaituo: about the coordinator should be the single repository: we discussed offline.  Our conclusion so far is: in single stream AD, coordinating node will be the single threaded owner.  The model node will send different messages depending on the coordinating is old or new.  if the number of trees is different, it will send ExceptionA to the new coordinating node; but send scores to the old node.  The new coordinating node will get ExceptionA and try to merge or cold start. In HCAD, the worker node itself will do the conversion as each entity model is located in one node.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
wnbts: yes, it's a suggestion for constructing a list with a known number of elements.

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
wnbts: yes, it's a suggestion for control the number of adds. a side-benefit is no changes are required for SamplerConverter

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
wnbts: the suggested method is
```
public RandomCutForestState convert(List<String> jsons, int numberOfTrees, Precision precision)
```

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
wnbts: the suggested code is
```
ArrayList<V1SerializedRandomCutForest> forests = new ArrayList<>(jsons.size());
```

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
wnbts: the suggested code is
```
return convert(Arrays.asList(serializedForest), serializedForest.getNumberOfTrees(), precision);
```

sudiptoguha: Done. Java IDE suggested "return convert(Collections.singletonList(serializedForest), serializedForest.getNumberOfTrees(), precision);"

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
sudiptoguha: Done

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
wnbts: the suggested code for the for loop is something like this and the changes to SamplerCoverter are not needed:
```
            serializedForests.stream().flatMap(f -> Arrays.steam(f.getExecutor().getExecutor().getTreeUpdaters()))
                .limit(numberOfTrees).map(V1SerializedRandomCutForest.TreeUpdater::getSampler).forEach(samplerConverter::addSampler);
```

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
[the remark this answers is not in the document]
sudiptoguha: Done

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
sudiptoguha: Done. I think keeping the added check is beneficial in SamplerConverter. If that block of code is executed then the number of trees should not be changed -- because the book keeping in pointstore will be incorrect.

sudiptoguha:
> when sum is smaller, why this function returns empty and line 172 throws an exception? should the behaviors be the same?

No the behavior does not need to be the same. line 172 should never be executed. It is essential that the number of trees are never changed in the state class without knowing what one is doing. Line 172 should prevent adventures.

PR #268, opened 2021-08-12
Mergemodels

sudiptoguha:
*Description of changes:* This is the same as 267.

### DIFF
--- a/Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
+++ b/Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
@@ -106,36 +106,36 @@ public class V1JsonToV2StateConverter {
         }
 
         public void addSampler(V1SerializedRandomCutForest.Sampler sampler) {
-            V1SerializedRandomCutForest.WeightedSamples[] samples = sampler.getWeightedSamples();
-            int[] pointIndex = new int[samples.length];
-            float[] weight = new float[samples.length];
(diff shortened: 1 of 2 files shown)


Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
```diff
@@ -46,6 +50,22 @@ public RandomCutForestState convert(String json, Precision precision) throws IOE
         return convert(forest, precision);
     }
 
+    public Optional<RandomCutForestState> convert(ArrayList<String> jsons, int numberOfTrees, Precision precision)
```
wnbts, line 53: a javadoc would be useful such as on when the result will be empty.

wnbts (approved):

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v1/V1JsonToV2StateConverter.java
sudiptoguha: This code is run once and not intended for general use.

wnbts: ok. The data type for jsons should be `List<String>`

sudiptoguha: "not intended for general use" -> this is to spec of what the caller suggested.

kaituo (approved):

PR #274, opened 2021-08-27
cleanup + tests

sudiptoguha:
*Description of changes:* Adding tests and default values.


Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/threshold/ThresholdedRandomCutForest.java
```diff
@@ -203,20 +241,43 @@ public ThresholdedRandomCutForest(Builder<?> builder) {
...
+
         BasicThresholder basic = new BasicThresholder(builder.anomalyRate);
         if (forest.getDimensions() / forest.getShingleSize() == 1) {
-            basic.setLowerThreshold(1.1);
-            basic.setHorizon(0.75);
+            basic.setLowerThreshold(BasicThresholder.DEFAULT_LOWER_THRESHOLD_ONED);
+            basic.setHorizon(BasicThresholder.DEFAULT_HORIZON_ONED);
         }
         thresholder = basic;
     }
```
cswiercz, line 262: How do you set the horizon and lowerThreshold from the ThresholdedRandomCutForest builder?

sudiptoguha: Not feasible in this PR. Fixing. In general if the numeric parameters of BasicThresholder has to be changed from the larger class then the current interface is inappropriate because the interface should now have all the variable setters/getters in the same class. It also may make sense to be able to change these dynamically on the fly.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/threshold/ThresholdedRandomCutForest.java
```diff
@@ -361,14 +489,20 @@ public AnomalyDescriptor process(double[] inputPoint, long timestamp) {
             scaledInput = applyImpute(scaledInput, timestamp);
         }
 
-        // the following handles both external and internal shingling
-        double[] point = forest.transformToShingledPoint(scaledInput);
+        AnomalyDescriptor result = null;
+        if (getResult) {
```
jotok, line 493: If this block only needs to executed once after all the updates are in, maybe we can create a separate method for it instead of passing a boolean flag?

sudiptoguha: Execute once only after the initial updates are in. Refactored. The update steps are repeated though (without providing any output).

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/threshold/ThresholdedRandomCutForest.java
```diff
@@ -341,6 +417,36 @@ protected void updateState(double[] inputPoint, double[] input, long timeStamp)
...
+     */
+    void dischargeInitial() {
+        if (initialTimeStamps != null) {
+            for (int i = 0; i < initialTimeStamps.length - 1; i++) {
+                timeStampDeviation.update(initialTimeStamps[i + 1] - initialTimeStamps[i]);
+            }
+
+        }
+        if (initialValues != null && normalizeValues) {
+            if (differencing) {
```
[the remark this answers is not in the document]
sudiptoguha: applyDifferencing() is for the more continuous case. This is an execute once block where we only update the information needed for the normalization.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/threshold/BasicThresholder.java
```diff
@@ -195,4 +206,34 @@ public Deviation getSecondaryDeviation() {
         return secondaryDeviation;
     }
 
+    public void setZfactor(double factor) {
```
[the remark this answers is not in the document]
sudiptoguha: Done

jotok (approved):

PR #275, opened 2021-09-08
sampler changes for initialFraction

sudiptoguha:
*Description of changes:* currently samplers admit every point till they are full. However that may make all the samplers correlated and would indicate a change of behavior at sampleSize number of points. The behavior is now smoothed so that samplers could (potentially) diverge after initialFraction * sampleSize number of points and not store each of the first sampleSize points each. 

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
@@ -84,7 +84,7 @@ public abstract class AbstractStreamSampler<P> implements IStreamSampler<P> {
      * @param currentSize the current size of the sampler
      * @return the probability of admitting the next point
      */
-    protected double compute_fraction(int currentSize) {
+    protected double initialAcceptProbability(int currentSize) {
         if (currentSize < initialAcceptFraction * capacity) {
             return 1.0;
         } else if (initialAcceptFraction >= 1.0) {
--- a/Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
@@ -67,6 +67,34 @@ public abstract class AbstractStreamSampler<P> implements IStreamSampler<P> {
     protected final double initialAcceptFraction;
 
     /**
+     * a function that computes the probability of admittance of a new value when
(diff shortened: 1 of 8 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
```diff
@@ -66,6 +66,34 @@
...
+     * initialFraction * sampleSize number of points and then becomes a monotonic
+     * decreasing function.
+     *
+     * This function is supposed to be a parallel to the outputAfter() setting in
+     * the forest which controls how scores are emitted
+     * 
+     * @param currentSize the current size of the sampler
+     * @return the probability of admitting the next point
+     */
+    protected double compute_fraction(int currentSize) {
```
jotok, line 87: This needs a better name, but I'm having trouble coming up with a good one. `initialRegimeAcceptDecision` is a bit of mouthful. Maybe just `initialAcceptDecision`?

Java/core/src/test/java/com/amazon/randomcutforest/sampler/CompactSamplerTest.java
```diff
@@ -216,12 +216,19 @@ public void testAddPoint(Random random, CompactSampler sampler) {
...
             assertTrue(sampler.acceptPoint(i));
             assertNotNull(sampler.acceptPointState);
             sampler.addPoint(i);
         }
 
+        for (int i = 0; i < sampleSize * 10; i++) {
+            if (sampler.acceptPoint(i)) {
+                sampler.addPoint(i);
+            }
+            ;
```
jotok, line 229: Remove

Java/core/src/main/java/com/amazon/randomcutforest/sampler/AbstractStreamSampler.java
sudiptoguha: initialAcceptDecision sounds good.

sudiptoguha: initialAcceptProbability is perhaps more accurate

Java/core/src/test/java/com/amazon/randomcutforest/sampler/CompactSamplerTest.java
sudiptoguha: Done

jotok (approved):

PR #276, opened 2021-09-15
refactor

sudiptoguha:
*Description of changes:* refactors ThresholdedRCF to split the preprocessing and actual thresholding.


Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/preprocessor/BasicPreprocessor.java
```diff
@@ -0,0 +1,701 @@
...
+        return result;
+    }
+
+    /**
+     * applies transformations if desired
+     *
+     * @param inputPoint input point
+     * @return a differenced version of the input
+     */
+    protected double[] transformValues(double[] inputPoint, double[] factors) {
```
jotok, line 469: This method makes it look like we should have a  preprocessor base class or interface with 4 implementations. BasicPreprocessor does not seem very basic!

sudiptoguha: Fair point :) "There is no telling where the road takes us to ...". Fixing names.

jotok (approved):

PR #279, opened 2021-09-30
pointstore fix

sudiptoguha:
*Description of changes:* While using internal shingling, if the size of the store is small compared to the number of updates, then a new unshingled point can be stored as an incorrect new shingled point. 

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
@@ -178,18 +178,6 @@ public abstract class PointStore<Store, Point> implements IPointStore<Point> {
         return location;
     }
 
-    void verifyAndMakeSpace(int amount) {
-        if (startOfFreeSegment > currentStoreCapacity * dimensions - amount) {
-            // try compaction and then resizing
-            compact();
-            if (startOfFreeSegment > currentStoreCapacity * dimensions - amount) {
-                checkState(dynamicResizingEnabled, " out of store, enable dynamic resizing ");
-                resizeStore();
-                checkState(startOfFreeSegment + amount <= currentStoreCapacity * dimensions, "out of space");
-            }
-        }
-    }
-
     /**
      * Add a point to the point store and return the index of the stored point.
      *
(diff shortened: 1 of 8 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -221,7 +209,18 @@ public int add(double[] point, long sequenceNum) {
...
-            verifyAndMakeSpace(amountToWrite);
+            if (startOfFreeSegment > currentStoreCapacity * dimensions - amountToWrite) {
+                // try compaction and then resizing
+                compact();
+                // the compaction can change the array contents
+                amountToWrite = checkShingleAlignment(startOfFreeSegment, tempPoint) ? baseDimension : dimensions;
+                if (startOfFreeSegment > currentStoreCapacity * dimensions - amountToWrite) {
+                    checkState(dynamicResizingEnabled, " out of store, enable dynamic resizing ");
+                    resizeStore();
+                    checkState(startOfFreeSegment + amountToWrite <= currentStoreCapacity * dimensions, "out of space");
```
jotok, line 220: When would this happen? Is there a max size that the point store can grow to?

sudiptoguha: Yes. The maximum size is "capacity * dimensions" (and twice that if shingle rotations are enabled) ; if dynamicResizing is turned on then currentStoreCapacity starts small and can reach capacity. However that depends on the timeDecay factor and it is possible that the pointStore never reaches full capacity -- for example if the time decay is large, we are essentially ensuring a "sliding window" (last few data points) behavior and that would determine the pointStore (and maybe a few rare straggler points).

kaituo: Could you explain why shingle rotation need twice the max capacity? I thought one is a cyclic buffer and another is a sliding window. The buffer and window's size are the same, right?

Also, if shingle rotation is enabled, should you change the check condition to startOfFreeSegment + amountToWrite <= currentStoreCapacity * dimensions * 2 ?

Java/core/src/test/java/com/amazon/randomcutforest/store/PointStoreDoubleTest.java
```diff
@@ -257,4 +257,27 @@ public void checkRotationAndCompact() {
...
     }
+
+    @Test
+    void CompactionTest() {
+        int shinglesize = 2;
+        PointStoreDouble store = new PointStoreDouble.Builder().capacity(6).dimensions(shinglesize)
+                .shingleSize(shinglesize).indexCapacity(6).directLocationEnabled(false).internalShinglingEnabled(true)
+                .build();
+
+        store.add(new double[] { 0 }, 0L);
```
kaituo, line 268: the 2nd parameter sequenceNum is unused, should we remove it in the add method?

PR #283, opened 2021-12-20
initiating RCF3.0

sudiptoguha:
*Description of changes:*  This PR initiates RCF 3.0.

RCF 2.0 introduced multiple primitives, such as bounding box caching, tree representations, columnar compression to provide a space-time tradeoff. However those primitives were not optimized in any manner. RCF 3.0 reorganizes those primitives to provide a more optimal (and hopefully cleaner) tradeoff. 

None of the visitor classes were changed and no inference is expected to vary in this PR. There is unlikely to be change in performance of RCF2.0 and RCF3.0 for large model sizes (bounding box cache 1). The RCF 2.0 files were not changed (except minor bug fixes and plumbing to allow direct comparison).

RCF3.0 will focus exclusively on single precision. Double precision calculations over randomized data structures is unlikely to be helpful in the context of RCF. Moreover the newer logic in RCF 3.0 literally "thinks outside the bounding box". While the bounding box description is easy to follow, relying on such descriptions inside an algorithm may be inimical to algorithmic efficiency. We note that efficiency is a key distinguishing feature of RCFs vis-a-vis random forests.

This PR also overhauls the Rust version to the same exact algorithm as the Java version. RCF 2.0 was designed with Rust and memory safely in mind. The parallel implementation underscores that point. We see the Rust version as a mechanism of additional verification of the algorithmic ideas and expect the different language implementations to remain in sync. None of these versions are superior to the other in performance, and we expect to continue to do everything to keep the Java version as performant as possible.


Java/core/src/main/java/com/amazon/randomcutforest/executor/ITraversable.java
```diff
@@ -79,4 +81,11 @@
      *         traversal query, otherwise false.
      */
     boolean isOutputReady();
+
+    default double dynamicScore(double[] point, int ignoreMass, BiFunction<Double, Double, Double> scoreSeen,
```
jotok, line 85: If we want to special-case the anomaly scoring path, my suggestion would be to create a new anomaly score visitor (possibly a sub-interface of Visitor, but not necessarily) that defines `scoreSeen` etc. as methods. Then add a method to ITraversable that takes an anomaly score visitor.

I think there's potentially an even more generic solution here, where we define "lightweight" and "heavyweight" visitors, where the lightweight visitor has access to exactly what the anomaly scoring algorithm needs to work, and the heavyweight visitors get to act on the full node view context.

The name `dynamicScore` is not very descriptive in the context of a Traversable.

Java/core/src/main/java/com/amazon/randomcutforest/store/IntervalManager.java
```diff
@@ -0,0 +1,126 @@
...
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+import static com.amazon.randomcutforest.CommonUtils.checkState;
+
+import java.util.Arrays;
+
+/**
+ * This class defines common functionality for Store classes, including
+ * maintaining the stack of free pointers.
+ */
+public class IntervalManager {
```
jotok, line 27: I assume you didn't want to create confusion with the existing Index Manager, but the methods in this class are `takeIndex` and `releaseIndex`, it's unclear what this has to do with intervals.

Java/core/src/main/java/com/amazon/randomcutforest/store/IntervalManager.java
```diff
@@ -0,0 +1,126 @@
...
+        freeIndexesEnd = new int[1];
+        freeIndexesStart = new int[1];
+        lastInUse = 1;
+        this.capacity = capacity;
+        freeIndexesStart[0] = 0;
+        freeIndexesEnd[0] = capacity - 1;
+    }
+
+    public void extendCapacity(int newCapacity) {
+        checkArgument(newCapacity > capacity, " incorrect call, we can only increase capacity");
```
jotok, line 45: Is this check correct? Looks like it should be `<`.

Java/core/src/main/java/com/amazon/randomcutforest/store/RCF3PointStoreLarge.java
```diff
@@ -0,0 +1,58 @@
...
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+
+import java.util.Arrays;
+
+public class RCF3PointStoreLarge extends RCF3PointStore {
+
+    static int INFEASIBLE_LOCN = (int) -1;
+    protected int[] locationList;
+
+    void setInfeasiblePointstoreLocationIndex(int index) {
```
jotok, line 27: I think we can make these methods protected instead of package private. I can't think of a reason to disallow sub-classing.

Java/core/src/main/java/com/amazon/randomcutforest/executor/ITraversable.java
sudiptoguha: Agree with all three points -- did not edit visitors in this PR (it's already large). At the current moment this provides a mechanism for us to measure overhead of the visitor. The name comes from the DynamicScoringRandomCutForest (since the 1.0) which we should absorb (eventually, and not in this PR). That would enable a simple path to using "displacement score" discussed in the original paper. But yes, the name is up for discussion (both from the forest context as well as from Traversable).

Java/core/src/main/java/com/amazon/randomcutforest/store/IntervalManager.java
sudiptoguha: Yes (re not wanting to change IndexManager). This class should replace the IndexManager. This class uses intervals to manage the free indices (which seem to be consumed and released in bursts of intervals in PointStore). The current IndexManager has the implicit way of compressing the stack etc. This new class simplifies all of that (keeping the intended implicit compression). It also eliminates the BitSet etc., and is easier to change it's size dynamically.

Java/core/src/main/java/com/amazon/randomcutforest/store/IntervalManager.java
sudiptoguha: for the index managers, the capacity is not the maximum capacity but a current capacity which we kept increasing in RCF2.0. We may be able to eliminate that need with 3.0 altogether (because of the intervals).

Java/core/src/main/java/com/amazon/randomcutforest/store/RCF3PointStoreLarge.java
sudiptoguha: Agree.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
```diff
@@ -0,0 +1,527 @@
...
+ * Nodes.
+ *
+ * Note that a NodeStore does not store instances of the
+ * {@link com.amazon.randomcutforest.tree.Node} class.
+ */
+public abstract class AbstractNodeStore {
+
+    public static double SWITCH_FRACTION = 0.499;
+
+    protected final int capacity;
```
jotok, line 52: I think this is either the number of internal nodes or the number of leaf nodes that we can store now. Should we change the name?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
```diff
@@ -0,0 +1,527 @@
...
+                rangeSum += boundingBoxData[mid + i] - boundingBoxData[base + i];
+            }
+            boolean answer = (rangeSumData[idx] == rangeSum);
+            rangeSumData[idx] = rangeSum;
+            return answer;
+        }
+        return false;
+    }
+
+    public BoundingBoxFloat getBox(int index, IPointStoreView<float[]> pointStoreView) {
```
jotok, line 155: It seems strange to me to make the point store view a method parameter and not an instance field. By having it as a method argument, that suggests that you should be able to pass any point store view in and get some kind of answer.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
```diff
@@ -0,0 +1,527 @@
...
+            sibling = getRightIndex(parent);
+        }
+        return sibling;
+    }
+
+    public abstract void spliceEdge(int parent, int node, int newNode);
+
+    public abstract void replaceParentBySibling(int grandParent, int parent, int node);
+
+    public double dynamicScore2(int root, int ignoreMass, double[] point, IPointStoreView<float[]> pointStoreView,
```
jotok, line 338: Do we need all of these scoring methods?

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
```diff
@@ -0,0 +1,527 @@
...
+        return cutValue[index - 1];
+    }
+
+    public double getNodeCacheFraction() {
+        return nodeCacheFraction;
+    }
+
+    protected <R> void traversePathToLeafAndVisitNodes(double[] point, Visitor<R> visitor, int root,
+            IPointStoreView<float[]> pointStoreView, Function<double[], double[]> projectToTree) {
+        AbstractNodeView currentNodeView = new AbstractNodeView(this, pointStoreView, root);
```
jotok, line 463: Is AbstractNodeView an abstract class? If it is, I feel like we shouldn't be able to construct one like this.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeView.java
```diff
@@ -0,0 +1,94 @@
...
+package com.amazon.randomcutforest.tree;
+
+import static com.amazon.randomcutforest.CommonUtils.toDoubleArray;
+
+import java.util.Collections;
+import java.util.Set;
+
+import com.amazon.randomcutforest.store.IPointStoreView;
+
+public class AbstractNodeView implements INodeView {
```
jotok, line 25: We should change the name of the class, since it implies that it's `abstract`. Maybe NodeStoreBackedView or NodeStoreView or something like that.

Java/core/src/main/java/com/amazon/randomcutforest/store/RCF3PointStore.java
```diff
@@ -0,0 +1,819 @@
...
+     */
+    public int incrementRefCount(int index) {
+        // indexManager.checkValidIndex(index);
+        Integer value = refCountMap.remove(index);
+        if (value == null) {
+            if ((refCount[index] & 0xff) == 255) {
+                refCountMap.put(index, 1);
+                return 256;
+            } else {
+                int newVal = (byte) ((refCount[index] & 0xff) + 1);
```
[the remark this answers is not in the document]
sudiptoguha: yes

Java/core/src/main/java/com/amazon/randomcutforest/store/RCF3PointStore.java
```diff
@@ -0,0 +1,819 @@
...
+     * Decrement the reference count for the given index.
+     *
+     * @param index The index value.
+     * @throws IllegalArgumentException if the index value is not valid.
+     * @throws IllegalArgumentException if the current reference count for this
+     *                                  index is non positive.
+     */
+    @Override
+    public int decrementRefCount(int index) {
+        // indexManager.checkValidIndex(index);
```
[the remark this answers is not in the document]
sudiptoguha: removed or instantiated to the current notion of valid index, next commit

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
sudiptoguha: yes.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
sudiptoguha: I think the 2.0 version called for a more lightweight nodeview -- let me revisit this, it's possible that we can simplify these.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
sudiptoguha: No, the goal was to measure (sort of) had optimized computation vis-a-vis the traversals. It seems that the traversals can be optimized to be pretty much the same.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
sudiptoguha: No ... bad naming. Will change.

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeView.java
sudiptoguha: Yes.

Java/core/src/main/java/com/amazon/randomcutforest/store/IntervalManager.java
```diff
@@ -0,0 +1,128 @@
...
+            freeIndexesStart = Arrays.copyOf(freeIndexesStart, lastInUse + 1);
+            freeIndexesEnd = Arrays.copyOf(freeIndexesEnd, lastInUse + 1);
+        }
+
+        freeIndexesStart[lastInUse] = index;
+        freeIndexesEnd[lastInUse] = index;
+        lastInUse += 1;
+    }
+
+    public int getSize() {
```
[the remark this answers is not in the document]
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/executor/ITraversable.java
sudiptoguha: yes - done (removed from the interface - but the tree classes still needs to have them while we compare different paths)

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -178,6 +178,27 @@ int takeIndex() {
         return location;
     }
 
+    int getAmountToWrite(double[] tempPoint) {
+        if (checkShingleAlignment(startOfFreeSegment, tempPoint)) {
+            if (!rotationEnabled
+                    || startOfFreeSegment % dimensions == (nextSequenceIndex - 1) * baseDimension % dimensions) {
+                return baseDimension;
+            }
+        } else {
```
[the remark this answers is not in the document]
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/AbstractNodeStore.java
sudiptoguha: Removed

jotok (approved):

Java/core/src/main/java/com/amazon/randomcutforest/store/RCF3PointStoreLarge.java
```diff
@@ -0,0 +1,58 @@
...
+
+package com.amazon.randomcutforest.store;
+
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+
+import java.util.Arrays;
+
+public class RCF3PointStoreLarge extends RCF3PointStore {
+
+    static int INFEASIBLE_LOCN = (int) -1;
```
ylwu-amzn, line 24: @sudiptoguha I think this bug https://github.com/aws/random-cut-forest-by-aws/pull/322 is caused by this line, right?

sudiptoguha: It's not a bug. The representation of the pointstore changed :) Look at lines 28 and 31-37; the unused locations have value -1. However for the 2.0.1 models the unused valued could have been anything (the check was via refList). I think 1.3 picked up something earlier than 283.

ylwu-amzn: Got it. Thanks, agree this not a bug, just some breaking change compared with RCF jar used by AD 1.3 (PR #282).  Should be ok for `ml-commons` 1.3.1, as it's using PR #309.

PR #286, opened 2022-01-12
Collapse PointStoreEdit and PointStoreView into a single PointStoreTrait

jotok:
- Create a single PointStore trait with the methods from PointStoreEdit
  and PointStoreView combined. We can use mutable/non-mutable references
  to distinguish between read-only and read-write use cases.
- Create a new Location trait to simplify the type constraints used for
  store locations.
- Create a new types.rs file to define traits used in multiple other files.
- Create a rustfmt.toml file and configure rustfmt to organize imports.

### DIFF
--- a/Rust/src/pointstore.rs
+++ b/Rust/src/pointstore.rs
@@ -1,14 +1,28 @@
-use crate::rcf::Max;
-use std::fmt::Debug;
 extern crate num;
-use crate::intervalstoremanager::IntervalStoreManager;
+use std::{collections::HashMap, convert::TryFrom, fmt::Debug};
 
-use std::collections::HashMap;
+use crate::{intervalstoremanager::IntervalStoreManager, types::Location};
 
-use std::convert::TryFrom;
+pub trait PointStore {
+    fn get_shingled_point(&self, point: &[f32]) -> Vec<f32>;
+    fn get_size(&self) -> usize;
+    fn get_missing_values(&self, values: &[usize]) -> Vec<usize>;
+    fn get_copy(&self, index: usize) -> Vec<f32>;
+    fn is_equal(&self, point: &[f32], index: usize) -> bool;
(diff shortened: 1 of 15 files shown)


Rust/src/pointstore.rs
```diff
@@ -1,14 +1,28 @@
...
+    fn add(&mut self, point: &[f32]) -> usize;
+    fn inc(&mut self, index: usize);
+    fn dec(&mut self, index: usize);
+    fn adjust_count(&mut self, result: &[(usize, usize)]);
+    fn compact(&mut self);
+}
 
 #[repr(C)]
-pub struct PointStore<L> {
+pub struct VectorPointStore<L>
```
sudiptoguha, line 22: This is correct -- we are using a vector as a pointstore. But there may be a confusion if some one expects that we are storing the points as individual vectors... (which we are not, with shinglesize>1)

Rust/src/randomcuttree.rs
```diff
@@ -116,7 +117,7 @@ where
 
                 let mut parent = saved_parent;
                 let mut saved_cut = Cut::new(usize::MAX, 0.0);
-                /** the loop has the execute once */
+                /* the loop has the execute once */
```
sudiptoguha, line 120: Oops. Errors from previous commit -- "this loop will execute at least once" ?

Rust/src/types.rs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+impl Max for usize {
+    const MAX: usize = usize::MAX;
+}
+
+/// The Location trait is used as a shorthand for the various traits needed by store (e.g., point
+/// store, node store) locations. These are the values vended by stores to reference a stored
+/// value.
+pub trait Location: Copy + Max + std::cmp::PartialEq + Into<usize> + TryFrom<usize> + Sync {}
```
sudiptoguha, line 21: Nice!

sudiptoguha (approved):

PR #287, opened 2022-01-14
serde for nodestore

sudiptoguha:
*Description of changes:* Mappers etc for the new trees (along with conversion from 2.0/2.1 versions for double precision). The next step would be deletion and deprecation of double precision.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/CommonUtils.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/CommonUtils.java
@@ -199,20 +199,21 @@ public class CommonUtils {
 
     };
 
-    public static double[] toDoubleArray(float[] point) {
-        checkNotNull(point, "array must not be null");
-        double[] result = new double[point.length];
-        for (int i = 0; i < point.length; i++) {
-            result[i] = point[i];
+    public static double[] toDoubleArray(float[] array) {
+        checkNotNull(array, "array must not be null");
+        double[] result = new double[array.length];
+        for (int i = 0; i < array.length; i++) {
+            result[i] = array[i];
         }
         return result;
     }
 
-    public static float[] toFloatArray(double[] point) {
-        checkNotNull(point, "array must not be null");
-        float[] result = new float[point.length];
-        for (int i = 0; i < point.length; i++) {
-            result[i] = (float) point[i];
+    public static float[] toFloatArray(double[] array) {
+        checkNotNull(array, "array must not be null");
+        float[] result = new float[array.length];
+        for (int i = 0; i < array.length; i++) {
+            result[i] = (array[i] == 0) ? 0 : (float) array[i];
+            // eliminating -0.0 issues
         }
         return result;
     }
--- a/Java/core/src/main/java/com/amazon/randomcutforest/CommonUtils.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/CommonUtils.java
@@ -200,7 +200,7 @@ public class CommonUtils {
     };
 
     public static double[] toDoubleArray(float[] point) {
-        checkNotNull(point, "point must not be null");
+        checkNotNull(point, "array must not be null");
         double[] result = new double[point.length];
         for (int i = 0; i < point.length; i++) {
             result[i] = point[i];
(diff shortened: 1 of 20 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/CommonUtils.java
```diff
@@ -200,7 +200,7 @@ public static double defaultScalarNormalizerFunction(double scalarValue, double
     };
 
     public static double[] toDoubleArray(float[] point) {
-        checkNotNull(point, "point must not be null");
+        checkNotNull(point, "array must not be null");
```
jotok, line 203: The error message was written to match the formal parameter name, so I would either change this back or change the formal parameter to `array`.

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
```diff
@@ -219,6 +219,12 @@ public void reheap(boolean validate) {
         }
     }
 
+    public void addPoint(Integer pointIndex, float weight, long sequenceIndex) {
```
jotok, line 222: Why did this method get added now? What changed?

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -823,6 +823,10 @@ public boolean pointEquals(int index, float[] point) {
         int address = getLocation(index);
 
         if (!rotationEnabled) {
+
+              if (store.length < (address + dimensions)) {
+                  System.out.println("AAA " + store.length + " " + address + " " + (address + dimensions));
```
jotok, line 828: Remove

Java/core/src/main/java/com/amazon/randomcutforest/tree/NewRandomCutTree.java
```diff
@@ -58,38 +59,42 @@
...
 
     private Random testRandom;
+    protected boolean storeSequenceIndexesEnabled;
+    protected boolean centerOfMassEnabled;
     private long randomSeed;
     protected int root;
     protected IPointStoreView<float[]> pointStoreView;
-    protected int treeMass;
-    protected int capacity;
+    // protected int treeMass;
```
jotok, line 67: Remove?

Java/core/src/main/java/com/amazon/randomcutforest/sampler/CompactSampler.java
sudiptoguha: This is to enable double[] -> float[] conversion for the trees. Up until now we were bringing up the exact same samplers in ser-de. The guarantee we maintained is that for duplicates in the same tree, the corresponding sampler would have the same pointIndex value. In the conversion, we first round off the pointstore -- but then we may be creating duplicates. So we have to insert the points into a fresh tree and update the pointIndex. One option is to add the point to the samplerPlusTree and update both sampler and tree -- we could use the existing code. Other option is to add them to tree but not affect the sampler (other than updating the pointIndex value) -- which is happening here. Advantage of option 2 is that the sampler effectively remains in the same state; the scores seem a bit smoother in the conversion. Looking far down the road, this should also enable adding "sticky"  points to a sampler -- but that would require a resize() for grow() option to dynamically change (increase is the difficult part) the sampler size.

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/NewRandomCutTree.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/CommonUtils.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreFloatMapper.java
```diff
@@ -50,7 +53,27 @@ public PointStore toModel(PointStoreState state, long seed) {
         int[] refCount = ArrayPacking.unpackInts(state.getRefCount(), indexCapacity, state.isCompressed());
         int[] locationList = new int[indexCapacity];
         int[] tempList = ArrayPacking.unpackInts(state.getLocationList(), state.isCompressed());
-        System.arraycopy(tempList, 0, locationList, 0, tempList.length);
+        if (!state.getVersion().equals(Version.V3_0)) {
```
jotok, line 56: If we start having to create more branching logic like this, I suggest we start creating separate methods for clarity:
```
public PointStore toModel(PointStoreState state, long seed) {
  if (state.getVersion.compareTo(Version.V3_0) < 0) {
    return toModelV2(state, seed);
  } else {
    return toModelV3(state, seed);
  }
}

private PointStore toModelV2(PointStoreState state, long seed) { ... }
private PointStore toModelV3(PointStoreState state, long seed) { ... }
```
}
```

jotok (approved):

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreFloatMapper.java
sudiptoguha: Agree wholeheartedly. Lets even start now - while we remember changes. Will push changes in the next few PRs (maybe keep the next one focused on just delete).

PR #288, opened 2022-01-14
Simplifying type declarations

jotok:
Use the Location trait to simplify the struct and method definitions in node store, random cut tree, etc. Renamed VectorPointStore and NodeStoreView for clarity.

- Use the Location trait consistently where applicable in all files
- Rename VectorPointStore to VectorizedPointStore
- Rename NodeStoreView to NodeStore

### DIFF
--- a/Rust/src/nodestore.rs
+++ b/Rust/src/nodestore.rs
@@ -1,8 +1,11 @@
 use std::{collections::HashMap, fmt::Debug, mem};
 
 use crate::{
-    boundingbox::BoundingBox, cut::Cut, intervalstoremanager::IntervalStoreManager,
-    pointstore::PointStore, types::Max,
+    boundingbox::BoundingBox,
+    cut::Cut,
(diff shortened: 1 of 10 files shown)


Rust/src/nodestore.rs
```diff
@@ -21,7 +24,15 @@ use crate::{
 */
 
 #[repr(C)]
-pub struct NewNodeStore<C, P, N> {
+pub struct VectorNodeStore<C, P, N>
```
sudiptoguha, line 27: VectorizedNodesStore? One reason for not using structs has been that we use different sizes for the left(right), cut and mass (structs may force a loss because if packing/alignment).

sudiptoguha (approved):
Looks good

Rust/src/nodestore.rs
jotok: Yes, I changed it for `VectorizedPointStore` but forgot this one. Will fix in the next PR.

PR #289, opened 2022-01-19
removing v1.0 nodes

sudiptoguha:
Begins the removal of RCF1.0 classes


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -1424,4 +1390,183 @@ public Random getRandom() {
...
+     * @param ignoreLeafMassThreshold said threshold
+     * @param seen                    the function that applies if input is equal to
+     *                                a previously seen sample in a leaf
+     * @param unseen                  if the input does not have a match in the
+     *                                leaves
+     * @param damp                    damping function based on the duplicity of the
+     *                                previously seen samples
+     * @return anomaly score
+     */
+    public double getDynamicScore(double[] point, int ignoreLeafMassThreshold, BiFunction<Double, Double, Double> seen,
```
jotok, line 1407: I'm still not a fan of having these methods in the API. I would prefer a method like:
```
public double getAverageScalarScore(double[] point, VisitorFactory<Double> visitorFactory)
```
or
```
public double getAverageScalarScore(double[] point, Supplier<Visitor<Double>> supplier)
```
This would provide a simple generic interface for any kind of visitor that produces a scalar score, where you just want to average the scores over the trees.

jotok: Created issue for tracking
#290

jotok (approved):

PR #294, opened 2022-02-09
upgrading impute

sudiptoguha:
*Description of changes:* currently the imputation works differently for one versus multiple missing values. This is remedied in these changes. Also adds an summary return type.


Java/core/src/main/java/com/amazon/randomcutforest/imputation/ConditionalSampleSummarizer.java
```diff
@@ -0,0 +1,416 @@
+/*
```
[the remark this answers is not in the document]
sudiptoguha: Not sure if the copyright header should be changed since there has been no other change. I think the tweet https://twitter.com/_msw_/status/1478820243506343936 summarizes the discussion.

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/ConditionalSampleSummary.java
```diff
@@ -0,0 +1,82 @@
...
+
+package com.amazon.randomcutforest.returntypes;
+
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+import static com.amazon.randomcutforest.CommonUtils.toFloatArray;
+
+public class ConditionalSampleSummary {
+
+    /**
+     * a collection of typical points (in the shingled, input space) seen from the
```
jotok, line 24: "typical" is not very descriptive. Given that this is a container class, Can we just call this class `Sample`, and we can call this field `SampledPoints`. The fact that the imputation method fills this container with "typical" points is a property of the imputation algorithm, not of the container.

Java/core/src/main/java/com/amazon/randomcutforest/tree/INodeView.java
```diff
@@ -41,4 +41,6 @@
 
     double probailityOfSeparation(float[] point);
 
+    int getLeafPointIndex();
```
jotok, line 44: Documentation: for a leaf node, return the index in the point store for the leaf point. If this method is invoked on a non-leaf node then it should throw an IllegalStateException.

Java/core/src/main/java/com/amazon/randomcutforest/tree/NodeView.java
```diff
@@ -76,6 +76,12 @@ public double probailityOfSeparation(float[] point) {
         return nodeStore.probabilityOfCut(currentNodeOffset, point, pointStoreView, currentBox);
     }
 
+    @Override
+    public int getLeafPointIndex() {
+        checkArgument(isLeaf(), "incorrect call");
```
jotok, line 81: use `checkState` instead of `checkArgument` and give a better error message like "cannot invoke 'getLeafPointIndex' from a non-leaf node".

Java/core/src/main/java/com/amazon/randomcutforest/anomalydetection/AbstractAttributionVisitor.java
```diff
@@ -15,14 +15,14 @@
 
 package com.amazon.randomcutforest.anomalydetection;
 
+import java.util.Arrays;
```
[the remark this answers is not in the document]
sudiptoguha: Will do.

Java/core/src/main/java/com/amazon/randomcutforest/Visitor.java
```diff
@@ -56,12 +56,14 @@ default void acceptLeaf(INodeView leafNode, final int depthOfNode) {
...
-     * This method can be overriden to optimize visitors that do not need to visit every node on the root to leaf path
-     * before returning a value.
+     * This method short-circuits the evaluation of the Visitor at nodes on the
+     * traversal path. By default, the accept (or acceptLeaf) method will be invoked
+     * for each Node in the traversal path. But the NodeView has to prepare
+     * information to support that visitor invocation. Before invocation, the value
+     * of isConverged will be checked. If it is true, some of that preparation can
+     * be skipped -- because the visitor would not be upodated. This method can be
+     * overriden to optimize visitors that do not need to visit every node on the
+     * root to leaf path before returning a value.
```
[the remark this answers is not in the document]
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -15,24 +15,25 @@
...
  * are first imputed with the corresponding values in the leaf node in the
  * traversal path. Then, when this MultiVisitor is merged with another
  * MultiVisitor, we keep the imputed values with a lower rank, where the rank
  * value is the anomaly score for the imputed point.
  */
-public class ImputeVisitor implements MultiVisitor<double[]> {
+public class ImputeVisitor implements MultiVisitor<ConditionalTreeSample> {
 
     // default large values for initialization; consider -ve log( 0 )
     public static double DEFAULT_INIT_VALUE = Double.MAX_VALUE / 4;
```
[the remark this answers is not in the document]
sudiptoguha: :) I cannot recall. Removed the / 4. But re the question: if the distances are on this scale then changes are that there will be overflow (for example if we wish to compute average distance from 50 trees) ...

Java/core/src/main/java/com/amazon/randomcutforest/tree/INodeView.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/tree/NodeView.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/tree/INodeView.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/ConditionalSampleSummary.java
sudiptoguha: Changed the comment -- it is supposed to be "typical" in the information theory sense, basically ion the samples correspond to union of two well-separated gaussians, we should see the two centers. That corresponds to the summarization aspect.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ConditionalSampleSummarizer.java
```diff
@@ -0,0 +1,416 @@
...
+import static java.lang.Math.sqrt;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import com.amazon.randomcutforest.returntypes.ConditionalSampleSummary;
+import com.amazon.randomcutforest.returntypes.ConditionalTreeSample;
+
+public class ConditionalSampleSummarizer {
```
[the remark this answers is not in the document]
sudiptoguha: Agree in principle. But the park services framework needs revisions (even without this class). Moreover imputation is more increasingly a core functional use -- in terms of explaining anomalies etc., so something similar to this function should belong in core. Probably a broader refactor of core + park services will be a good solution in the future.

jotok (approved):

PR #320, opened 2022-04-11
Rust : refactor + attribution

sudiptoguha:
*Description of changes:* Adds attribution to the rust version. Further refactors the visitor classes to enable three types of visitors based on the data requirement (simple computations, needing data point at leaf, or requiring bounding boxes) which are optimized . The optimization is non-trivial because it interacts with the maintenance/non-maintenance of the bounding boxes at a node (which affects the state/memory).

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/imputation/ConditionalSampleSummarizer.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/imputation/ConditionalSampleSummarizer.java
@@ -16,9 +16,7 @@
 package com.amazon.randomcutforest.imputation;
 
 import static com.amazon.randomcutforest.CommonUtils.checkArgument;
-import static java.lang.Math.max;
 import static java.lang.Math.min;
-import static java.lang.Math.sqrt;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -26,22 +24,10 @@ import java.util.List;
 
 import com.amazon.randomcutforest.returntypes.ConditionalTreeSample;
 import com.amazon.randomcutforest.returntypes.SampleSummary;
+import com.amazon.randomcutforest.summarization.Summarizer;
+import com.amazon.randomcutforest.util.Weighted;
 
 public class ConditionalSampleSummarizer {
-
-    /**
-     * the following determines the ratio between the sum of the (average) radius
-     * and the sepration between centers for a merge; ratio greater than 1 means
-     * significant overlap a ratio of 0 means merge closest pairs without
-     * consideration of separartion
-     **/
-    public static double SEPARATION_RATIO_FOR_MERGE = 0.8;
-
-    /**
-     * a factor that controls weight assignment for soft clustering; this is the
-     * multiple of the minimum distance and should be greater or equal 1.
-     */
-    public static double WEIGHT_ALLOCATION_THRESHOLD = 1.25;
 
     /**
      * this limits the number of valueswe would see per dimension; note that it may
@@ -72,16 +58,34 @@ public class ConditionalSampleSummarizer {
      */
     protected double centrality;
 
+    /**
+     * a boolean that determines if the summarization should use the missing
+     * dimensions or the full dimensions.
+     */
+    protected boolean project = false;
+
     public ConditionalSampleSummarizer(int[] missingDimensions, float[] queryPoint, double centrality) {
         this.missingDimensions = Arrays.copyOf(missingDimensions, missingDimensions.length);
         this.queryPoint = Arrays.copyOf(queryPoint, queryPoint.length);
         this.centrality = centrality;
     }
 
+    public ConditionalSampleSummarizer(int[] missingDimensions, float[] queryPoint, double centrality,
+            boolean project) {
+        this.missingDimensions = Arrays.copyOf(missingDimensions, missingDimensions.length);
+        this.queryPoint = Arrays.copyOf(queryPoint, queryPoint.length);
+        this.centrality = centrality;
+        this.project = project;
+    }
+
     public SampleSummary summarize(List<ConditionalTreeSample> alist) {
         checkArgument(alist.size() > 0, "incorrect call to summarize");
+        return summarize(alist, true);
+    }
+
+    public SampleSummary summarize(List<ConditionalTreeSample> alist, boolean addTypical) {
         /**
-         * first we dedupe over the points in the pointStore -- it is likely, and
+         * first we dedup over the points in the pointStore -- it is likely, and
          * beneficial that different trees acting as different predictors in an ensemble
          * predict the same point that has been seen before. This would be specially
          * true if the time decay is large -- then the whole ensemble starts to behave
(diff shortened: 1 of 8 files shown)


Rust/src/lib.rs
```diff
@@ -1,31 +1,30 @@
...
 extern crate rand;
+extern crate rand_chacha;
 
 use num::abs;
-use crate::rcf::{create_rcf, RCF};
-extern crate rand_chacha;
 
-pub fn L1distance(a: &[f32], b : &[f32]) -> f64{
+pub fn vec_l1distance(a: &Vec<f32>, b : &[f32]) -> f64 {
+    l1distance(a, b)
```
jotok, line 14: You should be able to write a single generic version of this function. Here's a version that that takes two iterators that iterate over `&f32`s. There should be a way to write it so that we only require that the types being iterated over can be cast to f64, but I was having trouble getting the type bounds right for that.

```rust
use num::abs;

fn l1distance<'a, L, R>(left: L, right: R) -> f64
where
  L: IntoIterator<Item=&'a f32>,
  R: IntoIterator<Item=&'a f32>,
{
    left.into_iter().zip(right).map(|(a, b)| abs(f64::from(*a) - f64::from(*b))).sum()
}

fn main() {
    let point1: Vec<f32> = vec![1.0, 3.0, 5.0];
    let point2: Vec<f32> = vec![2.0, -1.0, 7.0];
    let point2_as_slice: &[f32] = &point2[..];
    
    println!("|point1 - point2| = {}", l1distance(&point1, &point2));
    println!("|point1 - point2_as_slice| = {}", l1distance(&point1, point2_as_slice));
}
```

Rust/src/rcf.rs
```diff
@@ -31,6 +37,20 @@ pub(crate) fn damp(x: usize, y: usize) -> f64 {
 
 const max_number_in_summary : usize = 5;
 
+fn add_to(a: &f64, b : &mut f64){
```
jotok, line 40: Why do we need these specialized functions? Can't we do the dereferencing inline?

Rust/src/lib.rs
sudiptoguha: This seemed to be needed due to the paramteric_test macro -- it is possible something is amiss there. But changing to IntoIterator is forcing `for<'r, 's> fn(&'r [f32], &'s [f32]) ->  .... and then recursively it is forcing all the function call definitions to split. It would be a bad idea to get into annotations across the different calls, we would not be able to handle recursion. If this is galling then we should unroll the parametric test. We can also move this definition from the lib to the test.

Rust/src/rcf.rs
sudiptoguha: For this case we actually can use Sum() -- but that would not fit the pattern of {a1,a2,a3....} being collected in b (which is not necessarily the same type as that of a1, a2 ...). Maybe there is a better way to rewrite the single visitor traversal we are using now (for example when type of a is the same as that of b).

Rust/src/common/divector.rs
```diff
@@ -0,0 +1,174 @@
...
+            .map(|(((x, &y), &z), &b)| if !b && y - z > 0.0 {
+                *x = (y - z) as f64;
+                *x
+            } else {
+                *x = 0.0;
+                *x
+            })
+            .sum();
+        let maxsum: f64 = self.high
+            .iter_mut()
```
jotok, line 78: At some point we should do an experiment to see if there's any performance penalty to using all these `zip`s together versus a for-loop and accessing values by index.

Rust/src/common/samplesummary.rs
```diff
@@ -107,28 +100,17 @@ impl Center {
...
-        answer
-    }
-
-    pub fn recompute(&mut self,points:&[(&[f32],f32)],distance : fn(&[f32],&[f32]) -> f64){
+    pub fn recompute(&mut self,points:&[(T,f32)],distance : fn(&Q,T) -> f64){
         self.sum_of_radii  = 0.0;
         if self.weight == 0.0 {
             assert!(self.points.len() == 0, "adding points with weight 0.0 ?");
-            self.coordinate = vec![0.0;self.coordinate.len()];
+            //self.coordinate = vec![0.0;self.coordinate.len()];
```
jotok, line 107: Remove

Rust/src/rcf.rs
```diff
@@ -31,6 +37,20 @@ pub(crate) fn damp(x: usize, y: usize) -> f64 {
...
+    *b += *a;
+}
+
+fn divide(a : &mut f64, b: usize){
+    *a /= b as f64;
+}
+
+fn add_nbr(a:&(f64,usize,f64), b: &mut Vec<(f64,usize,f64)>){
```
jotok, line 48: Is this function used in the same way as the `add_to` and `divide` functions? Why is the first argument mutable in those but the second argument is mutable in this one?

Rust/src/rcf.rs
```diff
@@ -63,7 +83,50 @@ pub trait RCF {
...
+        self.score_visitor_traversal(point,&VisitorInfo::use_score(ignore_mass,score_seen,score_unseen,damp,normalizer))
+    }
+
+    fn score_visitor_traversal(&self, point :&[f32], visitor_info : &VisitorInfo) -> f64;
+
+    fn attribution(&self, point : &[f32]) -> DiVector {
+        self.attribution_visitor_traversal(point,&VisitorInfo::default())
+    }
+
+    fn generic_attribution(
```
jotok, line 96: Do we need this function? Is it really more convenient than letting users create their own VisitorInfo struct with the functions they want?

Rust/src/rcf.rs
```diff
@@ -262,18 +303,71 @@ where
...
-        list.sort_by(|o1,o2| o1.0.cmp(&o2.0));
+        list.sort_by(|&o1,&o2| o1.2.partial_cmp(&o2.2).unwrap());
         list
     }
+
+    pub fn simple_traversal<NodeView,V,R,S>(
+        &self,
+        point: &[f32],
+        parameters : &[usize],
+        visitor_info: &VisitorInfo,
```
jotok, line 319: Why is VisitorInfo it's own top-level struct that has to be passed into every traversal? Why aren't these parameters in the Visitor struct itself?

Rust/src/lib.rs
sudiptoguha: Fixed with changes

Rust/src/rcf.rs
sudiptoguha: moved to util. A subsequent PR may clean up the traversal.

Rust/src/common/divector.rs
sudiptoguha: Its a 50%+ penalty to not have these iterators

Rust/src/common/samplesummary.rs
sudiptoguha: fixed

Rust/src/rcf.rs
sudiptoguha: The idea of this function is a collector applied to an iterator ... it is convenient if the first is immutable. But a more focused PR can make these in line with the TAR collectors

Rust/src/rcf.rs
sudiptoguha: :) calling a function with different parameters (even if those parameters are functions) seems simpler than writing code -- specially in rust when the new code can start issues with the borrow checker/compiler

Rust/src/rcf.rs
```diff
@@ -63,7 +83,50 @@ pub trait RCF {
...
+        percentile: usize,
+    ) -> Vec<(f64,Vec<f32>,f64)> {
+        self.near_neighbor_traversal(point,percentile, &VisitorInfo::default())
+    }
+
+    fn near_neighbor_traversal(
+        &self,
+        point: &[f32],
+        percentile: usize,
+        visitor_info : &VisitorInfo
```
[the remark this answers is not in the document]
sudiptoguha: The visitorInfo is a refactoring of the information common to all visitors. The distance function is one such uniform function, like the scoring function etc. Copying these for each visitor makes the visitor structs larger and slower.

Rust/src/rcf.rs
```diff
@@ -262,18 +303,71 @@ where
...
+        list.sort_by(|&o1,&o2| o1.2.partial_cmp(&o2.2).unwrap());
         list
     }
+
+    pub fn simple_traversal<NodeView,V,R,S>(
+        &self,
+        point: &[f32],
+        parameters : &[usize],
+        visitor_info: &VisitorInfo,
+        visitor_factory : fn(usize,&[usize],&VisitorInfo) -> V,
```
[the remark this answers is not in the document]
sudiptoguha: Maybe go over this issue in a more focused PR? Completely plausible ...

Rust/src/rcf.rs
sudiptoguha: Saves a lot of space ...

PR #316, opened 2022-03-31
version bump to 3.0-rc2

amitgalitz:
Signed-off-by: Amit Galitzky <amgalitz@amazon.com>

*Issue #, if available:*
#304 
*Description of changes:*
This PR will version bump main to RCF 3.0-rc2. This doesn't mean current latest commit is good for stable release but we are preemptively bumping the current version to reflect next release. Once a RCF 3.0-rc2 is ready for release a 3.0 branch will be cut and the latest commit for the release will be tagged appropriately. This will also make it possible to implement a CICD that will upload snapshots of RCF on merge to main that will be labeled as 3.0-rc2-SNAPSHOT.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #309, opened 2022-03-28
fix for issue 305

sudiptoguha:
*Issue #, if available:* Issue 305. The most recent changes to main broke the state classes older models (version 2.0.1). 

*Description of changes:* Changes the PointStoreMapper to address the transformation and adds tests.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #322, opened 2022-04-14
handling models created via rc1 in rc2

sudiptoguha:
*Issue #, if available:* 321

*Description of changes:* If a model is created using 3.0- rc-1 then the initialization of unused part of the pointstore array is 0. Whereas in the newer encoding the unused part if INFEASIBLE_LOCN (-1). This causes models created using 3.0-rc1 to generate errors when used with 3.0-rc2. We recommend using the most recent models for efficiency of size and compute.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreMapper.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreMapper.java
@@ -15,20 +15,21 @@
 
 package com.amazon.randomcutforest.state.store;
 
-import static com.amazon.randomcutforest.CommonUtils.checkArgument;
-import static com.amazon.randomcutforest.CommonUtils.checkNotNull;
-import static com.amazon.randomcutforest.CommonUtils.toDoubleArray;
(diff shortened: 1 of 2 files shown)


jotok (approved):

ylwu-amzn:
Thanks for fixing this quickly. For `ml-commons`, 1.3.1 is using PR #309 to build local jar. But RCF RC2 is built with PR #316. Not sure if this bug introduced by which PR. Will it impact the model in `ml-commons` 1.3.1?

sudiptoguha:
Well, a numbering like rc2.1 is probably more appropriate. This change should not affect 1.3.1 -- because AFAIK the models are not being persisted (is that correct?). This issue arises because the two versions of the model use different representations for the unused part of the array, the two versions should be internally consistent. But it would be good to switch to most recent models as soon as possible, since it would get harder to fix bugs (should they appear and there likely is one somewhere)

ylwu-amzn:
`ml-commons` also stores model, using the same way of AD. But `ml-commons` 1.3.1 is using newer version of RCF. Good to know "two versions should be internally consistent". But will follow your suggestion to move to latest version once it's ready on Maven.

PR #324, opened 2022-04-25
cleanup/refactor of summarization

sudiptoguha:
*Description of changes:* Before rc2 we were performing imputation differently for single and multiple dimensions. This was resolved in rc2, however the summarization based method that was in use in rc1 has been exposed via conditional fields. This PR refactors that monolithic implementation, adds documentation (and tests) with the hope that the summarization can be used in a standalone manner. The behavior of the existing functions in the library is not affected. We aim to perform similar targeted cleanup as we get ready for releasing 3.0.


Java/core/src/main/java/com/amazon/randomcutforest/imputation/ProjectedPoint.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public float getCoordinate(int i) {
+        return coordinate[i];
+    }
+
+    public void accumulateForDeviation(double[] sumArray, double[] squareSumArray) {
+        for (int i = 0; i < coordinate.length; i++) {
+            checkArgument(!Float.isNaN(weight) && Float.isFinite(weight),
+                    " improper weight, accumulation not meaningful");
```
jotok, line 55: I think it's better when the error messages explicitly say what's wrong:
```
"weights must be finite, non-NaN values"
```

Java/core/src/main/java/com/amazon/randomcutforest/imputation/Summarizer.java
```diff
@@ -0,0 +1,371 @@
...
+    /**
+     * picks an item such that the prefix sum of the weight of the projectedPoints
+     * is at most wt
+     * 
+     * @param points a list of projected points (points with weight)
+     * @param wt     a selection parameter
+     * @return the index of the point for which the condition is satisfied (or the
+     *         last index)
+     */
+    public static int pick(ArrayList<ProjectedPoint> points, double wt) {
```
jotok, line 75: This is a very generic method name for very specialized functionality.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/Summarizer.java
```diff
@@ -0,0 +1,371 @@
...
+     * be used with intended spherical clusters as in here, or in future
+     * non-spherical clusters as well
+     *
+     * @param points      list of points with weights
+     * @param seed        a randomseed to control/reproduce the sampling
+     * @param totalWeight the sum of weights of the points (this can be approximate)
+     * @return a sample of points (preserving the same order)
+     */
+
+    static ArrayList<ProjectedPoint> createSample(ProjectedPoint[] points, long seed, double totalWeight) {
```
jotok, line 107: Maybe we should move this to a utility class and give it a name indicating that it's a specific sampling algorithm and maybe the the first one people would think of if you handed them a list of points and weights and asked them to create a sample.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/Center.java
```diff
@@ -0,0 +1,218 @@
...
+ * points
+ */
+public class Center {
+    /**
+     * a factor that controls weight assignment for soft clustering; this is the
+     * multiple of the minimum distance and should be greater or equal 1.
+     */
+    public static double WEIGHT_ALLOCATION_THRESHOLD = 1.25;
+
+    float[] coordinate;
```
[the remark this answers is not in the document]
sudiptoguha: using "representative", that also brings the naming in line with multi-centroid methods

Java/core/src/main/java/com/amazon/randomcutforest/imputation/Center.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.BiFunction;
+
+/**
+ * the following class abstracts a single centroid representation of a group of
+ * points
+ */
+public class Center {
```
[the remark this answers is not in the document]
sudiptoguha: Defined an interface ICluster -- it can enable other clustering algorithms; the use case in the imputation was to summarize a random conditional field. But clustering can be a standalone interest.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ProjectedPoint.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ProjectedPoint.java
```diff
@@ -0,0 +1,62 @@
...
+package com.amazon.randomcutforest.imputation;
+
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+
+/**
+ * at the current moment the following behaves as a weighted-point, however
+ * "Weighted-X" has been used elsewhere in the library and in the fullness of
+ * time, this class can evolve.
+ */
+public class ProjectedPoint {
```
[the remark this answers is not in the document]
sudiptoguha: This class is now removed/redundant based on the generic version of the WeightedIndex class

Java/core/src/main/java/com/amazon/randomcutforest/imputation/Summarizer.java
sudiptoguha: fixed.

Java/core/src/main/java/com/amazon/randomcutforest/util/WeightedIndex.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+/**
+ * a container class that manages weights
+ * 
+ * @param <Q>
+ */
+public class WeightedIndex<Q> {
```
jotok, line 29: Since there is no type restriction on Q, I would just call this `Weighted<Q>`.

Java/core/src/main/java/com/amazon/randomcutforest/util/WeightedIndex.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        ArrayList<WeightedIndex<Q>> samples = new ArrayList<>();
+        Random rng = new Random(seed);
+        double totalWeight = input.stream().map(x -> (double) x.weight).reduce(Double::sum).get();
+        double remainder = totalWeight;
+
+        if (heavyItems) {
+            remainder = input.stream().map(e -> {
+                if (e.weight > totalWeight * 5.0 / lengthBound) {
```
jotok, line 70: 5.0 seems arbitrary. What if instead of a `heavyItems` method argument we provide `forceSampleWeightFraction` or something like that. Then the user can say that if a point's weight is greater than a fraction of the total weight that they specify, then it will be included in the sample.

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/SampleSummary.java
```diff
@@ -83,4 +92,62 @@ public SampleSummary(double[] point) {
         this.deviation = new float[point.length];
     }
 
+    public SampleSummary(int dimension, float[][] typicalPoints, float[] relativeLikelihood) {
+        this.addTypical(dimension, typicalPoints, relativeLikelihood);
+    }
+
+    public void addTypical(int dimension, float[][] typicalPoints, float[] relativeLikelihood) {
```
jotok, line 99: If we are assigning `typicalPoints` to the `summaryPoints` array, why don't we just call the method argument `summaryPoints`?

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/SampleSummary.java
```diff
@@ -83,4 +92,62 @@ public SampleSummary(double[] point) {
...
+        checkArgument(typicalPoints.length == relativeLikelihood.length, "incorrect lengths of fields");
+        this.summaryPoints = new float[typicalPoints.length][];
```
jotok, line 101: This method completely overwrites the contents of the class. Is that the intention? Should this just be part of the constructor code?

Java/core/src/main/java/com/amazon/randomcutforest/summarization/Center.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    public static Center initialize(float[] coordinate, float weight) {
+        return new Center(coordinate, weight);
+    }
+
+    // adds a point; only the index to keep space bounds lower
+    // note that the weight may not be the entire weight of a point in case of a
+    // "soft" assignment
+    public void addPoint(int index, float weight, double dist) {
```
jotok, line 55: It's seems strange that this class is missing (what I assume are) point store indexes and actual point values. What's the purpose of the indexes?

Java/core/src/main/java/com/amazon/randomcutforest/summarization/ICluster.java
```diff
@@ -0,0 +1,57 @@
...
+public interface ICluster {
+
+    // adding a point to a cluster
+    void addPoint(int index, float weight, double distance);
+
+    // restting statistics for a potential reassignment
+    void reset();
+
+    // average distance of a point from a cluster representative
+    double average_radius();
```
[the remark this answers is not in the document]
sudiptoguha: Fixed. This is bleed over from the rust version ...

Java/core/src/main/java/com/amazon/randomcutforest/summarization/ICluster.java
```diff
@@ -0,0 +1,57 @@
...
+
+    // weight computation
+    double getWeight();
+
+    // is a point well expressed by the cluster
+    boolean captureBeforeReset(float[] point, BiFunction<float[], float[], Double> distance);
```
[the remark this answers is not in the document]
sudiptoguha: The intent of this function (not currently in use) is to perform a "contrafactual" estimation during the recompute phase (when the cluster representatives are optimized). We may want to know -- what would have been the L2 distance (even though we are using L1, and we have a separate L2 implementation). In fact the distance function is left in many of the functions for this precise use case.

Java/core/src/main/java/com/amazon/randomcutforest/util/WeightedIndex.java
sudiptoguha: Done.

Java/core/src/main/java/com/amazon/randomcutforest/util/WeightedIndex.java
sudiptoguha: Makes sense. The goal was 0.005% ... but setting a fraction directly is better.

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/SampleSummary.java
sudiptoguha: That was the intention -- the summary has both the basic statistical component and this "extra". We need to separate them -- we can have two constructors, but then repeating the basic statistical estimation would be too much. Let me take another shot.

Java/core/src/main/java/com/amazon/randomcutforest/summarization/Center.java
sudiptoguha: The index is a number that currently refers to the index of the point in the sampledPoint list. It is absolutely true :) that this same index can refer to a pointStore index and we would have  clustering of the pointstore (among many other things). AT the current moment, the (int) index serves as a space saver (in the spirit of CompactRCF) and if the same point is (soft) allocated to multiple clusters then there is no duplication. So the PointStore plug and play is the next step :)

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/SampleSummary.java
sudiptoguha: Done

Java/core/src/main/java/com/amazon/randomcutforest/summarization/Center.java
jotok: This interface method implying something about the implementation. Does it makes sense to include it in a generic cluster interface?

sudiptoguha: Fixed -- the interface is partitioned to a generic one and a more index specific one -- both of these can have usage down the road.

jotok (approved):

PR #325, opened 2022-05-06
Reopening the rust + refactor PR

sudiptoguha:
PR was closed by mistake ...


Rust/src/common/cluster.rs
```diff
@@ -0,0 +1,338 @@
...
+
+const PHASE2_THRESHOLD : usize = 2;
+const separation_ratio_for_merge : f64 = 0.8;
+const weight_threshold : f64 = 1.25;
+const LENGTH_BOUND : usize = 1000;
+
+pub trait Cluster<Q,T: ?Sized> {
+    fn get_weight(&self) -> f64;
+    fn distance_to_point(&self, point : &T, distance: fn(&T, &T) -> f64) -> f64;
+    fn distance_to_cluster(&self, other : &dyn Cluster<Q,T>, distance: fn(&T, &T) -> f64) -> f64;
```
jotok, line 15: Another approach to this in rust could be to define a trait for distance functions:
```
trait DistanceFn<A, B> {
  fn distance(a: A, b: B) ->  f64;
}
```

And then write implementations for the things we want to measure:
```
struct L1Distance {}

impl DistanceFn<f32, f32> for L1Distance {
    fn distance(&self, a: f32, b: f32) -> f64 {
        (a as f64 - b as f64).abs()
    }
}

impl<A, B> DistanceFn<Vec<A>, Vec<B>> for L1Distance
where
    A: Copy,
    B: Copy,
    L1Distance: DistanceFn<A, B>
{
    fn distance(&self, a: Vec<A>, b: Vec<B>) -> f64 {
       a.iter().zip(b.iter()).map(|(x, y)| self.distance(*x, *y)).sum()
    }
}
```

If we take this approach, we might be able to write a few generic impls for distance functions that cover a lot of use cases, and we wouldn't have to add `distance_to_foo` functions to other traits.

Rust/src/common/cluster.rs
```diff
@@ -0,0 +1,338 @@
...
+use rand_chacha::ChaCha20Rng;
+use rayon::iter::{IntoParallelRefMutIterator, ParallelIterator};
+use crate::common::samplesummary::SampleSummary;
+
+const PHASE2_THRESHOLD : usize = 2;
+const separation_ratio_for_merge : f64 = 0.8;
+const weight_threshold : f64 = 1.25;
+const LENGTH_BOUND : usize = 1000;
+
+pub trait Cluster<Q,T: ?Sized> {
```
jotok, line 12: We should document the trait functions. If someone wants to create a new cluster implementation, they need to know how they're expected to implement `recompute`, `reset`, etc.

Rust/src/common/cluster.rs
```diff
@@ -0,0 +1,338 @@
...
+        let dimensions = self.representative.len();
+        let total : f64 = self.points.iter().map(|a| a.1 as f64).sum();
+        for i in 0..dimensions {
+            self.points.sort_by(|a, b| points[a.0].0[i].partial_cmp(&points[b.0].0[i]).unwrap());
+            let position = pick(&self.points,(total/2.0) as f32);
+            self.representative[i] = points[self.points[position].0].0[i];
+        }
+    }
+
+    fn optimize(&mut self, points:&[(&[f32],f32)], list: &mut[(usize,f32)]) {
```
jotok, line 54: Documentation would be helpful here.

Rust/src/common/divector.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn assign(&mut self, other :&DiVector){
+        for (x, &y) in self.high.iter_mut().zip(&other.high) {
+            *x = y;
+        }
+        for (x, &y) in self.low.iter_mut().zip(&other.low) {
+            *x = y;
+        }
+    }
+
+    pub fn add_from(&mut self, other: &DiVector, factor : f64) {
```
jotok, line 107: Looks like you could implement the `AddAssign` trait instead:
https://doc.rust-lang.org/stable/std/ops/trait.AddAssign.html

We've discussed a larger reworking of this code when we revisit the logic of generic map-reduce operations, so I think it's OK to leave this for now.

Rust/src/common/divector.rs
```diff
@@ -0,0 +1,174 @@
...
+            *x += y * factor;
+        }
+        for (x, &y) in self.low.iter_mut().zip(&other.low) {
+            *x += y * factor;
+        }
+
+    }
+
+
+    pub fn add_to(&self, other: &mut DiVector) {
```
jotok, line 119: Couldn't you implement this as:
```
other.add_from(self)
```

Rust/src/common/cluster.rs
```diff
@@ -0,0 +1,338 @@
+use std::cmp::{max, min};
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha20Rng;
+use rayon::iter::{IntoParallelRefMutIterator, ParallelIterator};
+use crate::common::samplesummary::SampleSummary;
+
+const PHASE2_THRESHOLD : usize = 2;
+const separation_ratio_for_merge : f64 = 0.8;
+const weight_threshold : f64 = 1.25;
```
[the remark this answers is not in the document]
sudiptoguha: Done

Rust/src/common/cluster.rs
```diff
@@ -0,0 +1,338 @@
...
+const LENGTH_BOUND : usize = 1000;
+
+pub trait Cluster<Q,T: ?Sized> {
+    fn get_weight(&self) -> f64;
```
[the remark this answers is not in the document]
sudiptoguha: Done

Rust/src/common/cluster.rs
```diff
@@ -0,0 +1,338 @@
...
+    fn distance_to_point(&self, point : &T, distance: fn(&T, &T) -> f64) -> f64;
+    fn distance_to_cluster(&self, other : &dyn Cluster<Q,T>, distance: fn(&T, &T) -> f64) -> f64;
+    fn recompute(&mut self, points: &[(&T, f32)], distance: fn(&T, &T) -> f64) -> f64;
+    fn reset(&mut self);
+    fn average_radius(&self) -> f64;
+    fn add_point(&mut self, index: usize, weight: f32, dist: f64);
+    fn absorb(&mut self, another: &dyn Cluster<Q,T>, distance: fn(&T, &T) -> f64);
+    fn get_representatives(&self) -> Vec<Q>;
+    fn get_primary_representative(&self,distance: fn(&T, &T) -> f64) -> Q;
```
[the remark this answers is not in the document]
sudiptoguha: Done

Rust/src/common/cluster.rs
sudiptoguha: This approach a possibility -- but this would be the path to erasure of types and dynamic dispatch. (But the clusters themselves are dynamic dispatch already, but that is essential for extensibility). More of these traits can be a slowdown. But more importantly we may not have to solve this issue at this point.

The existence of two distance functions is fundamental. In this case the (&T, &T) -> f64 is the true distance function over immutable references to points. The Cluster however can (and does) a different representation of a point (thought another way, a cluster is a "different type" of an object). So any clustering algorithm should specify the distance to cluster -- as an extension of the point-to-point distance. So that Cluster--to--point is necessary, and so is Cluster--to--Cluster, for every new algorithm/implementation. So the trait definition just forces that clarity. Whether we use traits or just plain objects is probably less of an issue -- but we should probably index on simplicity.

Maybe shelve this and come back to revisit if we are staring at implementing  n x m distances? I expect 3 implementations per algorithm (as described) and no cross products.

Rust/src/common/cluster.rs
sudiptoguha: Agree. Done.

Rust/src/common/divector.rs
sudiptoguha: Agree. However add_to_scaled is the more fundamental function (given the current setup), changed.

Rust/src/common/divector.rs
sudiptoguha: Yes ;) But we need an explicit function that given collection of R's in R_1,R_2... iterates over and collects to a type S. In this case R = S; but need not be (think of approximate median of a sequence of values). Also, even though we wish to iterate, the type of the iterator would be different based on rayon par_iter() or native iter(). This will probably get rewritten once the TAR collector framework is implemented,

Rust/src/common/cluster.rs
sudiptoguha: Added

jotok (approved):

PR #326, opened 2022-05-11
rust: enabling dynamic directional density

sudiptoguha:
*Description of changes:* enables the dynamic density estimation (available in Java since V1.0) for Rust and brings the Rust version to (almost) feature parity with the Java RCF 2.0 (but using the layout and space saving of RCF3.0).

We note that the calibration/scaling is different from the Java version. This Rust version is calibrated such that for "high" density regions, the density would be less dependent on sample size. We note that density, in itself, can be less helpful than comparisons of densities (often their logarithms). The basic functions used are very similar to the displacement scoring used in the original paper and thus displacement_score() is enabled as well. Displacement score is calibrated to be in range [0,1] and less sensitive to the sample size (for potential anomalies).

### DIFF
--- a/Rust/src/common/directionaldensity.rs
+++ b/Rust/src/common/directionaldensity.rs
@@ -1,0 +1,128 @@
+
+use crate::common::divector::DiVector;
+use crate::samplerplustree::boundingbox::BoundingBox;
+
+#[repr(C)]
+#[derive(Clone)]
+pub struct InterpolationMeasure {
+    pub measure: DiVector,
+    pub distance: DiVector,
+    pub probability_mass: DiVector,
+    pub sample_size: f32
+}
+
+impl InterpolationMeasure {
+    pub fn empty(dimension: usize, sample_size: f32)  -> Self {
+        InterpolationMeasure {
+            measure: DiVector::empty(dimension),
+            distance: DiVector::empty(dimension),
+            probability_mass: DiVector::empty(dimension),
+            sample_size,
+        }
+    }
+
+    pub fn new(measure : DiVector, distance: DiVector, prob_mass:DiVector,sample_size: f32) -> Self{
+        assert!(measure.dimension() == distance.dimension(), " incorrect lengths");
+        assert!(measure.dimension() == prob_mass.dimension(), " incorrect lengths");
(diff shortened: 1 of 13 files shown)


Rust/src/common/directionaldensity.rs
```diff
@@ -0,0 +1,128 @@
...
+        InterpolationMeasure {
+            measure: DiVector::empty(dimension),
+            distance: DiVector::empty(dimension),
+            probability_mass: DiVector::empty(dimension),
+            sample_size,
+        }
+    }
+
+    pub fn new(measure : DiVector, distance: DiVector, prob_mass:DiVector,sample_size: f32) -> Self{
+        assert!(measure.dimension() == distance.dimension(), " incorrect lengths");
```
jotok, line 25: We should start to move away from panics and toward using Result types for validation, but we can do that in a future PR.

Rust/src/common/directionaldensity.rs
```diff
@@ -0,0 +1,128 @@
...
+        assert!(measure.dimension() == prob_mass.dimension(), " incorrect lengths");
+        InterpolationMeasure{
+            measure: measure.clone(),
```
jotok, line 28: Do we need to call `clone` here? It looks like the function takes ownership of the arguments.

Rust/src/common/divector.rs
```diff
@@ -163,4 +163,12 @@ impl DiVector {
         }
     }
 
+    pub fn dimension(&self) -> usize {
```
jotok, line 166: Do we call this `dimensions` (plural) in Java?

Rust/src/common/directionaldensity.rs
sudiptoguha: Agree.

Rust/src/common/directionaldensity.rs
sudiptoguha: No., not needed. Fixed.

Rust/src/common/divector.rs
sudiptoguha: Yes. Fixed.

jotok (approved):

PR #327, opened 2022-05-18
parallel execution test

sudiptoguha:
*Description of changes:* RCF has parallelism enabled via a specific thread pool implementation. There has been questions about using such (parameter ranges where parallelism helps etc. etc.). Over the long set of changes from V1.0 to now, it seems that parallelEnabled almost always helps (for a large range of parameters) for a single model. However it also seems that if there are a large number of models (as in high cardinality anomaly detection), it is verifiably better by some percentage to turn off parallelism within a model, but use multiple threads to process different models. The conclusions are testable for different settings of boundingboxcache parameter.

### DIFF
--- a/Java/core/src/test/java/com/amazon/randomcutforest/CPUTest.java
+++ b/Java/core/src/test/java/com/amazon/randomcutforest/CPUTest.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest;
+
+import com.amazon.randomcutforest.testutils.ShingledMultiDimDataWithKeys;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.concurrent.ForkJoinPool;
+
+@Tag("functional")
+public class CPUTest {
+
+    int numberOfTrees = 30;
+    int DATA_SIZE = 10000;
+    int numberOfForests = 6;
+    int numberOfAttributes = 5;
+    int shingleSize = 30;
+    int sampleSize = 256;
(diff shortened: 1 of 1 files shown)


Java/core/src/test/java/com/amazon/randomcutforest/CPUTest.java
```diff
@@ -0,0 +1,127 @@
...
+import com.amazon.randomcutforest.testutils.ShingledMultiDimDataWithKeys;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.concurrent.ForkJoinPool;
+
+@Tag("functional")
+public class CPUTest {
+
```
kaituo, line 27: Can you write the conclusion (async is 10%~15% faster than sync version) and purpose of the tests in the comments? Also, I usually write assertEquals or assertTrue in tests to verify sth, but I cannot find any verification here. Should we add some?

sudiptoguha: Would that conclusion remain true if the IDE/JVM/machine is changed? Also some folks like the JVM to be "warmed", so all those considerations come in before making a statement "X is faster than Y by Z". Also the conclusions can change based on the number of models (in this example 6) being tracked. :) 

The reason for providing this was a tool for folks to "test" the machine/environment they are considering running RCF in parallel. The verification is that the two methods are producing the exact same scores and forecast printed out. 
One could merge them as a test, but then one loses the granularity of time measurement. One could of course write all this up as a benchmarking suite -- but that needs resourcing.

kaituo: Got it. Can you add the purpose of these tests?

kaituo (approved):

Java/core/src/test/java/com/amazon/randomcutforest/CPUTest.java
```diff
@@ -0,0 +1,155 @@
...
+ * To summarize the lessons, it appears that parallelism almost always helps (upto resource limitations).
+ * If an user is considering a single model -- say from a console or dashboard, they should consider
+ * having parallel threads enabled. For large number of models, it may be worthwhile
+ * to also investigate different ways of achieving parallelism and not just attempt to
+ * change the executor framework.
+ *
+ */
+
+@Tag("functional")
+public class CPUTest {
```
jotok, line 51: What's the rationale for including this here and not in the benchmark package?

sudiptoguha: Resourcing. A good benchmarking needs calibration and needs to handle all of the caveats mentioned in the comments. This simple test helps reveal to an user which path they need to follow, and should be of interest because enabling parallelism (toggling the numberOfThreads) shows improvement in both async/sync versions, contrary to experiments in some prior version.

jotok (approved):

PR #329, opened 2022-05-27
fix for bit conversion

sudiptoguha:
*Issue #, if available:* 328

*Description of changes:* The index manager used in multiple store classes used tobits() to convert an array into a sequence of bits during deserialization and as a consequence 256 was considered as 0. This appeared only in the case of PointStore (and in a presence of sufficiently many duplicates) because the trees were small. Once the PointStore is corrupted, it would start showing "incorrect state" errors or attempt to delete incorrect points.


Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
```diff
@@ -744,6 +744,19 @@ public PointStore(PointStore.Builder builder) {
                         : new float[dimensions];
             }
 
+            /*
+             * an explicit conversion to bits; using toBits(refCount) in the constructor of
+             * IndexIntervalManager converts 256 into 0 and for a location appearing 256
```
jotok, line 749: Can we fix this in the IndexIntervalManager instead?

kaituo (approved):

Java/core/src/main/java/com/amazon/randomcutforest/store/PointStore.java
sudiptoguha: Done

jotok (approved):

PR #330, opened 2022-05-27
first steps of forecast

sudiptoguha:
*Description of changes:* RCF already provides a forecast/extrapolate functionality based on repeated estimation of the conditional median. This PR exposes the errors in the estimation of those conditional medians. They may be helpful in. estimating confidence bounds using model assumptions extraneous to the non-parametric estimation. In the next few PRs we would enable forecasting for TRCFs as well.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
@@ -569,16 +569,12 @@ public class RandomCutForest {
 
     /**
      *
-     * @return the sequence index of the last known shingled point if internal
-     *         shingling is not enabled, then the expectation would be the last
-     *         position
+     * @return the sequence index of the last known shingled point. If internal
+     *         shingling is not enabled, then this would correspond to the number of
+     *         updates
      */
     public long nextSequenceIndex() {
-        if (!internalShinglingEnabled) {
-            return shingleSize - 1;
-        } else {
-            return stateCoordinator.getStore().getNextSequenceIndex();
-        }
+        return stateCoordinator.getStore().getNextSequenceIndex();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -568,11 +569,16 @@ public float[] lastShingledPoint() {
 
     /**
      *
-     * @return the sequence index of the last known shingled point
+     * @return the sequence index of the last known shingled point if internal
```
jotok, line 572: If shingle-size is 10 and internal shingling is disabled, then `nextSequenceIndex` should always return 9?

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
```diff
@@ -174,6 +182,13 @@ public void acceptLeaf(final INodeView leafNode, final int depthOfNode) {
             }
         }
 
+        if (centrality < 1.0) {
+            Random rng = new Random(randomSeed);
+            randomSeed = rng.nextLong();
+            rng.nextDouble(); // discard
```
jotok, line 188: Why?

jotok (approved):

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
sudiptoguha: Fixed. I was confused/forgot what the function was doing.

Java/core/src/main/java/com/amazon/randomcutforest/imputation/ImputeVisitor.java
sudiptoguha: I was attempting to decor relate a linear congruential pseudo random generator. But really this cannot be solved in this library. Removing.

jotok (approved):

PR #333, opened 2022-06-10
adding forecasts to thresholded models

sudiptoguha:
*Description of changes:* RCFs, just like any RF models, can be used to forecast and the extrapolate primitive has always been available in basic RCFs. This PR adds that capability to the (relatively easier) thresholded models. As a consequence, the forecasts may also be more robust because of the predictor-corrector design of the thresholded models. It also may be possible to forecast signals with simpler drifts. To enable forecast for every transformation would require refactoring of the preprocessors which would be subsequent PRs. However adding the forecast functionality now would allow testing/verifying the subsequent refactor.


Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -240,6 +243,89 @@ public AnomalyDescriptor process(double[] inputPoint, long timestamp, int[] miss
...
+                        && transformMethod != TransformMethod.NORMALIZE_DIFFERENCE)
+                        || horizon <= preprocessor.getShingleSize() / 2 + 1,
+                "reduce horizon or use a different transformation, single step differencing will be noisy");
+        int shingleSize = preprocessor.getShingleSize();
+        checkArgument(shingleSize > 1, "extrapolation is not meaningful for shingle size = 1");
+        // note the forest may have external shingling ...
+        int dimensions = forest.getDimensions();
+        int blockSize = dimensions / shingleSize;
+        double[] lastPoint = preprocessor.getLastShingledPoint();
+        boolean ifZero = (forest.getBoundingBoxCacheFraction() == 0);
```
jotok, line 298: what if we called this `cacheDisabled`.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/preprocessor/transform/DifferenceTransformer.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * note that weight == 0, would produce 0 values in the inversion
+     *
+     * @param values        what the RCF would like to observe
+     * @param previousInput what was the real (or previously imputed) observation
+     * @return the observations that would (approximately) transform to values[]
+     */
+    public double[] invert(double[] values, double[] previousInput) {
```
jotok, line 41: minor: is this an inherited method? can we put the `@Override` annotation on it?

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
sudiptoguha: Done. Also fixed at the earlier place in singleStepProcess().

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/preprocessor/transform/DifferenceTransformer.java
sudiptoguha: Done. Also fixed for the other transforms.

jotok (approved):

Issue #303, opened 2022-03-14
Add a CD pipeline that releases this project to maven


dblock:
Currently this project is built on a developer machine, and occasionally pushed to maven.

1. Add a CD process that publishes a -SNAPSHOT to a snapshot maven repo with every commit.
2. Add a CD process that publishes a release to maven central on demand.

dblock:
This will close https://github.com/aws/random-cut-forest-by-aws/issues/301

dblock:
RCF is not part of opensearch-project, so I don't think it should use any OpenSearch project infrastructure unless you want to move the repo into that organization

I would follow https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-gradle

ylwu-amzn:
Thanks @dblock , do you think we should follow the same practice of OpenSearch:  publish pre-release RCF to sonatype repo first. Once test done, we can publish to public maven(maven central).

dblock:
> Thanks @dblock , do you think we should follow the same practice of OpenSearch: publish pre-release RCF to sonatype repo first. Once test done, we can publish to public maven(maven central).

Yes, that's what most projects do, right?

ylwu-amzn:
From AD experience, RCF added TRCF and bumped version to 2.0, then we use local jar file to test TRCF in AD and found some bugs, then RCF fixed these bugs and back to AD. We did several iteration to reach a stable version. I think we should follow the same practice of OpenSearch and publish RCF to sonatype repo first. Once the new RCF version tested in AD and other places like MLCommmons, we can publish to public maven.

ylwu-amzn:
Yes, the story much like SQL and MLCommons plugin, we publish MLCommons `1.3-SNAPSHOT` first, then SQL will consume this `1.3-SNAPSHOT` to develop new PPL ML command and test. The difference is we have same release date for all plugins and OpenSearch core. So we can hold all `-SNAPSHOT` on sonatype first, once the IT passed and OpenSearch bundle test done, we can release OpenSearch and publish to public Maven. 

For RCF, as this is not part of OpenSearch, maybe we can follow different release timeline and decouple the testing. RCF should test and make sure all new features work as expected first, then they can choose to publish as `-SNAPSHOT` or final public version. We don't need to test RCF new features in AD or MLCommons. If AD or MLCommons side found any bug, we can report to RCF and ask for a fix and new version, just like using any other external lib.

dblock:
[...]

Issue #301, opened 2022-03-10
Publish latest code to maven


ylwu-amzn:
Now [AD](https://github.com/opensearch-project/anomaly-detection/tree/1.3/lib) and [ml-commons](https://github.com/opensearch-project/ml-commons/tree/1.3/ml-algorithms/lib) are using local jar dependency to use latest RCF code. This is not best practice. How about we publish the latest RCF code to maven to remove such local jar dependency?

Related issue https://github.com/opensearch-project/index-management/issues/123

dblock:
https://github.com/opensearch-project/anomaly-detection/issues/433

dblock:
I opened https://github.com/aws/random-cut-forest-by-aws/issues/303 to get a CI/CD pipeline instead of doing it by hand.

sudiptoguha:
A CI/CD pipeline would be a great improvement.

sudiptoguha:
closed

Issue #97, opened 2020-12-15
Test coverage: state package


jotok:

amitgalitz:
@jotok do you remember what the intention behind this issue was? Is it to add a test coverage label in github? To increase test coverage in specific packages?

sudiptoguha:
Line coverage is over 90+% in all the packages. Branch coverage is 85+%

PR #334, opened 2022-06-17
Maven release and snapshot workflows

amitgalitz:
*Issue #, if available:*
#303
#301
*Description of changes:*

#### **maven-release workflow:**
This workflow is executed manually by a repo maintainer when releasing an official version to maven. It will publish the stable release to a staging repo where with one click we can publish it fully to maven central. It will also create a tag and a github release. If someone accidentally runs this when version still has `-SNAPSHOT`, there is extra safety to stop workflow.

#### **maven-snapshot:** 
releases a snapshot when commit is pushed to main and version is `-SNAPSHOT`.

Changed all version numbers to `3.0-rc4-SNAPSHOT`. We will be changing to a better practice in what version RCF is on when developing and releasing. The new process will involve incrementing to the next version right after a release as all new development is technically a part of the next release. We are also adding a `-SNAPSHOT` qualifier to the end of the version to indicate that this current version is during development.

_**side-note:**_ it is up to us to decide if we want to move the next development iteration to `3.0-rc4-SNAPSHOT` or `3.0-SNAPSHOT` and that is something I would like feedback on.

### General flow example:
1. 3.0.0 is released by manual click run on the new github `maven-release` workflow (more details in bottom)
2. We submit a PR to increment to 3.1.0-SNAPSHOT (we are actively developing features for 3.1 release)
    - This can be done easily with one command `mvn versions:set -DnewVersion=3.1.0-SNAPSHOT -DgenerateBackupPoms=false`
3. We are actively developing features that will be a part of 3.1 release, on every merge to main the latest snapshot can be fetched from the snapshot repository
4. We decide that we want to release 3.1 as on official release, we now submit a PR to bump the version from `3.1.0-SNAPSHOT` to `3.1.0`.
   - This can be done with one easy command: `mvn versions:set -DremoveSnapshot -DgenerateBackupPoms=false`
5. We manually run (1 click) the `maven-release` workflow, new tag is created, a new Github release created, and artifacts are uploaded to staging repository. To finalize this release we login to nexus and ensure it is closed and click release to officially push to maven central. 

The last step where we release to maven, we are actually only releasing to the staging repository where we have to login and click release one more time. This adds an extra layer of security and fault tolerance, however we can easily change this in the future to skip the staging part. For now I think we should first release just to the staging repository at least for the first official release with this new workflow. 

### Next steps:
- I spent some time trying to convert the repository over to gradle since gradle makes it a lot easier to create custom tasks, a more concise build script and overall better development experience. Gradle also is known to have much faster build times. However I was having multiple different errors with aligning dependencies and having everything work. I decided this should be a different task as we want to make sure we do it right and ensure the jar made from gradle is identical to maven.
- Create more workflows for auto incrementing versions and adding code coverage

### DIFF
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -1,30 +1,35 @@
 # This workflow will build a package using Maven and then publish it to
 # Maven central each time a release is created on github.
 
-name: Publish Maven Package
+name: Publish Official Release to Maven Staging
 
 defaults:
   run:
     working-directory: Java/
 
 on:
-  release:
-    types: [created]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the release with, e.g., 1.2.0, 1.2.1-alpha.1
+        required: true
+
 
 jobs:
   build:
-
+    name: Build project and publish release
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v2
     - name: Set up JDK 1.8
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-        server-id: ossrh
+        server-id: ossrh-release
         server-username: MAVEN_USERNAME
         server-password: MAVEN_PASSWORD
+        gpg-private-key: ${{ secrets.MAVEN_GPG_PRIVATE_KEY }}
+        gpg-passphrase: MAVEN_GPG_PASSPHRASE
 
     - name: Build with Maven
       run: mvn -B package --file pom.xml
@@ -34,4 +39,4 @@ jobs:
       env:
         MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
         MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
-
+        MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -1,6 +1,11 @@
 # This workflow will build a package using Maven and then publish it to
-# Maven central each time a release is created on github.
+# to our staging repo for a final release to maven central
 
+# This workflow will also create a tag and github release for the current commit
+# The github release will be have '-java' added to name to distinguish rust vs java releases
+# Example:
+# current version in POM.XML and in maven central = 3.1.0
+# tag and github release = 3.1.0-java
 name: Publish Official Release to Maven Staging
 
 defaults:
@@ -9,11 +14,6 @@ defaults:
 
 on:
   workflow_dispatch:
-    inputs:
-      version:
-        description: The version to tag the release with, e.g., 1.2.0, 1.2.1-alpha.1
-        required: true
-
 
 jobs:
   build:
@@ -40,3 +40,16 @@ jobs:
         MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
         MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
         MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
+    - name: Extract project version
+      id: project
+      run: echo ::set-output name=version::$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+    - name: Create Release
+      id: create_release
+      uses: actions/create-release@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        tag_name: ${{ steps.project.outputs.version }}-java
+        release_name: ${{ steps.project.outputs.version }}-java
+        draft: false
+        prerelease: false
--- a/.github/workflows/maven-release.yml
+++ b/.github/workflows/maven-release.yml
@@ -30,19 +30,19 @@ jobs:
         server-password: MAVEN_PASSWORD
         gpg-private-key: ${{ secrets.MAVEN_GPG_PRIVATE_KEY }}
         gpg-passphrase: MAVEN_GPG_PASSPHRASE
-
+    - name: Extract project version
+      id: project
+      run: echo ::set-output name=version::$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
     - name: Build with Maven
       run: mvn -B package --file pom.xml
-
     - name: Publish to Maven central
+      if: ${{ !endsWith(steps.project.outputs.version, '-SNAPSHOT') }}
       run: mvn deploy --file pom.xml
       env:
         MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
         MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
         MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
-    - name: Extract project version
-      id: project
-      run: echo ::set-output name=version::$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+    
     - name: Create Release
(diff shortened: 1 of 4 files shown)


ylwu-amzn (approved):
LGTM. Thanks for making the change!

sudiptoguha:
Thanks for the PR. But (a) this PR should be serialized after PR 333 -- currently there is a version in Maven which is most recent and another RC4 is not needed. (b) the next planned version (after 333 is resolved) should be 3.0.0. What is schedule for testing/benchmarking to be completed ?

sean-zheng-amazon:
@sudiptoguha we are definitely prioritizing the benchmarking work, and the experiments will be wrapped up by end of June. Meanwhile since this PR is more about the ci/cd set up, should we have it in first to benefit future releases?

amitgalitz:
> Thanks for the PR. But (a) this PR should be serialized after PR 333 -- currently there is a version in Maven which is most recent and another RC4 is not needed. (b) the next planned version (after 333 is resolved) should be 3.0.0. What is schedule for testing/benchmarking to be completed ?

I don't think the PR itself needs to wait for the testing/benchmarking work or to be after pr 333 is merged. However if you believe rc4 doesn't make since I can change the version to be 3.0-SNAPSHOT instead. 3.0-rc3-SNAPSHOT wouldn't make sense as that is supposed to be before 3.0-rc3 official release which already happened.

sudiptoguha:
As I have said above, let me say it again, we will prioritize PR 333 based on the scientific mission and content of the library. The issues referee to in the PR are 90+ days old, it is unseemly to clamor for a fact action after 90+ days of delays. We will discuss this PR after benchmarks are done -- (a) some changes may be required (I do not think so, but there is no need to prejudge) and (b) I believe that holding a firm line is the on ly way the benchmarking will ever be completed up to scientific fidelity.

sudiptoguha:
Given that 333 is now merged, can the version numbers be changed to 3.0.0 ? We begin the release to maven -- unless of course there are emergent issues from the benchmarking/etc. tests.

dblock:
Can the 3.0 version increment be done separately by the person releasing the library? That version increment otherwise would be conflated with the GH workflows added here. I'd want to see a separate commit that says "version now is 3.0.0".

sudiptoguha:
So far the assumption was that Amit was the person releasing the library, but the increment can happen independent of him.

dblock:
@amitgalitz looks like the version conflicts in this PR need to be resolved, but looking forward to snapshot/build/release automation!

sean-zheng-amazon:
@sudiptoguha agree that we should now bump the version to 3.0, Amit will do that when he back. Meanwhile is there any other concern/dependencies before we merge this PR to enable the ci/cd?

sudiptoguha:
The version discussion was never a matter of negotiation, I am not sure how that got mixed in an automation PR in the first place. 

That being said I am reading next steps as "I spent some time trying to convert the repository over to gradle since gradle makes it a lot easier to create custom tasks, a more concise build script and overall better development experience. Gradle also is known to have much faster build times. However I was having multiple different errors with aligning dependencies and having everything work. I decided this should be a different task as we want to make sure we do it right and ensure the jar made from gradle is identical to maven. Create more workflows for auto incrementing versions and adding code coverage". 

That seems to be the bar. Open source should raise bars.

 If there are multiple different errors with aligning dependencies of this library (which has minimal dependencies) then I foresee that exact same issue will be present in every automation script that consumes this library. @dblock, thoughts?

Why not do things right? It is of course ok to stage things -- but halfway solutions are habit of being requiring attention again.

dblock:
I rebased this PR for ya, https://github.com/aws/random-cut-forest-by-aws/pull/337, in case you're ready to merge it.

amitgalitz:
@dblock @sudiptoguha Just read through all the comments as I was out for a little bit. Since there might be some issues showing up from benchmarking I still think there is a case to be made to making the version 3.0.0-SNAPSHOT until that's been decided but we can also just keep at 3.0.0 for now and not do the move back to -SNAPSHOT and then back to 3.0.0. Other then that I can rebase this PR and close the one you made if that is okay @dblock? In regards for gradle, I don't think Maven is a halfway solution if I made it seem like that in my description. For RCF currently it provides the benefits of handeling `-SNAPSHOT` to release more smoothly and faster build times (which are documented as a whole but not for this repo yet) but maven is still a good solution. I think it shouldn't be a reason to stop this PR but it could be part of a larger issue to potentially move to gradle and integrate other things like code coverage as I assume this issue is referring too (https://github.com/aws/random-cut-forest-by-aws/issues/97)

sean-zheng-amazon:
+1. I honestly don't see obvious benefits converting maven to gradle for rcf, especially considering the minimal dependencies the library has (gradle is more flexible on dependency management rules for sure). There are pros and cons between them, but both are used by a majority of java applications for production. I do agree with Amit that we should merge this to enable ci/cd asap and looking into gradle conversion in parallel, but we should also be open to suggestions/concerns from repo owners. @sudiptoguha @jotok , thoughts?

dblock:
@amitgalitz you should rebase this one, I thought you were out for a while, I closed my PR

Re: version. I am used to seeing `3.0.0-SNAPSHOT` in source code, and every commit gets built and published as a snapshot, until the maintainer (@sudiptoguha) decides to release, at that time they switch it to `3.0.0`, release, and increment to `3.0.1-SNAPSHOT`. 

@sudiptoguha Are you saying something else? Other than a rebase, what needs to happen in this PR to get merged?

sudiptoguha:
answering @dblock -- I (and Josh) are ready to release RCF 3.0 to maven. That was the PR 335.

Since this project had no resourcing, it was my understanding that Amit would be the person to release to Maven. 

Independently AD plugin required tooling for its build.

It seems to me that the requirement of tooling has bled into the decision mechanism of who gets to decide RCF is released.

dblock:
Thanks @sudiptoguha, I understand! Let's leave tooling alone.

This PR 1) adds a workflow to publish SNAPSHOT builds to maven, and 2) adds a workflow to publish the final release to maven. I think both make maintainers life a lot easier, so I recommend rebasing this PR, merging, ensuring -SNAPSHOT works, then using the new workflow to release. This way we know it works for next time. And I think @amitgalitz doing it sounds good to me ;)

amitgalitz:
>Given that 333 is now merged, can the version numbers be changed to 3.0.0 ? We begin the release to maven -- unless of course there are emergent issues from the benchmarking/etc. tests

I didn't mean to make it seem like I am deciding when RCF should be released. I just wanted to make sure I address this previous comment and be transparent of what I am trying to reproduce on my end, but as I still haven't been able to reproduce the issue in a unit test it's fair to not base it off this. I agree the correct way would be to reproduce and open a Github issue on RCF not base RCF on some things I notice on AD and so far I am not able to to 100% say I have a new **reproducible** bug.

I'll rebase this PR now, the only conflicting messaging I am getting is that if I want to test out the `-SNAPSHOT` I should change the version to 3.0.0-SNAPSHOT for now, make sure that works on this PR merging and then submit a new PR changing the version to 3.0.0 for the official release (all today). If I want to test out the snapshot workflow which is less invasive first as @dblock is recommending. Is that okay or should I just go ahead with 3.0.0 as is right now?

sudiptoguha:
@dblock - makes sense.
@amitgalitz (A) please release the current version to maven and (B) use 3.0.1-SNAPSHOT 

There is a possibility that there are issues at next release but let's not change versions back and forth. There will be bugs in this RCF code (now or some day). They will get fixed (to the ability we possess).

amitgalitz:
Okay rebasing this PR so its set to 3.0.0 and then will run release workflow and then will increment codebase to 3.0.1

sudiptoguha:
Thanks @amitgalitz!

dblock:
@amitgalitz Do you want to open a new proposal for moving the infrastructure here to Gradle with the pros/cons/benefits and getting @sudiptoguha's opinion before you/anyone else does more work on it?

PR #336, opened 2022-06-29
 Update the RCF trait to use Result in return types

jotok:
Currently the library uses asserts and panics when errors are encountered. This change updates the RCF trait to use a Result type instead to allow for error handling (see https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html). In future PRs we will extend this pattern to other modules.

- Functions in the RCF trait interface that can fail are given a Result return type.
- Some existing asserts are replaced a function that returns a Result.
- Remove the `get` prefix from simple getters in the RCF trait interface to follow the Rust convention.
- Additional changes were made by `cargo fmt`

### DIFF
--- a/Rust/src/rcf.rs
+++ b/Rust/src/rcf.rs
@@ -1,29 +1,34 @@
-
 extern crate num;
 extern crate rand;
 extern crate rand_chacha;
 
-use rayon::prelude::*;
-use rand::SeedableRng;
 use core::fmt::Debug;
+
+use rand::SeedableRng;
 use rand_chacha::ChaCha20Rng;
 use rand_core::RngCore;
-use crate::common::conditionalfieldsummarizer::FieldSummarizer;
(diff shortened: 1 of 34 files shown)


Rust/src/rcf.rs
```diff
@@ -662,42 +662,42 @@ where
             visitor_info,
         );
         let field_summarizer = FieldSummarizer::new(centrality, project, max_number, l1distance);
-        field_summarizer.summarize_list(&self.point_store, &raw_list, &new_positions)
+        Ok(field_summarizer.summarize_list(&self.point_store, &raw_list, &new_positions))
```
sudiptoguha, line 665: This will probably be edited in the next few rounds -- but would the unwrap() be automatic within the OK()?

sudiptoguha (approved):

PR #342, opened 2022-07-19
rust: initialing TRCF

sudiptoguha:
*Description of changes:* initiates parkservices/TRCF for the rust version.

### DIFF
--- a/Rust/src/common/anomalydescriptor.rs
+++ b/Rust/src/common/anomalydescriptor.rs
@@ -10,16 +10,16 @@ use crate::common::divector::DiVector;
 pub struct AnomalyDescriptor {
 
     // the current input point; can have missing values
-    pub current_input: vec<f32>,
+    pub current_input: Vec<f32>,
 
     // current timestamp
     pub current_timestamp: usize,
 
     // potential missing values in the current input (ideally None)
-    pub missing_values: Option<vec<i32>>,
+    pub missing_values: Option<Vec<i32>>,
 
     // potentially transformed point used by RCF, can have different dimensions than input
-    pub rcf_point: Option<vec<f32>>,
+    pub rcf_point: Option<Vec<f32>>,
 
     pub score: f32,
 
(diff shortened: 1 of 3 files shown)


Rust/src/common/anomalydescriptor.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub past_timestamp: Option<usize>,
+
+    pub expected_values_list: Option<Vec<Vec<f32>>>,
+
+    pub likelihood_of_values: Option<Vec<f32>>
+
+}
+
+impl AnomalyDescriptor {
+    pub fn default(point: Vec<f32>, timestamp: usize) -> Self {
```
jotok, line 64: In Rust, `default` is typically used to define 0-argument constructors. Should we call this `new` and rename the second constructor to `new_with_missing_values` or something like that?

Rust/src/common/deviation.rs
```diff
@@ -0,0 +1,113 @@
...
+        let temp = self.sum / self.weight;
+        let answer = self.sum_squared / self.weight - temp * temp;
+        if answer > 0.0 {
+            f64::sqrt(answer)
+        } else {
+            0.0
+        }
+    }
+
+    pub fn empty(&self) -> bool{
```
jotok, line 82: I think it's typical to use `is` for boolean functions, like `is_empty`.

Rust/src/common/deviation.rs
```diff
@@ -0,0 +1,113 @@
...
+        self.weight <= 0.0
+    }
+
+    pub fn discount(&self) -> f64 {
+        self.discount
+    }
+
+    pub fn set_discount(&mut self, discount:f64) {
```
jotok, line 90: An alternative to creating all these setters would be to annotate the struct fields as `pub`, and then the caller could just assign to the struct field directly.

Rust/src/common/anomalydescriptor.rs
sudiptoguha: done

Rust/src/common/deviation.rs
sudiptoguha: Done

Rust/src/common/deviation.rs
sudiptoguha: Let's keep this one as is for now -- since there are so few fields in this structs. But some of the other structs are much longer ...

jotok (approved):

PR #350, opened 2022-10-16
Add support for boxing errors

ankane:
*Issue #, if available:* None

*Description of changes:*

Adds support for [boxing errors](https://doc.rust-lang.org/rust-by-example/error/multiple_error_types/boxing_errors.html)

### DIFF
--- a/Rust/src/errors.rs
+++ b/Rust/src/errors.rs
@@ -1,7 +1,20 @@
+use std::error;
+use std::fmt;
+
 /// Errors that can be returned by RCF operations.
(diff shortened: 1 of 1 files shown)


sudiptoguha (approved):
I may not be the best person to judge if Formatter the final necessary abstraction -- but this seems like a reasonable step for making progress and can always be revisited if needed.

sudiptoguha:
Thanks!

ankane:
Thanks! (fwiw, `std::error::Error` requires `std::fmt::Display` to be implemented)

PR #354, opened 2022-11-11
RCFCast

sudiptoguha:
*Issue #, if available:* 352

*Description of changes:* This PR introduces a calibrated TRCF which can be used in forecasting (and self calibrating). The basic operation is a TRCF but the sequential analysis is necessary for calibration. To this end the PR also provides SequentialAnalysis which takes the guesswork and inadvertent interpretations out of sequential analysis.

We note that calibration is a vast field; and the strong emphasis on sequential analysis can be seen in
A.P. Dawid's opinions of "Prequential Analysis", in _Present Position and Potential Developments: Some Personal Views: Statistical Theory: The Prequential Approach_, Journal of the Royal Statistical Society. Series A (General)
Vol. 147, No. 2, The 150th Anniversary of the Royal Statistical Society (1984), pp. 278-292 , https://www.jstor.org/stable/2981683

We recommend the article and note that in conclusion David emphasizes Markov models, cluster analysis and kernel width, Kalman-Bucy filters and many other topics of general Stochastic Processes. Serendipitously, RCF has many of these capabilities, and in fact as "reverse the usual direction of historical view" (a reference to the same article)  is a natural foil of calibration. 

At the same time, we differ in two positions. Both of these arise from assertions using Jeffrey's law that for Prequential Analysis, "in just those cases where we cannot choose empirically between several forecasting systems, it turns out that we have no need to do so". From an algorithmic standpoint it is obvious that (i) different algorithms would have different resource consumption, state requirements and (ii) we may simply not have enough data for a full blown probabilistic interpretation.Also (almost) the entire development of streaming algorithms is post 1984, and many of the resource consumption issues are intimately tied to formal notion of streaming algorithms.  

And finally, if forecasts are to be consumed by alerts, it is likely that the alert is most indicative of a breakdown of stationarity and calibration is less useful at precisely those moments of alerts. Thus, even if we can calibrate, we recommend that user of calibration actively chooses to do so and decide if it is useful.

To try it out check out examples/src/main/java/com/amazon/randomcutforest/examples/parkservices/RCFCasterExample.java

### DIFF
--- a/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ErrorHandler.java
+++ b/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ErrorHandler.java
@@ -221,7 +221,7 @@ public class ErrorHandler {
      *                   these error percentiles will be less meaningful
      * @param newHorizon a horizon which could be smaller than the errorHorizon (for
      *                   auditing)
-     * @param error      the function that defies error
+     * @param error      the function that defines error
      * @return a RangeVector where the value field corresponds to the p50 and the
      *         upper and lower arrays correspond to percentile and 1-percentile
      *         quantiles. We reiterate that do be cautious in use of percentiles.
--- a/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ErrorHandler.java
+++ b/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ErrorHandler.java
@@ -1,0 +1,372 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.randomcutforest.parkservices;
+
+import static com.amazon.randomcutforest.CommonUtils.checkArgument;
+import static java.lang.Math.abs;
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+
+import java.util.Arrays;
+import java.util.function.BiFunction;
+
+import com.amazon.randomcutforest.parkservices.calibration.Calibration;
+import com.amazon.randomcutforest.returntypes.DiVector;
+import com.amazon.randomcutforest.returntypes.RangeVector;
+
+public class ErrorHandler {
+
+    /**
+     * this to constrain the state
+     */
+    public static int MAX_ERROR_HORIZON = 1024;
+
+    int sequenceIndex;
+    double percentile;
+    int forecastHorizon;
+    int errorHorizon;
+    protected RangeVector[] pastForecasts;
+    protected float[][] actuals;
+    // the following are derived quantities and present for efficiency reasons
+    RangeVector errorDistribution;
+    DiVector errorRMSE;
+    float[] errorMean;
+    DiVector multipliers;
+    float[] intervalPrecision;
+
+    public ErrorHandler(RCFCaster.Builder builder) {
+        checkArgument(builder.errorHorizon >= builder.forecastHorizon,
+                "intervalPrecision horizon should be at least as large as forecast horizon");
+        checkArgument(builder.errorHorizon <= MAX_ERROR_HORIZON, "reduce error horizon of change MAX");
+        forecastHorizon = builder.forecastHorizon;
+        errorHorizon = builder.errorHorizon;
+        int inputLength = (builder.dimensions / builder.shingleSize);
+        int length = inputLength * forecastHorizon;
+        percentile = builder.percentile;
+        pastForecasts = new RangeVector[errorHorizon + forecastHorizon];
+        for (int i = 0; i < errorHorizon + forecastHorizon; i++) {
+            pastForecasts[i] = new RangeVector(length);
+        }
+        actuals = new float[errorHorizon + forecastHorizon][inputLength];
+        sequenceIndex = 0;
+        errorMean = new float[length];
+        errorRMSE = new DiVector(length);
+        multipliers = new DiVector(length);
+        Arrays.fill(multipliers.high, 1);
+        Arrays.fill(multipliers.low, 1);
+        intervalPrecision = new float[length];
+        errorDistribution = new RangeVector(length);
+        Arrays.fill(errorDistribution.upper, Float.MAX_VALUE);
+        Arrays.fill(errorDistribution.lower, Float.MIN_VALUE);
+    }
+
+    /**
+     * the folloqing would be useful when states and mappers get written
+     */
+    public ErrorHandler(int errorHorizon, int forecastHorizon, int sequenceIndex, double percentile, int inputLength,
+            int dimensions, float[] actualsFlattened, float[] pastForecastsFlattened, float[] auxilliary) {
+        checkArgument(forecastHorizon > 0, " incorrect forecast horizon");
+        checkArgument(errorHorizon >= forecastHorizon, "incorrect error horizon");
+        checkArgument(actualsFlattened != null || pastForecastsFlattened == null,
+                " actuals and forecasts are a mismatch");
+        checkArgument(inputLength > 0 && dimensions > 0 && dimensions % inputLength == 0, "incorrect parameters");
+        this.sequenceIndex = sequenceIndex;
+        this.errorHorizon = errorHorizon;
+        this.percentile = percentile;
+        this.forecastHorizon = forecastHorizon;
+        int currentLength = (actualsFlattened == null) ? 0 : actualsFlattened.length;
+        checkArgument(currentLength % inputLength == 0, "actuals array is incorrect");
+        int forecastLength = (pastForecastsFlattened == null) ? 0 : pastForecastsFlattened.length;
+        checkArgument(forecastLength == currentLength * dimensions * 3 / inputLength, "misaligned forecasts");
+        int arrayLength = max(forecastHorizon + errorHorizon, currentLength / inputLength);
+        this.pastForecasts = new RangeVector[arrayLength];
+        this.actuals = new float[arrayLength][inputLength];
+
+        int length = forecastHorizon * inputLength;
+
+        this.errorMean = new float[length];
+        this.errorRMSE = new DiVector(length);
+        this.intervalPrecision = new float[length];
+        this.multipliers = new DiVector(length);
+        this.errorDistribution = new RangeVector(length);
+
+        if (pastForecastsFlattened != null) {
+            for (int i = 0; i < currentLength / inputLength; i++) {
+                float[] values = Arrays.copyOfRange(pastForecastsFlattened, i * 3 * dimensions,
+                        (i * 3 + 1) * dimensions);
+                float[] upper = Arrays.copyOfRange(pastForecastsFlattened, (i * 3 + 1) * dimensions,
+                        (i * 3 + 2) * dimensions);
+                float[] lower = Arrays.copyOfRange(pastForecastsFlattened, (i * 3 + 3) * dimensions,
+                        (i * 3 + 3) * dimensions);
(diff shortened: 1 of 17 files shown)


Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/PredictorCorrector.java
```diff
@@ -195,8 +195,10 @@ protected boolean trigger(DiVector candidate, int gap, int baseDimension, DiVect
                 for (int i = dimensions - difference; i < dimensions; i++) {
                     remainder += candidate.getHighLowSum(i);
                 }
-                return thresholder.getAnomalyGrade(remainder * dimensions / difference, previousIsPotentialAnomaly,
-                        triggerFactor) > 0;
+                // simplifying the following since remainder * dimensions/difference corresponds
+                // to the
+                // impact of the new data since the last anomaly
+                return thresholder.getAnomalyGrade(remainder * dimensions / difference, previousIsPotentialAnomaly) > 0;
```
kaituo, line 201: Tested the change on AD. The precision does not change much after the change. Our benchmark does not have such small gap anomaly.

kaituo:
Comment on ConditionalSampleSummarizer.summarize:


Why do we add the largest distance * (1 - centrality) to threshold ?

```
        // note that the threshold is currently centrality * (some distance in the list)
        // thus the sequel uses a convex combination; and setting centrality = 0 removes
        // the entire filtering based on distances
        threshold += (1 - centrality) * newList.get(newList.size() - 1).distance;
        int num = 0;
        while (num < newList.size() && newList.get(num).distance <= threshold) {
            ++num;
        }

        ArrayList<Weighted<float[]>> typicalPoints = new ArrayList<>();
```

Is the computation of values necessary when j >= num? If no, can we change the loop condition from  j < newList.size() to j < num?

```
        for (int j = 0; j < newList.size(); j++) {
            ConditionalTreeSample e = newList.get(j);

            float[] values;
            if (project) {
                values = new float[missingDimensions.length];
                for (int i = 0; i < missingDimensions.length; i++) {
                    values[i] = e.leafPoint[missingDimensions[i]];
                }
            } else {
                values = Arrays.copyOf(e.leafPoint, dimensions);
            }
            if (j < num) { // weight is changed for clustering,
                // based on the distance of the sample from the query point
                double weight = (e.distance <= threshold) ? e.weight : e.weight * threshold / e.distance;
                typicalPoints.add(new Weighted<>(values, (float) weight));
            }
        }
```

kaituo:
Comment on NormalizedTransformer.invertForecastRange:

Why does NormalizedTransformer.invertForecastRange need to add i * deviations[j + inputLength].getMean()? (X+i) corresponds to the discounted average of the single step differences of the same variable i. I don't see you minus the same amount during normalization.

```
@Override
    public void invertForecastRange(RangeVector ranges, int baseDimension, double[] previousInput) {
        int inputLength = weights.length;
        int horizon = ranges.values.length / baseDimension;
        for (int i = 0; i < horizon; i++) {
            for (int j = 0; j < inputLength; j++) {
                double factor = 2 * (deviations[j].getDeviation() + DEFAULT_NORMALIZATION_PRECISION);
                ranges.scale(i * baseDimension + j, (float) factor);
                double shift = deviations[j].getMean() + i * deviations[j + inputLength].getMean();
                ranges.shift(i * baseDimension + j, (float) shift);
                float weight = (weights[j] == 0) ? 0 : 1.0f / (float) weights[j];
                ranges.scale(i * baseDimension + j, weight);
            }
        }
    }
```

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/Calibration.java
```diff
@@ -0,0 +1,36 @@
...
+     * precision will likely be close to 1 - 2 * percentile
+     */
+    MINIMAL,
+
+    /**
+     * a Markov inequality based interval, where the past error and model errors are
+     * additive. The interval precision is likely higher than MINIMAL but so are the
+     * interval.
+     */
+    SIMPLE;
```
kaituo, line 34: The difference between Calibration simple and minimal mode

```
void adjust(RangeVector rangeVector, RangeVector other) {
        checkArgument(other.values.length == rangeVector.values.length, " mismatch in lengths");
        for (int i = 0; i < rangeVector.values.length; i++) {
            rangeVector.values[i] += other.values[i];
            rangeVector.upper[i] = max(rangeVector.values[i], rangeVector.upper[i] + other.upper[i]);
            rangeVector.lower[i] = min(rangeVector.values[i], rangeVector.lower[i] + other.lower[i]);
        }
    }

    void adjustMinimal(RangeVector rangeVector, RangeVector other) {
        checkArgument(other.values.length == rangeVector.values.length, " mismatch in lengths");
        for (int i = 0; i < rangeVector.values.length; i++) {
            float oldVal = rangeVector.values[i];
            rangeVector.values[i] += other.values[i];
            rangeVector.upper[i] = max(rangeVector.values[i], oldVal + other.upper[i]);
            rangeVector.lower[i] = min(rangeVector.values[i], oldVal + other.lower[i]);
        }
    }
```

Why can simple mode  deal with stationary data (mean is constant) and minimum model can deal with data with distribution shift (e.g., mean changes)? From the above code, the only difference is how we compute upper and lower. In Simple mode, the upper and lower range adds new value while minimal adds old value.

kaituo:
haven't finished a full pass of the PR. Will continue.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public RCFCaster(Builder builder) {
+        super(builder);
+        checkArgument(errorHorizon >= 2 * forecastHorizon,
+                "intervalPrecision horizon should be at least twice as large as forecast horizon");
```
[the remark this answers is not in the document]
sudiptoguha: Yes, fixed.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/Calibration.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.amazon.randomcutforest.parkservices.calibration;
+
+public enum Calibration {
+
+    NONE,
+
+    /**
+     * a basic staring point where the itervals are adjusted to be the minimal
```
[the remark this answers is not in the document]
sudiptoguha: Thanks, fixed.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/PredictorCorrector.java
sudiptoguha: Thanks. We now have a small gap test in the RCF library, so it can get tracked.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/SequentialAnalysis.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public static List<AnomalyDescriptor> detectAnomalies(double[][] data, int shingleSize, double timeDecay,
+            TransformMethod transformMethod, double transformDecay) {
+        return detectAnomalies(data, shingleSize, DEFAULT_SAMPLE_SIZE, DEFAULT_NUMBER_OF_TREES, timeDecay,
+                DEFAULT_SAMPLE_SIZE / 4, transformMethod, transformDecay, new Random().nextLong());
+    }
+
+    /**
+     * Same as the anomaly detector but provides a
```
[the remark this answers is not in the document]
sudiptoguha: Thanks, fixed.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ErrorHandler.java
```diff
@@ -0,0 +1,372 @@
...
+     * 
+     * @param percentile the desired percentile (we recomment leaving this at 0.1 --
+     *                   the algorithm likely will never have sufficiently many
+     *                   observations to have a very fine grain distribution;
+     *                   moreover any forecasting should concentrate the measure (no
+     *                   pun intended) towards 0.5 and for a successful forecaster
+     *                   these error percentiles will be less meaningful
+     * @param newHorizon a horizon which could be smaller than the errorHorizon (for
+     *                   auditing)
+     * @param error      the function that defies error
```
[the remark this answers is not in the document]
sudiptoguha: Thanks, fixed.

sudiptoguha:
Re comment on Comment on NormalizedTransformer.invertForecastRange,

take a look at the superclass WeightedTransformer and updateDeviations;
    public void updateDeviation(double[] inputPoint, double[] previousInput) {
        checkArgument(inputPoint.length * 2 == deviations.length, "incorrect lengths");
        for (int i = 0; i < inputPoint.length; i++) {
            deviations[i].update(inputPoint[i]);
            if (deviations[i + inputPoint.length].getCount() == 0) {
                deviations[i + inputPoint.length].update(0);
            } else {
                deviations[i + inputPoint.length].update(inputPoint[i] - previousInput[i]);
            }
        }
    }

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/Calibration.java
sudiptoguha: The simple mode adds the error intervals (uncertainty from the model and the actual error distribution). The minimal mode fixes the mean,  but uses the error of "other" (which is the error distribution) and ignores the ranges produced by the model. The choice depends on the assumption on the data, the first is a pessimistic choice -- the second is a minimal choice (just adjusting based on the previous errors).  Using simple may be simpler, but will give wider intervals. There can be many more calibration steps based on assumptions one is willing to make on the input data.

sudiptoguha:
Re Comment on ConditionalSampleSummarizer.summarize:

we add the ( 1- centrality) * largest distance because the threshold (after the addition) becomes

centrality * distance[j] + (1-centrality)*largest distance

So setting centrality = 1 would pick a lower threshold and perform filtering. Setting centrality = 0 would basically choose every point returned by all the trees. The difference between the num and newList.size() is for the clustering method (used for the most likely value) -- for clustering, the filtering is a useful denoting step. But otherwise we provide the p50 value of every  value returned from the trees in the forecast (this is the j>num part).

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -109,7 +109,7 @@ public ThresholdedRandomCutForest(Builder<?> builder) {
         forest = builder.buildForest();
         preprocessorBuilder.weights(builder.weights);
         preprocessorBuilder.weightTime(builder.weightTime.orElse(1.0));
-        preprocessorBuilder.timeDecay(forest.getTimeDecay());
+        preprocessorBuilder.timeDecay(builder.transformDecay.orElse(forest.getTimeDecay()));
```
kaituo, line 112: What is transform decay? How is it different from time decay?

kaituo:
Finished one pass. Will discuss the followup questions offline.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
sudiptoguha: TransformDecay is the exponential moving average during normalization -- this value is subtracted from the (corresponding element of) the input value and divided by the corresponding deviation (with the same exponential weights). This corresponds to the 50-value moving average or 100-value moving average computations; the transform decays would be 0.02 or 0.01. 

TimeDecay (existed earlier in the library) corresponds to the half-life of detain the weighted online sampling. 

It may make sense to use the same number -- but it is a flexibility to be able to use two different numbers. Anomaly detection applications are not particularly sensitive to the transformDecay and hence it did not have to exist earlier. Many folks believe that the moving average computation is of interest to forecasting, specially long term trends. So making that available as an option makes more sense. The example RCFCasterExample.lava shows how to use it (and has some comments). It makes some difference in how quickly does forecasts adapt to changes -- but does not have a strong impact for stationary sequences.

sudiptoguha:
Re "Is the computation of values necessary when j >= num? If no, can we change the loop condition from j < newList.size() to j < num?" --> no, the code can be simplified as suggested. Changing.

kaituo (approved):

jotok (approved):

PR #355, opened 2022-11-13
Remove output from the `create_rcf` function

ankane:
*Issue #, if available:* N/A

*Description of changes:* Removes output from the `create_rcf` function to reduce noise for users (alternatively, could add a `verbose` parameter, but it'd be a breaking change).


sudiptoguha (approved):
Thanks - there were holdovers from a very long time ago!

ankane:
Thanks!

PR #356, opened 2022-11-13
Reduced warnings with Rust library

ankane:
*Issue #, if available:* N/A

*Description of changes:* Reduced warnings from 98 to 77 with `rcf` lib and 1 to 0 with `example` bin (with rustc 1.64.0).

### DIFF
--- a/Rust/src/trcf/predictorcorrector.rs
+++ b/Rust/src/trcf/predictorcorrector.rs
@@ -149,7 +149,7 @@ impl PredictorCorrector {
                 }
                 let internal_timestamp = result.internal_timestamp;
                 let base_dimension = forest.dimensions() / shingle_size;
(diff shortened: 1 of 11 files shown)


Rust/src/trcf/predictorcorrector.rs
```diff
@@ -149,7 +149,7 @@ impl PredictorCorrector {
                 }
                 let internal_timestamp = result.internal_timestamp;
                 let base_dimension = forest.dimensions() / shingle_size;
-                let mut start_position = (shingle_size - 1) * base_dimension;
+                let start_position = (shingle_size - 1) * base_dimension;
```
sudiptoguha, line 152: This whole line should removed -- it seems start_position has been redefined in all of the branches. But removing the mut makes sense.

sudiptoguha (approved):

PR #357, opened 2022-11-19
Fixed more warnings with Rust library

ankane:
*Issue #, if available:* N/A

*Description of changes:* Reduced `cargo build` warnings from 77 to 68 and fixed all the `cargo test` warnings. The remaining warnings are all for unused variables or functions.

### DIFF
--- a/Rust/src/common/cluster.rs
+++ b/Rust/src/common/cluster.rs
@@ -39,7 +39,7 @@ pub trait Cluster<Q, T: ?Sized> {
     fn representatives(&self) -> Vec<(Q, f32)>;
     // a function to distill all representatives into a single representative under a given
     // distance function over the base type of the points
-    fn primary_representative(&self, distance: fn(&T, &T) -> f64) -> Q;
+    fn primary_representative(&self, _distance: fn(&T, &T) -> f64) -> Q;
 }
 
 #[repr(C)]
@@ -189,7 +189,7 @@ impl Cluster<Vec<f32>, [f32]> for Center {
         vec![(self.representative.clone(), self.weight as f32); 1]
     }
 
-    fn primary_representative(&self, distance: fn(&[f32], &[f32]) -> f64) -> Vec<f32> {
+    fn primary_representative(&self, _distance: fn(&[f32], &[f32]) -> f64) -> Vec<f32> {
         self.representative.clone()
     }
 }
--- a/Rust/src/common/rangevector.rs
+++ b/Rust/src/common/rangevector.rs
@@ -43,32 +43,28 @@ impl RangeVector {
     }
 
     pub fn shift(&mut self, i:usize, shift: f32) {
-        for _ in 0..self.values.len() {
-            self.values[i] += shift;
-            self.upper[i] += shift;
-            self.lower[i] += shift;
-            // managing precision explicitly
-            if self.upper[i] < self.values[i] {
-                self.upper[i] = self.values[i];
-            }
-            if self.lower[i] > self.values[i] {
-                self.lower[i] = self.values[i];
-            }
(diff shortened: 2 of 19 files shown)


Rust/src/trcf/predictorcorrector.rs
```diff
@@ -149,7 +149,6 @@ impl PredictorCorrector {
                 }
                 let internal_timestamp = result.internal_timestamp;
                 let base_dimension = forest.dimensions() / shingle_size;
-                let start_position = (shingle_size - 1) * base_dimension;
```
sudiptoguha, line 152: Thanks!

Rust/src/common/rangevector.rs
```diff
@@ -43,7 +43,7 @@ impl RangeVector {
     }
 
     pub fn shift(&mut self, i:usize, shift: f32) {
-        for j in 0..self.values.len() {
+        for _ in 0..self.values.len() {
```
sudiptoguha, line 46: Yikes, this is a bug! The intention was to change the single coordinate once. This loop should be removed. We can have a separate PR or fix in this one.

Rust/src/common/rangevector.rs
```diff
@@ -58,7 +58,7 @@ impl RangeVector {
     }
 
     pub fn scale(&mut self, i:usize, scale: f32) {
-        for j in 0..self.values.len() {
+        for _ in 0..self.values.len() {
```
sudiptoguha, line 61: Loop should be removed, we can have a separate PR or use this one.

Rust/src/common/cluster.rs
```diff
@@ -39,7 +39,7 @@ pub trait Cluster<Q, T: ?Sized> {
     fn representatives(&self) -> Vec<(Q, f32)>;
     // a function to distill all representatives into a single representative under a given
     // distance function over the base type of the points
-    fn primary_representative(&self, distance: fn(&T, &T) -> f64) -> Q;
+    fn primary_representative(&self, _distance: fn(&T, &T) -> f64) -> Q;
```
sudiptoguha, line 42: The primary_representative  is needed when we move to multi-centroid clustering (CURE, as in the Java version). But this change makes sense.

Rust/src/common/rangevector.rs
ankane: Great, fixed this and `scale` in a follow-up commit.

Rust/tests/samplesummarytest.rs
```diff
@@ -25,8 +24,6 @@ fn core(
     seed: u64,
     distance: fn(&[f32], &[f32]) -> f64,
 ) -> bool {
-    let mut rng = ChaCha20Rng::seed_from_u64(seed);
-    let data_size = 200000;
```
ankane, line 29: This changes the data size to `1000000` for the parameterized test, but we can update it there if that's not desired.

Rust/src/common/cluster.rs
ankane: Was guessing many of the unused variable / function warnings were due to not yet implemented features, so was not planning to fix any others (to avoid interfering with that). I can undo this one if you want.

Rust/tests/samplesummarytest.rs
sudiptoguha: Should be fine with the change!

Rust/src/common/cluster.rs
sudiptoguha: The change is fine as is -- should change it when the feature is updated.

sudiptoguha (approved):

ankane:
Thanks @sudiptoguha!

PR #367, opened 2023-02-22
fixing gpg signing

amitgalitz:
*Description of changes:*
Fixing gpg signing as the official release script failed for 2.5 because it didn't recognize the mvn deploy command. Tried this fix locally with mvn install but no way to officially test it all the way as it would require actually releasing to staging first. took inspiration from change that was made in this PR, https://github.com/acm19/aws-request-signing-apache-interceptor/pull/44/files


sudiptoguha:
Thanks!

PR #370, opened 2023-03-01
adding release notes workflow

amitgalitz:
*Description of changes:*

This PR adds a workflow to easily add release notes from any PRs after this (same as OpenSearch Org uses). If we want to correspond a PR to a specific type of change (infra, bug fix, enhancement and etc.) we can simply add a label to the PR. These can always be easily edited in the releases section of github as well. Also making a small change to server id for maven release as we use the staging plugin in our pom file which caused the issues with auto release that could only be caught in official releases. 

### DIFF
--- a/Java/RELEASING.md
+++ b/Java/RELEASING.md
@@ -1,0 +1,24 @@
+- [Overview](#overview)
+- [Feature Branches](#feature-branches)
+- [Release Labels](#release-labels)
+- [Releasing](#releasing)
+
+## Overview
+
+This document explains the release strategy for the Random Cut Forest project.
+
+## Feature Branches
+
+Do not create branches in the upstream repo, use your fork, for the exception of long lasting feature branches that require active collaboration from multiple developers. Name feature branches `feature/<thing>`. Once the work is merged to `main`, please make sure to delete the feature branch.
+
+## Release Labels
+
+Repositories create consistent release labels, such as `3.0.0-java`. Use release labels to target an issue or a PR for a given release.
+
+## Releasing
+
+The release process is run by a release manager volunteering from amongst the maintainers.
+
+1. Create a PR to bump version to desired release candidate (e.g. 3.0.0)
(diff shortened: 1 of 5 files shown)


dblock:
You might want to add something like https://github.com/opensearch-project/opensearch-java/blob/main/RELEASING.md that explains how to make a release. Also take a look at the implementation in that repo and maybe in https://github.com/acm19/aws-request-signing-apache-interceptor, both are fully hands off now, single click release.

amitgalitz:
> You might want to add something like https://github.com/opensearch-project/opensearch-java/blob/main/RELEASING.md that explains how to make a release. Also take a look at the implementation in that repo and maybe in https://github.com/acm19/aws-request-signing-apache-interceptor, both are fully hands off now, single click release.

Added `RELEASING.md` with latest commit. I also took a look at the implementation of the repos you mentioned. It is not as easy to swap for the `opensearch-java` way as it uses some gradle methods but I can open an issue on the effort it takes to eventually move to gradle here. It also uses the jenkins infra and workflow that this repo doesn't have. For the other request-signing repo, we could definitely move to something closer to that in the future but will take a little more effort and also I know in OpenSearch we were against any Github actions auto merging changes but we can do it so a PR is opened. I propose to first release RCF this way successfully and we can re-evaluate improvements here if that makes sense.

jotok (approved):

sudiptoguha (approved):

Java/RELEASING.md
```diff
@@ -0,0 +1,24 @@
...
+
+Repositories create consistent release labels, such as `3.0.0-java`. Use release labels to target an issue or a PR for a given release.
+
+## Releasing
+
+The release process is run by a release manager volunteering from amongst the maintainers.
+
+1. Create a PR to bump version to desired release candidate (e.g. 3.0.0)
+2. Click run on the maven-release workflow in Github Actions which uploads the artifacts to our staging repository, creates a new tag, and a new Github release
+3. Login into the nexus staging repositor , verify artifact was signed succesfully and click release to officially push to maven central.
```
dblock, line 24: typo: repository

PR #401, opened 2023-11-03
RCF4.0 and PredictiveRCF

sudiptoguha:
*Description of changes:* This PR initiates RCF 4.0. The primary change is the realization that while RCF has been built in layers over time -- some of the streaming normalization is standard yet extremely useful. This preprocessing functionality existed in ParkServices and yet it is increasingly clear that using RCFs without these normalizations are not really helpful. Thus the entire preprocessing is now shifted to core rationalizing the configs with the code. As an example benefit, we introduce a PredictiveRCF that update on vectors over attribute dimensions A and B, and given values of the dimension A provides a clustering over candidate values in dimensions B. This capability existed in the imputeMissingValues() -- but the addition of of the preprocessing (and the inverse map) alongside exposure of the clustering would likely be useful. As a consequence we can use this predictor to estimate the errors of forecasting in RCFCast, reducing the amount of state required for calibration of the output.

In addition newer tests have been added and the coverage of ParkServices is significantly higher.


kaituo:
For TimedRangeVector, package name 'com.amazon.randomcutforest.parkservices.returntypes' does not correspond to the file path 'com.amazon.randomcutforest.returntypes'.

kaituo:
Should you change the following code 

```
public GenericAnomalyDescriptor(List<Weighted<P>> representative, double score, double threshold,
            double anomalyGrade) {
        this.representativeList = representativeList;
```

to

```
public GenericAnomalyDescriptor(List<Weighted<P>> representative, double score, double threshold,
            double anomalyGrade) {
        this.representativeList = representative;
```

?

kaituo:
In the constructor of ImputeVisitor, are we missing four fields including box, converged, pointIndex, and randomRank?

```
ImputeVisitor(ImputeVisitor original) {
        int length = original.queryPoint.length;
        this.queryPoint = Arrays.copyOf(original.queryPoint, length);
        this.missing = Arrays.copyOf(original.missing, length);
        this.dimensionsUsed = new int[original.dimensionsUsed.length];
        this.randomSeed = new Random(original.randomSeed).nextLong();
        this.centrality = original.centrality;
        anomalyRank = DEFAULT_INIT_VALUE;
        distance = DEFAULT_INIT_VALUE;
    }
```

kaituo (commented):
read until Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/Preprocessor.java

Java/examples/src/main/java/com/amazon/randomcutforest/examples/dynamicinference/ConditionalPredictive.java
```diff
@@ -0,0 +1,172 @@
...
+        Random random = new Random(seed + 10);
+
+        String name = "predictive_example";
+        BufferedWriter file = new BufferedWriter(new FileWriter(name));
+        for (int i = 0; i < dataSize; i++) {
+            float[] record = generateRecordKey(random);
+            checkArgument(record[3] == 0, " should not be filled");
+            checkArgument(record[4] == 0, " should not be filled");
+
+            SampleSummary answer = forest.predict(record, 0, new int[] { 3, 4 });
```
kaituo, line 77: Can we think predict multiple missing values multivariate forecasting? You told me multivariate forecasting is complicated and rcf hasn't supported it. If yes, how can we predict multiple missing values confidently?

I am thinking it aloud for the potential integration on AD plugin as follows.  

In the context of streaming supervised learning, this can be used for multi-output regression (e.g., When the outputs or targets are continuous values. For example, predicting the temperature and humidity levels for the next day based on today's weather data.) or multi-label classification (e.g., When each instance can belong to multiple classes. For instance, a text document might be about both 'science' and 'technology', which are not mutually exclusive categories.), right?

We can use a multi-target regression or classification rcf model depending on the nature of the KPIs. For example, a multi-output regression model might predict both the latency and error rate simultaneously. The model's input features would include the change indicators. When the model detects a pattern that historically correlates with certain types of system degradations, it can make predictions about the current state. When increased latency and error rates are observed, rcf can evaluate the likelihood of these being related to recent changes such as a specific deployment or a configuration change and we can find root cause using rcf's attribution.

Of course, rcf can be used to output not only the predicted KPIs but also the likelihood of various potential root causes. This can be done by structuring the output layer to have nodes corresponding to various root causes, trained to activate when the input data matches historical patterns leading to those causes. For instance, rcf might indicate that there is an 80% likelihood that the current latency issue is due to a recent deployment and a 50% likelihood that it's related to a configuration change.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/parkservices/ForecastWithLimits.java
```diff
@@ -0,0 +1,187 @@
...
+import com.amazon.randomcutforest.parkservices.config.Calibration;
+import com.amazon.randomcutforest.returntypes.DiVector;
+import com.amazon.randomcutforest.returntypes.RangeVector;
+import com.amazon.randomcutforest.testutils.MultiDimDataWithKey;
+import com.amazon.randomcutforest.testutils.ShingledMultiDimDataWithKeys;
+
+/**
+ * The following example demonstrates the self calibration of RCFCast. Change
+ * various parameters -- we recommend keeping baseDimension = 1 (for single
+ * variate forecasting -- multivariate forecasting can be a complicated
```
kaituo, line 38: In RCF space, we are dealing with shingled input, right? baseDimension =1 and shingle size = 8 is the same thing for RCF as baseDimension = 8 and shingle size 1, right? Of course, with shingle size = 8, we are looking at past 8 in the history and our forecast would be more accurate. But other than that, there is no fundamental difference between baseDimension =1,shingle size = 8 and baseDimension = 8, shingle size =1, right?

kaituo (commented):
read until Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java

sudiptoguha:
> For TimedRangeVector, package name 'com.amazon.randomcutforest.parkservices.returntypes' does not correspond to the file path 'com.amazon.randomcutforest.returntypes'.

fixed. Thx.

sudiptoguha:
> Should we change the following code
> 
> ```
> public GenericAnomalyDescriptor(List<Weighted<P>> representative, double score, double threshold,
>             double anomalyGrade) {
>         this.representativeList = representativeList;
> ```
> 
> to
> 
> ```
> public GenericAnomalyDescriptor(List<Weighted<P>> representative, double score, double threshold,
>             double anomalyGrade) {
>         this.representativeList = representative;
> ```
> 
> ?

Fixed. Thx.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/dynamicinference/ConditionalPredictive.java
```diff
@@ -0,0 +1,172 @@
...
+            fillInValues(record, random, normal);
+            forest.update(record, 0);
+            double tag = Double.MAX_VALUE;
+            for (int y = 0; y < answer.summaryPoints.length; y++) {
+                double t = Summarizer.L2distance(record, answer.summaryPoints[y]);
+                tag = min(tag, t);
```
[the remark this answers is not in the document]
sudiptoguha: Yes :) But alternate experiments used different versions of the tag -- for example, we may be interested in exp( - const * distance) as a tag, similar to a kernel.

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
```diff
@@ -0,0 +1,377 @@
...
+                checkArgument(missingValues.length <= inputPoint.length, " incorrect data");
+                for (int i = 0; i < missingValues.length; i++) {
+                    checkArgument(missingValues[i] >= 0, " missing values cannot be at negative position");
+                    checkArgument(missingValues[i] < inputPoint.length,
+                            "missing values cannot be at position larger than input length");
+                }
+            }
+            float[] scaledInput = preprocessor.getScaledInput(inputPoint, timestamp);
+            summary = preprocessor.invertSummary(
+                    forest.getConditionalFieldSummary(scaledInput, missingValues.length, missingValues,
```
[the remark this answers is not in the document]
sudiptoguha: yes.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/dynamicinference/ConditionalPredictive.java
sudiptoguha: > Can we think predict multiple missing values multivariate forecasting? You told me multivariate forecasting is  > complicated and rcf hasn't supported it. If yes, how can we predict multiple missing values confidently?

Actually RCF never disallowed it :) But I would always recommend using it with caution. Multivariate forecasting is slightly different from multi target regression. The issue is that in simple algorithms for Multivariate forecasting the multiple targets are then used to generate subsequent predictions -- and that can be a self reinforcing loop. 
For one dimension this can mostly be controlled -- but with multiple dimensions it becomes near infeasible.  More complicated algorithms also have the same problem in varying degrees (c.f. hallucinations). But multi-target regression in itself is fine -- in fact RCF can perform tree regression (just like RFs) and streaming comes for free. So a forecast horizon of 1 is not a big issue at all. 

> I am thinking it aloud for the potential integration on AD plugin as follows.

We should try it out.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/parkservices/ForecastWithLimits.java
sudiptoguha: The issue is for forecast horizon = 5, the first would produce three vectors of size 5 (one value for each lookahead step, with upper and lower bounds). The second would be difficult to interpret because shingle size = 1 -- so there is no time-series aspect. For base dimension = 4 and shinglesize = 2, a forecast horizon of 5 would produce three vectors of length 4 *5 = 20.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
```diff
@@ -0,0 +1,492 @@
...
+                    query[inputLength] = i;
+                    for (int j = 0; j < inputLength; j++) {
+                        missing[j] = inputLength + 1 + j;
+                    }
+                    // at this moment we use the PredictiveRCF more for the shorter term estimation,
+                    // and use an interpolation
+                    // with the observed error for the longer term
+                    SampleSummary answer = estimator.predict(query, 0, missing, 1, 0.5, 1.0, false);
+                    for (int j = 0; j < inputLength; j++) {
+                        errorHigh[i * inputLength + j] = (forecastHorizon - j)
```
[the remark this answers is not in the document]
sudiptoguha: Correct -- it does not make much sense. It's a bug. Fixed. Also this specific estimator seem to be improved when the last value is added.

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/SampleSummary.java
```diff
@@ -96,33 +101,56 @@ public SampleSummary(double[] point, float weight) {
...
-        this.addTypical(summaryPoints, relativeWeight);
+        float[][] transformedMeasure = new float[measure.length][];
+        double factor = 0;
+        for (int j = 0; j < this.upper.length; j++) {
+            factor += max(0, this.upper[j] - this.lower[j]);
+        }
+        for (int i = 0; i < measure.length; i++) {
+            transformedMeasure[i] = new float[this.upper.length];
+            for (int j = 0; j < this.upper.length; j++) {
+                transformedMeasure[i][j] = (factor == 0) ? 0
```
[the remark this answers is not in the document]
sudiptoguha: It's not used yet -- but it can be very useful. This indicates the "shape" of the box around the representative point -- the relative weight captures the mass, and this can capture the extent.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/GlobalLocalAnomalyDetector.java
```diff
@@ -314,8 +324,8 @@ public List<Weighted<P>> score(P current, BiFunction<P, P, Double> localDistance
                     for (int j = index + 1; j < candidateList.size(); j++) {
                         double occludeDistance = local.apply(head.representative, candidateList.get(j).representative);
                         double candidateDistance = candidateList.get(j).distance;
-                        if (occludeDistance < candidateDistance && head.distance > Math
-                                .sqrt(head.distance * head.distance + candidateDistance * candidateDistance)) {
+                        if (occludeDistance < candidateDistance && candidateDistance > Math
+                                .sqrt(head.distance * head.distance + occludeDistance * occludeDistance)) {
                             // delete element
                             candidateList.remove(j);
```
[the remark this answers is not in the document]
sudiptoguha: Fixed. This is intended to be single threaded anyways - a while loop should suffice.

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
```diff
@@ -0,0 +1,377 @@
...
+            }
+        }
+        return summary;
+    }
+
+    public void update(float[] record, long timestamp) {
+        float[] scaled = preprocessor.getScaledInput(record, timestamp);
+        preprocessor.update(toDoubleArray(record), scaled, timestamp, null, forest);
+        if (scaled != null) {
+            forest.update(scaled);
```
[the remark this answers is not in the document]
sudiptoguha: No - the preprocessor.update would already address the update. We do not need to repeat the update (it'll effectively approximately halve the sample size). Fixed. ImputeProcessor updates the forest via generateShingle() -- because it needs to regenerate the shingle (which itself may have missing values).

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/Preprocessor.java
```diff
@@ -613,35 +496,69 @@ protected double[] getExpectedValue(int relativeBlockIndex, double[] reference,
...
+
+    public SampleSummary invertSummary(SampleSummary summary, int numberOfMissing, int[] missingIndices,
+            float[] point) {
+        if (summary == null) {
+            return null;
+        }
+        SampleSummary newSummary = new SampleSummary(point.length);
+        newSummary.relativeWeight = Arrays.copyOf(summary.relativeWeight, summary.relativeWeight.length);
+        double[] scale = getScale();
+        if (summary.mean.length == point.length) {
```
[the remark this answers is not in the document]
sudiptoguha: If the "project" option is set to true -- then the summary would be constructed over the projected space. It is an open question if the clustering should be over the original space or the space of the missing values ...

Java/core/src/main/java/com/amazon/randomcutforest/returntypes/SampleSummary.java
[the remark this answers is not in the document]
kaituo: Can you explain "the relative weight captures the mass, and this can capture the extent"? What's mass and what's extent? How is it related to transformedMeasure?

I thought for internal node, mass is the number of leave node. For leaf node, mass is the leaf node's duplicity.

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
```diff
@@ -0,0 +1,377 @@
...
+import com.amazon.randomcutforest.preprocessor.Preprocessor;
+import com.amazon.randomcutforest.returntypes.SampleSummary;
+
+/**
+ * This class provides a predictive imputation based on RCF (respecting the
+ * arrow of time) alongside streaming normalization
+ */
+@Getter
+@Setter
+public class PredictiveRandomCutForest {
```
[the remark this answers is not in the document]
kaituo: You might have missed my comment.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -127,8 +128,7 @@ public ThresholdedRandomCutForest(Builder<?> builder) {
         // to be used later
         preprocessorBuilder.randomSeed(builder.randomSeed.orElse(0L) + 1);
         preprocessorBuilder.dimensions(builder.dimensions);
-        preprocessorBuilder
-                .stopNormalization(builder.stopNormalization.orElse(Preprocessor.DEFAULT_STOP_NORMALIZATION));
+        preprocessorBuilder.stopNormalization(builder.stopNormalization.orElse(DEFAULT_STOP_NORMALIZATION));
```
kaituo, line 131: When do we need to use stopNormalization?

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -127,8 +128,7 @@ public ThresholdedRandomCutForest(Builder<?> builder) {
         // to be used later
         preprocessorBuilder.randomSeed(builder.randomSeed.orElse(0L) + 1);
         preprocessorBuilder.dimensions(builder.dimensions);
```
kaituo, line 130: Question on line 127: 

preprocessorBuilder.timeDecay(builder.transformDecay.orElse(1.0 / builder.sampleSize));

Does that mean timeDecay cannot be set by user as we only read transformDecay?

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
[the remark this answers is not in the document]
sudiptoguha: No :) Stay tuned. The vision is PredictiveRCF to be a combination of Preprocessor + RandomCutForest. The current RandomCutForest class is backward compatible with RCF 1.0 (state only), 2.0, 3.0 ... changing that class to accommodate preprocessing seems daunting at the moment (may be feasible down the road), and we do not want to break the backward compatibility. PredictiveRCF would focus on the higher level and the existing RandomcutForest will focus on the more detail focused level. In fullness of time there can be convergence -- note that this is how we went from RCF1.0 (pointer trees) to RCF2.0 (compact RCF) by slowly changing in place. This would hopefully avoid any regression/unwanted/unexplained changes.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/dynamicinference/ConditionalPredictive.java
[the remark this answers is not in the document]
sudiptoguha: Stay tuned. Next commit ..

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
sudiptoguha: Let me change the name as well -- Preprocessor was using names from RCF2.0 era. Given that we are moving that, this may be a good time to change the name (state classes remain unchanged for backward compatibility).

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/GlobalLocalAnomalyDetector.java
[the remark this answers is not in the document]
sudiptoguha: pushed - take a look.

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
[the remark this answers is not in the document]
sudiptoguha: Take a look at ErrorHandler -- it can now use simple statistics or a more involved predictive model (PredictiveRCF) where we predict the last two columns of (current value, lookahead, _positive_error, _negative_error). Overall the estimation is better for shorter lookaheads -- but does take more space and time. It is an option and can be built upon. Also in the fullness of time ThresholdedRCF can use PredicitiveRCF directly and avoid messing with transformations. Right now ParkServices correspond to time series analysis (thresholding corresponds to a sequence of values, same for forecasts) and do not have flexibility of updates (because it is difficult to turn back on) whereas the core packages the flexibility of updates.

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/Preprocessor.java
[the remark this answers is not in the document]
sudiptoguha: Refactored to just consider the last block.

Java/examples/src/main/java/com/amazon/randomcutforest/examples/dynamicinference/ConditionalPredictive.java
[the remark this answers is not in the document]
sudiptoguha: As above - take a look at ErrorHandler

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
sudiptoguha: Done

kaituo:
read until Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java of commit https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/c8721d4a2b5f7c513153fe1d9279ac68ff546c9d

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
```diff
@@ -63,17 +77,28 @@ public class PredictiveRandomCutForest {
...
         if (builder.forestMode == ForestMode.TIME_AUGMENTED) {
             dimensions += builder.shingleSize;
+            // if time is not differenced, then it can be added as a column
+            // without much difficulty
             preprocessorBuilder.normalizeTime(true);
             // force internal shingling for this option
             builder.internalShinglingEnabled = Optional.of(true);
+        } else if (builder.forestMode == ForestMode.STREAMING_IMPUTE) {
+            preprocessorBuilder.normalizeTime(true);
+            builder.internalShinglingEnabled = Optional.of(false);
```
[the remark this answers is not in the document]
sudiptoguha: No - the invoking function will send inputLength size vectors. The shingling will be handled by the preprocessor but the RCF used internally will be externally shingled. The reason is that we/user may say "if we have shingle size X and have made up half the data then do not add the made up data to RCF' -- this is the imputedShingleFraction setting. This can be very helpful in stopping runaway feedback loops (hallucinations) since if the forecasted values are fed into the model we have a classic reinforcement scenario. This is the current state of ThresholdedRandomCutForest btw. One of the goals of this major refactor to make these options/capabities explicit.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -996,7 +996,8 @@ public List<ConditionalTreeSample> getConditionalField(float[] point, int number
     }
```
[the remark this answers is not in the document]
sudiptoguha: Clearly not! Let me look.

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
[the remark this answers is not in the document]
kaituo: Not sure I understand. Will ask offline.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
```diff
@@ -1152,7 +1154,7 @@ void extrapolateBasicSliding(RangeVector result, int horizon, int blockSize, flo
             System.arraycopy(queryPoint, blockSize, queryPoint, 0, dimensions - blockSize);
 
             SampleSummary imputedSummary = getConditionalFieldSummary(queryPoint, blockSize, missingIndexes, 1, 0,
-                    false, false, centrality);
+                    false, false, centrality, 1);
```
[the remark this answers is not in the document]
sudiptoguha: Lots of great discussion points. First, what is intended -- the shingleSize 1 means we will compute statistics for the entire vector and output that. ShingleSize 2 means that we would only compute the statistic of the second (last) half of the vector. Likewise 3, 4 ... etc. Keep in mind for forecasting the shingling size is large(ish, specially if the horizon is large) so doing this would save a lot of time (and indeed the current PR seems to save 60%+ time for the build for parkservices :)

Now for the ideas/point you raised -- we actually do some filtering based on the centrality parameter -- centrality = 1 will throw away the far points. Centrality = 0 will not perform any filtering. *Now* we could be using the using the distance (lets see harmonic mean/something) as weights -- that would be reasonable and better in some scenarios. But I have not managed to experiment much -- in the small examples I saw, I found that changing the weights was introducing variance/non-smoothness in the result. But I am sure it can be controlled. One other idea worth using is to use (inverse of) RCF scores themselves as weight. Same issue - the results are better in some cases, but then needs more controlling and experiments.

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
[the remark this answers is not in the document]
kaituo: yes, feel free to do it in the next PRs.

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
```diff
@@ -381,54 +242,59 @@ protected int determineGap(long timestampGap, double averageGap) {
...
-        // previous value should be defined
-        double[] previous = new double[inputLength];
-        System.arraycopy(lastShingledInput, lastShingledInput.length - inputLength, previous, 0, inputLength);
+        double[] input = Arrays.copyOf(inputTuple, inputLength);
+        double[] previous = getShingledInput(shingleSize - 1);
+
+        if (missingValues != null && imputationMethod != RCF) {
+            // for STREAMING_IMPUTE the timestamp cannot be missing
+            // hence missingValues[] can be 0 to inputLength - 1
+            // for next and Linear there are no current values
```
kaituo, line 271: For linear, can we use previous n value at dimension i to do linear interpolation?

When will next imputation be useful? I can only think of historical AD as next value is available.

Current approach defaults to fixedvalues or previous seems to silently make a decision for the customer. Should we fail fast and return empty shingle since current imputation strategy does not work?

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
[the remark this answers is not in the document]
sudiptoguha: Done. It wasn't as much -- we just edited those same lines for pauseSampling() and resumeSampling() for basic RCF (not TRCF -- because there is no good logic/decision mechanism to restart sampling for continuously evolving data once sampling is stopped). So the core of PointStore and the representations did not have to be touched. Of course we now have some redundant and deprecated function stubs -- those can be focus of subsequent PRs.

Java/core/src/main/java/com/amazon/randomcutforest/RandomCutForest.java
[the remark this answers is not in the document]
sudiptoguha: Fixed.

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
kaituo: I saw you imputed 3 times:

First, impute current input value:
```
if (missingValues != null && imputationMethod != RCF) {
            // for STREAMING_IMPUTE the timestamp cannot be missing
            // hence missingValues[] can be 0 to inputLength - 1
            // for next and Linear there are no current values
            // we are forced to use fixedvalues or previous
            for (int missingValue : missingValues) {
                input[missingValue] = (defaultFill == null) ? previous[missingValue] : defaultFill[missingValue];
            }
        }
```

Then impute the missing gaps between last seen value and the value immediately before current input:

```
double[] savedInput = Arrays.copyOf(previous, inputLength);
        // using the global dependency
        int numberToImpute = determineGap(timestamp - lastInputTimeStamp, averageGap) - 1;
        if (numberToImpute > 0) {
            double step = 1.0 / (numberToImpute + 1);
            // the last impute corresponds to the current observed value
            for (int i = 0; i < numberToImpute; i++) {
                // only the last tuple is partial
                double[] result = impute(input, missingValues, false, step * (i + 1), previous, forest);
                updateForest(changeForest, result, timestamp, forest, true);
            }
        }
```

Then you impute current input value again:

```
 boolean ignoreImpute = (missingValues == null) || (missingValues.length == 0);

        double[] newInput = (ignoreImpute) ? input : impute(input, missingValues, true, 0, previous, forest);
 ```
 
 Is my understanding correct? If yes, why do we need to impute current input again?

File: Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java [hunk 17]
```diff
@@ -381,54 +257,59 @@ protected int determineGap(long timestampGap, double averageGap) {
...
         // using the global dependency
         int numberToImpute = determineGap(timestamp - lastInputTimeStamp, averageGap) - 1;
         if (numberToImpute > 0) {
-            descriptor.setNumberOfNewImputes(numberToImpute);
             double step = 1.0 / (numberToImpute + 1);
             // the last impute corresponds to the current observed value
             for (int i = 0; i < numberToImpute; i++) {
-                double[] result = impute(false, descriptor, step * (i + 1), previous, forest);
+                // only the last tuple is partial
+                double[] result = impute(input, missingValues, false, step * (i + 1), previous, forest);
```
kaituo, line 301: You are imputing for all of the gaps. If the number of gaps are huge, this will take a lot of time. Anyway to prevent that?

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/transform/SubtractMATransformer.java
```diff
@@ -32,4 +32,18 @@ public SubtractMATransformer(double[] weights, Deviation[] deviations) {
...
+    @Override
+    public void invertForecastRange(RangeVector ranges, int baseDimension, double[] previousInput,
+                                    double[] correction) {
+        super.invertForecastRange(ranges,baseDimension,previousInput,correction);
+        int horizon = ranges.values.length / baseDimension;
+        int inputLength = weights.length;
+        for (int i = 0; i < horizon; i++) {
+            for (int j = 0; j < inputLength; j++) {
+                ranges.shift(i * baseDimension + j,
+                        (float) (0.5 * getDrift(j, deviations)));
```
kaituo, line 45: Why do we need to add half of the drift (which I understand is the mean of step difference) to all values in the horizon?

kaituo (commented):
finished:
https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/cfe1169c5df0db9dffc549c5e305a4b953eca45a
https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/dc0565adfa57ce13eba256a81760f33a7970980e

Unfinished:
https://github.com/aws/random-cut-forest-by-aws/commit/c8721d4a2b5f7c513153fe1d9279ac68ff546c9d#diff-d64fe969bf928b2d27a559c3f1a31c31b2774774e64aa0c6acc2148b06ba78bf (next: Java/core/src/main/java/com/amazon/randomcutforest/returntypes/InterpolationMeasure.java)

https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/4bc8e2d9dc4232eb49d7acdead0ca2b532de71f2 (haven't started)

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
sudiptoguha: Good q. The first part  corresponds to missingValues != null && imputationMethod != RCF.

This means that we have a partial current tuple. We cannot define next/linear for partial current tuples -- even though the next/linear is applied to the intervening tuples (hence minus 1). These tuples are imputed fully (not partial). The first step above is very cheap operation. The second block can use any method (including RCF).

Then finally we come back to current tuple. If there were indeed missing values (ignoreImpute = false) then we need to impute again (specially for RCF) -- and this impute is not (necessarily) cheap. Observe that previous could have changed due to the intervening block

The reason for separating into three parts was that we could have complete control and restore the changes if we ever wanted to. Because anomalyScore() can be posed on a partial tuple; and then update occurs with a different tuple.

All that being said, this does look clunky. Let me see if there ia an easier way. But we do have to separate the invocations where we make up a fully missing record and where we know some values (current tuple).

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
sudiptoguha: Good q. It would definitely take a long time if we use RCF (and that is indeed better). Maybe we should have alternate logic if the gap is very large. (Also if the gap is very large then we are restarting froma completely fake shingle ...). Let me get back on this.

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/transform/SubtractMATransformer.java
sudiptoguha: This seems to be the gap left over from the general case even for a simple linear case -- because we updated the moving average (before calling this code) and some of the step difference then was already added. Note that this corresponds to what was shifted. But SUBTRACT_MA seems to not be a very stable transformation -- added  that to TransformTest  and then took it away (because the test was too non-deterministic), see @EnumSource(value = TransformMethod.class, names = {"NORMALIZE", "NORMALIZE_DIFFERENCE",  "DIFFERENCE"}).

I'd recommend using NORMALIZE or NORMALIZE_DIFFERENCE -- but added this just so that we have captured correction.

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
sudiptoguha: Changed -- added the fast forward. But I think we need to move more changes to any subsequent PRs ... this seems to be much better than the 3.8 version of the preprocessor already.

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
sudiptoguha: Done. But let's move any more changes to impute processor to subsequent PRs. This new refactor is already much better than the 3.8 version (and has much higher test coverage).

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/transform/SubtractMATransformer.java
sudiptoguha: I think it makes sense to drop this change and just keep the discussion that Normalize/Normalize_difference are the recommended ones and being tested more.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/state/ThresholdedRandomCutForestState.java
```diff
@@ -22,10 +22,10 @@
 import lombok.Data;
 
 import com.amazon.randomcutforest.parkservices.state.predictorcorrector.PredictorCorrectorState;
-import com.amazon.randomcutforest.parkservices.state.preprocessor.PreprocessorState;
 import com.amazon.randomcutforest.parkservices.state.returntypes.ComputeDescriptorState;
 import com.amazon.randomcutforest.parkservices.state.threshold.BasicThresholderState;
 import com.amazon.randomcutforest.state.RandomCutForestState;
+import com.amazon.randomcutforest.state.preprocessor.PreprocessorState;
```
kaituo, line 28: Do you need to change ThresholdedRandomCutForestState.version to 4.0? I also saw other places have 3.x version.

(base) kaituo@88665a53bc93 random-cut-forest-by-aws % ag V3_
Java/core/src/main/java/com/amazon/randomcutforest/state/tree/AbstractNodeStoreMapper.java
64:        state.setVersion(Version.V3_0);

Java/core/src/main/java/com/amazon/randomcutforest/state/tree/RandomCutTreeMapper.java
67:        state.setVersion(Version.V3_0);

Java/core/src/main/java/com/amazon/randomcutforest/state/Version.java
21:    public static final String V3_0 = "3.0";
22:    public static final String V3_5 = "3.5";
23:    public static final String V3_7 = "3.7";
24:    public static final String V3_8 = "3.8";

Java/core/src/main/java/com/amazon/randomcutforest/state/store/PointStoreMapper.java
59:        if (!state.getVersion().equals(Version.V3_0)) {
97:        state.setVersion(Version.V3_0);

Java/serialization/src/main/java/com/amazon/randomcutforest/serialize/json/v2/V2StateToV3ForestConverter.java
73:        if (!state.getVersion().equals(Version.V3_0)) {

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/state/RCFCasterState.java
18:import static com.amazon.randomcutforest.state.Version.V3_8;
27:    private String version = V3_8;

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/state/ThresholdedRandomCutForestState.java
18:import static com.amazon.randomcutforest.state.Version.V3_8;
35:    private String version = V3_8;

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/state/predictorcorrector/PredictorCorrectorState.java
18:import static com.amazon.randomcutforest.state.Version.V3_8;
32:    private String version = V3_8;

Do you need to change all of them to 4.0?

kaituo (commented):
finished:
https://github.com/aws/random-cut-forest-by-aws/commit/cfe1169c5df0db9dffc549c5e305a4b953eca45a
https://github.com/aws/random-cut-forest-by-aws/commit/dc0565adfa57ce13eba256a81760f33a7970980e
https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/b11fcb0124c581c8b0c1d0a49c8743f5316d98bd
https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/eba5608ede24f1184bc543c2a1e8ec161ef44eaf
https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/b5b17cb99389004c9337532ec0294adc5c68068b

Unfinished:
https://github.com/aws/random-cut-forest-by-aws/commit/c8721d4a2b5f7c513153fe1d9279ac68ff546c9d#diff-d64fe969bf928b2d27a559c3f1a31c31b2774774e64aa0c6acc2148b06ba78bf (next: Java/core/src/main/java/com/amazon/randomcutforest/returntypes/InterpolationMeasure.java)

https://github.com/aws/random-cut-forest-by-aws/commit/4bc8e2d9dc4232eb49d7acdead0ca2b532de71f2 

https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/9770d6291476625f75313d338b38c6f30f62a63f

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
```diff
@@ -0,0 +1,404 @@
...
+    protected float[] generateShingle(double[] inputTuple, long timestamp, int[] missingValues, double averageGap,
+            boolean changeForest, RandomCutForest forest) {
+        long lastInputTimeStamp = previousTimeStamps[shingleSize - 1];
+        double[] input = Arrays.copyOf(inputTuple, inputLength);
+        double[] previous = getShingledInput(shingleSize - 1);
+        double[] savedInput = Arrays.copyOf(previous, inputLength);
+        int numberToImpute = determineGap(timestamp - lastInputTimeStamp, averageGap) - 1;
+
+        if (imputationMethod != RCF || !forest.isOutputReady()) {
+            ImputationMethod method = (imputationMethod == RCF) ? DEFAULT_DYNAMIC : imputationMethod;
```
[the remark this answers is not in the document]
sudiptoguha: Oops - fixed. method should be used below (for imputationMethod because we are covering for the case the forest is not yet ready -- and thus we cannot have RCF imputation).

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/state/ThresholdedRandomCutForestState.java
sudiptoguha: No - we do not need to change them -- we should keep them for now for both debugging and history. The logic we have followed is that when the state class is altered then we update field. Now if the state class contains a POJO which is altered then it allows us to debug where the failure is -- so in fullness of time we would probably have all of the classes changed to 4.0. But that would need care so that we do not break backward compatibility -- for example the point store mapper uses the version strongly (and changing that would need more thought -- thankfully the backward compatibility tests are now in the core as well).   I think taking these on at this point would increate more unknowns. That being said, the counterpoint -- we are making a large change, this may be the moment -- that is also true. I think if we feel strongly, we should change in subsequent PRs where we would be able to debug smaller changes!

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
```diff
@@ -0,0 +1,519 @@
...
+            for (int y = 0; y < lastInputs.length; y++) {
+                errorTuple[y] = (float) lastInputs[y];
+            }
+            for (int i = 0; i < forecastHorizon; i++) {
+                if (sequenceIndex > i) {
+                    for (int j = 0; j < inputLength; j++) {
+                        RangeVector a = pastForecasts[inputIndex];
+                        int offset = i * inputLength;
+                        errorTuple[lastInputs.length] = i;
+                        errorTuple[lastInputs.length + 1] = forecastHorizon - i;
```
kaituo, line 247: Why do you need to store both i and forecastHorizon - i ?

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
```diff
@@ -0,0 +1,519 @@
...
+            checkArgument(builder.lowerLimit.get().length == inputLength, "incorrect length");
+            for (int y = 0; y < inputLength; y++) {
+                checkArgument(builder.lowerLimit.get()[y] <= upperLimit[y], "incorrect limits");
+            }
+            lowerLimit = Arrays.copyOf(builder.lowerLimit.get(), inputLength);
+        } else {
+            lowerLimit = new float[inputLength];
+            Arrays.fill(lowerLimit, -Float.MAX_VALUE);
+        }
+        // uses lastInputs as a markov input, the +1 corresponds to lookahead and then
```
kaituo, line 131: You did +2 in the code.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
```diff
@@ -0,0 +1,519 @@
...
+    public ErrorHandler(int errorHorizon, int forecastHorizon, int sequenceIndex, double percentile, int inputLength,
+            float[] pastForecastsFlattened, float[] lastDataDeviations, double[] lastInput, Deviation[] deviations,
+            PredictiveRandomCutForest estimator, float[] auxiliary) {
+        checkArgument(forecastHorizon > 0, " incorrect forecast horizon");
+        checkArgument(errorHorizon >= forecastHorizon, "incorrect error horizon");
+        checkArgument(inputLength > 0, "incorrect parameters");
+        checkArgument(sequenceIndex >= 0, "cannot be negative");
+        checkArgument(Math.abs(percentile - 0.25) < 0.24, "has to be between (0,0.5) ");
+        checkArgument(deviations.length == 3 * inputLength * forecastHorizon, "incorrect length");
+        checkArgument(lastInput.length == 2 * inputLength, "incorrect length");
```
kaituo, line 157: Why do you keep last 2 input instead of 1? I know more recorded value in the estimator/model are better.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
```diff
@@ -0,0 +1,519 @@
...
+                for (int y = 0; y < errorRMSE.high.length; y++) {
+                    errorHigh[y] = (float) errorRMSE.high[y];
+                    errorLow[y] = (float) errorRMSE.low[y];
+                }
+            }
+        }
+        // a control loop
+        for (int y = 0; y < intervalPrecision.length; y++) {
+            if (intervalPrecision[y].getMean() < 1.0 - percentile) {
+                errorHigh[y] = (float) max(1.0, 1.0 / (intervalPrecision[y].getMean() + 0.1)) * errorHigh[y];
```
kaituo, line 349: Should we add percentile instead of hardcoded 0.1 ?

kaituo (commented):
finished:
https://github.com/aws/random-cut-forest-by-aws/commit/cfe1169c5df0db9dffc549c5e305a4b953eca45a
https://github.com/aws/random-cut-forest-by-aws/commit/dc0565adfa57ce13eba256a81760f33a7970980e
https://github.com/aws/random-cut-forest-by-aws/commit/b11fcb0124c581c8b0c1d0a49c8743f5316d98bd
https://github.com/aws/random-cut-forest-by-aws/commit/eba5608ede24f1184bc543c2a1e8ec161ef44eaf
https://github.com/aws/random-cut-forest-by-aws/commit/b5b17cb99389004c9337532ec0294adc5c68068b
https://github.com/aws/random-cut-forest-by-aws/commit/c8721d4a2b5f7c513153fe1d9279ac68ff546c9d
https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/7a7ebd196b3eb80807a9bbcf345d0170a50ab7b6
https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/3ac6fff8dbf825e9da49fe578470577d1192872b

Unfinished:

https://github.com/aws/random-cut-forest-by-aws/commit/4bc8e2d9dc4232eb49d7acdead0ca2b532de71f2

https://github.com/aws/random-cut-forest-by-aws/commit/9770d6291476625f75313d338b38c6f30f62a63f

https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/1434836710ca04893e06a4395b9bf3125a05018a (read until Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java)

https://github.com/aws/random-cut-forest-by-aws/pull/401/commits/306727ee1b8fb5b44955b643d83e3e3165fa0984

Java/core/src/main/java/com/amazon/randomcutforest/PredictiveRandomCutForest.java
```diff
@@ -0,0 +1,565 @@
...
+     * @param timestamp  the timestamp
+     * @return a score in the range (0:log_2(sampleSize)) where larger values
+     *         indicate potential anomalies. Note that the score can remain elevated
+     *         due to shingling for a while -- use
+     *         ParkServices/ThresholdedRandomCutForest where the PredictorCorrector
+     *         architecture is used to evaluate the score further and produce a more
+     *         refined anomalyGrade()in [0:1] ; including setting the grade to 0
+     *         (corrector) even though the score (of this function) is high.
+     */
+    public double getExpectedInverseDepthScore(float[] inputPoint, long timestamp) {
```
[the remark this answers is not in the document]
sudiptoguha: ThresholdedRandomCutForest (TRCF, for now) is geared towards time series -- it uses the predictor-corrector to flag anomalies, and uses the single interface process() that scores as well as updates -- which is significantly more efficient. For time series AD, TRCF makes sense (as is being used in OpenSearch at the moment I believe). For forecasting, the extension of TRCF, RCFCaster would be natural.


However there has always been questions about more control regarding updates and scoring -- not all such usages are clearly reasonable. For example updating on point X; and then scoring point X creates a dependency that is not information theoretically clean (or explainable, etc., because X can be a multitude of points and some were seen before, some were not etc). But the question is : should the controls exist (under the generous assumption that the calling function knows what it is doing)? It seems so -- and so if there is a desire to intersperse updates and scores, then this class would allow it -- whereas TRCF would not (and force the calling function to respect arrow of time/seen vs unseen data).

Java/core/src/main/java/com/amazon/randomcutforest/state/preprocessor/PreprocessorMapper.java
```diff
@@ -47,7 +49,7 @@ public Preprocessor toModel(PreprocessorState state, long seed) {
                 .transformMethod(TransformMethod.valueOf(state.getTransformMethod()))
                 .startNormalization(state.getStartNormalization()).useImputedFraction(state.getUseImputedFraction())
                 .timeDeviations(timeStampDeviations).deviations(deviations).dataQuality(dataQuality)
-                .timeDecay(state.getTimeDecay());
+                .transformDecay(state.getTimeDecay());
```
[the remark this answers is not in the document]
sudiptoguha: Preprocessors do not sample but there is a natural notion of decaying effect of the transformation. In earlier versions the same  word "time decay" was used for samplers and preprocessors. This change clarifies the name -- we are not assigning time decay to transform decay -- the calling function from ThresholdedRCF used different settings for "time decay" when invoking samplers and preprocessors (and in this case, the variable was transformDecay). Note that sometimes they are the same as default if no transform decay is provided ...

Java/core/src/main/java/com/amazon/randomcutforest/state/preprocessor/PreprocessorMapper.java
```diff
@@ -82,14 +85,15 @@ public PreprocessorState toState(Preprocessor model) {
         state.setStopNormalization(model.getStopNormalization());
         state.setPreviousTimeStamps(model.getPreviousTimeStamps());
         state.setLastShingledInput(model.getLastShingledInput());
-        state.setLastShingledPoint(model.getLastShingledPoint());
+        state.setLastShingledPoint(toDoubleArray(model.getLastShingledPoint()));
         state.setValuesSeen(model.getValuesSeen());
         state.setInternalTimeStamp(model.getInternalTimeStamp());
         DeviationMapper deviationMapper = new DeviationMapper();
-        state.setTimeDecay(model.getTimeDecay());
+        state.setTimeDecay(model.getTransformDecay());
```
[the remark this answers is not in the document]
sudiptoguha: Hopefully answered previously.

Java/core/src/main/java/com/amazon/randomcutforest/statistics/Deviation.java
```diff
@@ -84,6 +84,8 @@ public double getDiscount() {
     }
 
     public void setDiscount(double discount) {
+        checkArgument(discount >= 0, "cannot be negative");
+        checkArgument(discount < 1, "can be at most 1");
```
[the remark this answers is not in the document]
sudiptoguha: Stochastic discount of 1 means that every past value other than the current value has weight 0 (or does not exist/matter) because the weights decrease vy a factor of (1-discount). A discount of 1 does not seem to apply for these time series problems.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
```diff
@@ -59,6 +56,13 @@ public static class Builder extends ThresholdedRandomCutForest.Builder<Builder>
         double percentile = DEFAULT_ERROR_PERCENTILE;
         protected Calibration calibrationMethod = DEFAULT_CALIBRATION;
 
+        // default is to use less space
+        protected boolean useRCF = false;
```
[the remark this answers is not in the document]
sudiptoguha: Manually maintained = computed automatically from the observed values (without any dependence on lookahead) :) Difficult to say -- the experiments in the example say that useRCF gives a better error for the shorter lookahead predictions. But it is also slower (about 2x) as well as requires more space; and for longer lookahead the default is already ok. So it made sense to keep the default false. But if one wishes more precise short lookahead error predictions then this should be turned on.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -459,14 +597,13 @@ public static class Builder<T extends Builder<T>> {
...
         protected Optional<Integer> threadPoolSize = Optional.empty();
-        protected Precision precision = DEFAULT_PRECISION;
         protected double boundingBoxCacheFraction = DEFAULT_BOUNDING_BOX_CACHE_FRACTION;
         protected int shingleSize = DEFAULT_SHINGLE_SIZE;
         protected Optional<Boolean> internalShinglingEnabled = Optional.empty();
         protected double initialAcceptFraction = DEFAULT_INITIAL_ACCEPT_FRACTION;
         protected double anomalyRate = 0.01;
         protected TransformMethod transformMethod = TransformMethod.NONE;
-        protected ImputationMethod imputationMethod = PREVIOUS;
+        protected ImputationMethod imputationMethod = RCF;
```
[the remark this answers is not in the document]
sudiptoguha: But RCF is better in most cases :)

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -170,14 +171,31 @@ public ThresholdedRandomCutForest(ForestMode forestMode, TransformMethod transfo
...
+    // the function process() would be able to provide an anomaly score (which is
+    // likely near
+    // minimum, since RCF is used to fill in the missing values). As a result, high
+    // values of the
+    // anomaly score will continue to be likely anomalies.
+    // Note that the basic RandomCutForest cannot be changed easily
+    // but the process() function would only require a fraction of the input
+    // see ThresholdedRandomCutForestMapperTest
+    public ThresholdedRandomCutForest(RandomCutForest forest, double futureAnomalyRate, List<Double> values,
+            double[] lastShingledInput) {
```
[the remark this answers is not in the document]
sudiptoguha: Null/double[0] will likely cause exception -- the length has to be inputDimension*shingleSize (say 2 and 5). Then using new double [10] (all 0 values) or just pretending that the last known values are repeated shingleSize times are probably the best available options (need not be good options! -- if resources permit, it may be better to rebuild models, but if that is not an option the 0's or last known values are the ways to go).

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -432,6 +442,134 @@ public void setInitialThreshold(double initial) {
...
+                int index = result.getRelativeIndex();
+                if (index < 0) {
+                    reference = preprocessor.getShingledInput(shingleSize + index);
+                    result.setPastTimeStamp(preprocessor.getTimeStamp(shingleSize + index));
+                }
+                result.setPastValues(reference);
+                if (newPoint != null) {
+                    double[] values = preprocessor.getExpectedValue(index, reference, point, newPoint);
+                    if (forestMode == ForestMode.TIME_AUGMENTED) {
+                        int endPosition = (shingleSize - 1 + index + 1) * dimension / shingleSize;
```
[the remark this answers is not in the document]
sudiptoguha: Yes - but I kept the +1 and -1 so that it fits the pattern with the non time-augmented case. There were a bunch of errors while changing the code -0- this is redundant but maybe more explanatory.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -432,6 +442,134 @@ public void setInitialThreshold(double initial) {
...
+                        double timeGap = (newPoint[endPosition - 1] - point[endPosition - 1]);
+                        long expectedTimestamp = (timeGap == 0) ? result.getInputTimestamp()
```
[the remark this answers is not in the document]
sudiptoguha: No No -- the time gap here corresponds to the question "was the anomaly (likely) caused by a difference in the input timestamp which is in coordinate (end position - 1)". If time gap == 0, then that means that the input timestamp was not a factor and will use the input timestamp for the expected timestamp. If the gap is not 0, then we have to compute inverses etc. The reason for doing this was that we use single precision float internally, so if we always use the inverses then we will have 0.9999, 2.000001 like values instead of 1, 2 etc... That can be very annoying (and can cause more trouble). In fact this is the same reason (not to have floating point approximations all over the place) I decided to keep the inputs double precision in several places. The approximations still show up sometimes -- but are few and far between.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -170,14 +171,31 @@ public ThresholdedRandomCutForest(ForestMode forestMode, TransformMethod transfo
         this.lastAnomalyDescriptor = descriptor;
     }
 
-    // for conversion from other thresholding models
-    public ThresholdedRandomCutForest(RandomCutForest forest, double futureAnomalyRate, List<Double> values) {
+    // this constructor produces an internally shingled ThresholdedRCF model from an
+    // externally shingled ThresholdedRCF model (which may or may not be in use in
```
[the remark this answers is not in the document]
sudiptoguha: Well, version 3.8 allowed for externally shingled TRCF -- we are simplifying that. If the RCF underpinning the TRCF was externally shingled then there is n to much we can do (other than hope the opportunistic compression in RCF kicks in and mitigates external shingling setting and saves space anyways). None of the code used in previous versions should have been using TRCFs with external shingling ... but then if someone kicked off such a variant then here is a conversion.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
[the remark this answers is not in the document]
sudiptoguha: Ok - changing -- the entire block has a lot of "explanatory" repetition ....

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -432,6 +442,134 @@ public void setInitialThreshold(double initial) {
...
+                    result.setPastTimeStamp(preprocessor.getTimeStamp(shingleSize + index));
+                }
+                result.setPastValues(reference);
+                if (newPoint != null) {
+                    double[] values = preprocessor.getExpectedValue(index, reference, point, newPoint);
+                    if (forestMode == ForestMode.TIME_AUGMENTED) {
+                        int endPosition = (shingleSize - 1 + index + 1) * dimension / shingleSize;
+                        double timeGap = (newPoint[endPosition - 1] - point[endPosition - 1]);
+                        long expectedTimestamp = (timeGap == 0) ? result.getInputTimestamp()
+                                : (long) values[dimension / shingleSize - 1];
```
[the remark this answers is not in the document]
sudiptoguha: changed

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -432,6 +442,134 @@ public void setInitialThreshold(double initial) {
...
+                        if (index < 0) {
+                            expectedTimestamp = (timeGap == 0) ? preprocessor.getTimeStamp(shingleSize - 1 + index)
+                                    : (long) values[dimension / shingleSize - 1];
```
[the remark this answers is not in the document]
sudiptoguha: changed

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
sudiptoguha: It seemed to set up a better constrained optimization -- basically it used the fact that RCF has to work harder to present diagonal lines ;)

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
sudiptoguha: No No -- the hardcoded 0.1 is a "step size" -- if the interval precision was ruined (data shift or something) then it would not be great to make large changes when intervalPrecision[y] (approx) == 0.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
sudiptoguha: That seemed to help the error. Suppose we were using 1 variable -- then the error has two components high and low. But the current variable and lookahead are only 2 pieces of information -- so we were imputing 2 out of 4. With twice the input length, this makes imputing 2 out of 5 -- so the noise was lessened. Did not seem to have any benefit from going to thrice ...

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
sudiptoguha: Yes - it's lookahead and forecastHorizon - lookahead, as we discussed in previous comment. You're correct the comment needs to be changed. Let's come back to this after we have gone though the other code? Otherwise it seems I am generating too many commits :)

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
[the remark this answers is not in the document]
kaituo: I meant the first parameter is RandomCutForest, not TRCF. Should we change wording to "externally shingled RandomCutForest' from "externally shingled ThresholdedRCF"?

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -170,14 +171,31 @@ public ThresholdedRandomCutForest(ForestMode forestMode, TransformMethod transfo
...
+    // but the process() function would only require a fraction of the input
+    // see ThresholdedRandomCutForestMapperTest
+    public ThresholdedRandomCutForest(RandomCutForest forest, double futureAnomalyRate, List<Double> values,
+            double[] lastShingledInput) {
         this.forest = forest;
         int dimensions = forest.getDimensions();
-        int inputLength = (forest.isInternalShinglingEnabled()) ? dimensions / forest.getShingleSize()
-                : forest.getDimensions();
+
+        int inputLength = dimensions / forest.getShingleSize();
```
kaituo, line 194: We are currently operating under the assumption that the input to the RandomCutForest is internally shingled. However, this assumption could disrupt Anomaly Detection (AD) functionalities, as we utilize this constructor to maintain backward compatibility with the single-stream AD that relies on external shingling. To address this issue in AD, we would need to undertake a comprehensive overhaul, including redoing the cold start process and modifying all areas related to external shingling (such as imputation). However, this is not a straightforward change and would require a significant amount of work.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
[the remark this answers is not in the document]
kaituo: Not sure if I understand. I will ask offline.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
kaituo: don't understand. Will ask offline.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/PredictorCorrector.java
```diff
@@ -705,10 +709,12 @@ protected <P extends AnomalyDescriptor> P detect(P result, RCFComputeDescriptor
...
 
         if (workingGrade > 0 && lastDescriptor != null) {
-            if (score > lastDescriptor.getRCFScore()
-                    || lastDescriptor.getRCFScore() - lastDescriptor.getThreshold() > score
-                            - max(workingThreshold, lastDescriptor.getThreshold())
-                                    * (1 + max(0.2, runLength / (2.0 * max(10, shingleSize))))) {
+            if (score > lastDescriptor.getRCFScore()) {
+                candidate = true;
+            }
+            double runDiscount = max(workingThreshold, lastDescriptor.getThreshold())
```
kaituo, line 715: Do you want to add an else as you only want to do the following when score <= lastDescriptor.getRCFScore() ?

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
sudiptoguha: No no .. we do not change the random cut forest -- because any change could affect the cuts and corrupt the trees. To change the RCF we are better of extracting all points and then rebuilding them. But for RCFs external and internal shingling executes the same code and would perform opportunistic compression.  The forest is NOT changed -- 

this.forest = forest;

If this discussion is holding the PR up, then let's proceed. Note that a similar function existed earlier. This is an improvement and we should move on. What this does is *externally shingled TRCF* is converted to *internally shingled TRCF* and RCF is managed on its own -- the user should stop worrying about that. But if that is still a concern then this function should not be used, period. This is why I said this *may/may not* be helpful based on internals of what exists in prod currently.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/PredictorCorrector.java
```diff
@@ -705,10 +709,12 @@ protected <P extends AnomalyDescriptor> P detect(P result, RCFComputeDescriptor
...
+                    * (1 + max(0.2, runLength / (2.0 * max(10, shingleSize))));
+            if (lastDescriptor.getRCFScore() - lastDescriptor.getThreshold() > score - runDiscount) {
```
kaituo, line 717: Can you explain the intuition behind the condition? It looks like you are detecting decreasing gap by making the threshold larger and subtracting larger value. But the rationale is opaque to me.

kaituo (commented):
finished:
https://github.com/aws/random-cut-forest-by-aws/commit/cfe1169c5df0db9dffc549c5e305a4b953eca45a
https://github.com/aws/random-cut-forest-by-aws/commit/dc0565adfa57ce13eba256a81760f33a7970980e
https://github.com/aws/random-cut-forest-by-aws/commit/b11fcb0124c581c8b0c1d0a49c8743f5316d98bd
https://github.com/aws/random-cut-forest-by-aws/commit/eba5608ede24f1184bc543c2a1e8ec161ef44eaf
https://github.com/aws/random-cut-forest-by-aws/commit/b5b17cb99389004c9337532ec0294adc5c68068b
https://github.com/aws/random-cut-forest-by-aws/commit/c8721d4a2b5f7c513153fe1d9279ac68ff546c9d
https://github.com/aws/random-cut-forest-by-aws/commit/7a7ebd196b3eb80807a9bbcf345d0170a50ab7b6
https://github.com/aws/random-cut-forest-by-aws/commit/3ac6fff8dbf825e9da49fe578470577d1192872b
https://github.com/aws/random-cut-forest-by-aws/commit/4bc8e2d9dc4232eb49d7acdead0ca2b532de71f2
https://github.com/aws/random-cut-forest-by-aws/commit/9770d6291476625f75313d338b38c6f30f62a63f
https://github.com/aws/random-cut-forest-by-aws/commit/306727ee1b8fb5b44955b643d83e3e3165fa0984

Unfinished:
https://github.com/aws/random-cut-forest-by-aws/commit/1434836710ca04893e06a4395b9bf3125a05018a (read until Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java)

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
sudiptoguha: Ok - I think I see the disconnect. Changing the comment.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/calibration/ErrorHandler.java
sudiptoguha: Done

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/PredictorCorrector.java
sudiptoguha: Actually no -- and that relates to the comment below (fair point about opacity). Even for a single anomaly what often happens is that the score is high as soon as we see the value -- then score continues to increase till about middle of the shingle (all these are very unlikely patterns) and then the score falls off. So the score actually can keep rising even though no fresh anomaly has been seen. However what happens is that law of marginal increase takes over -- the jump in score when we first see the anomaly is typically higher, and unless there are subsequent anomalies then the marginal increase in score is less as the anomaly moves inward. The threshold is however a lagging computation and peaks much later. So this seemed to eliminate the effect of raised scores due to single anomalies -- but there can be other ways.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/PredictorCorrector.java
sudiptoguha: See above.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/returntypes/RCFComputeDescriptor.java
```diff
@@ -53,15 +54,15 @@ public class RCFComputeDescriptor extends Point {
...
     int inputLength;
 
     // sequence index (the number of updates to RCF) -- it is possible in imputation
     // that
     // the number of updates more than the input tuples seen by the overall program
     long totalUpdates;
 
-    // determines if values can be inputed and or expected point calculated
+    // determines if values can be input and or expected point calculated
     boolean reasonableForecast;
```
kaituo, line 66: Can you explain more about this field? When do you use reasonableForecast?

kaituo (commented):
Finished reading all commits.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/PredictorCorrector.java
sudiptoguha: Fair point. A single execution path may also (i) help debug as well as (ii) decide if the second part is truly important.
Changed.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -99,7 +101,7 @@ public ThresholdedRandomCutForest(Builder<?> builder) {
             // force internal shingling for this option
             builder.internalShinglingEnabled = Optional.of(true);
         } else if (builder.forestMode == ForestMode.STREAMING_IMPUTE) {
-            checkArgument(builder.shingleSize > 1, " shingle size 1 is not useful in impute");
+            // already validated
```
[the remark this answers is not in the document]
sudiptoguha: We could -- but then the STANDARD mode can already do that (fill in missing values in partial tuples, even when shingle size = 1). Shingle size = 1 stands for the fact that the elements are *disconnected/loosely coupled in time*.  Shingle size =1 is actually at odds with the notion of time series (where the series is conceptually connected in time). The very act of impute based on time (imputing intervening values) means that there is a connection and we can leverage it. That being said, let's discuss -- if this is a showstopper for some configuration space and creates complication then we can leave this as a comment. Although the problem I see is that down the road these algorithms with shingle size =1 would be used to infer things about time series and would likely not be accurate.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/returntypes/RCFComputeDescriptor.java
sudiptoguha: Great q :) We do not use it in any more ThresholdedRCF (TRCF) -- it was used in the prior versions when the forecast was more rudimentary. It does make sense that "we can say it's not right but we cannot quite say what the correct course is" -- but I think that nuance can be too much. Note that the predict() function now provides a clustered nearest neighbors and can be useful in answering the nuanced question. In TRCF, if the dimension of the space is 4 then we can forecast reasonably. Also when we have seen enough data for anomaly scoring (outputAfter) and transformations being stable -- then perhaps this field is not useful. I did not what to remove the field to break backward compatibility (but happy to). We should also discuss this restriction of 4. I did feel strongly that the calling function should be cognizant of the quality of forecast -- but it it true that such a decision can create a lot more work for the calling function (and a simpler solution is to advice users to not have 1 dimension and shingle size less than 4, of something).

kaituo (approved):

jotok (approved):

PR #404, opened 2024-07-08
Add Python Wrapper for RCF and Fix Error Message

kaituo:
*Issue #, if available:*
- Added a Python wrapper to enable calling RCF from Python.
- Fixed an error message that previously referenced the wrong variable.

Testing:
- Manually tested the Python binding to ensure it works correctly.


greaa-aws (approved):
Looks good to me, matches my code for python bindings.

Only point I'd raise is high-level: why does TRCF only have a `process()` method while RCF has `score()`, `update()`, etc? Can we add a little more detail to what `TRCF.process()` does - is it basically "score and update"? 

In a future commit I think we'd also want to add more on the documentation side, especially regarding the RCF / TRCF parameters and their meaning.

kaituo:
> Looks good to me, matches my code for python bindings.
> 
> Only point I'd raise is high-level: why does TRCF only have a `process()` method while RCF has `score()`, `update()`, etc? Can we add a little more detail to what `TRCF.process()` does - is it basically "score and update"?
> 
> In a future commit I think we'd also want to add more on the documentation side, especially regarding the RCF / TRCF parameters and their meaning.

process does more than score/update. It includes all of the pre and post processing besides score/update. I added more description on the process method parameters.

PR #407, opened 2024-07-31
Fix Confidence Adjustment for Larger Shingle Sizes

kaituo:
*Description of changes:*

This PR addresses further adjustments to the confidence calculation issue discussed in [PR 405](https://github.com/aws/random-cut-forest-by-aws/pull/405). While PR 405 successfully resolved the issue for a shingle size of 4, it did not achieve the same results for larger shingle sizes like 8.

Key Changes
1. Refinement of seenValues Calculation:
* Previously, the formula increased confidence even as numImputed (number of imputations seen) increased because seenValues (all values seen) also increased.
* This PR fixes the issue by counting only non-imputed values as seenValues.
2. Upper Bound for numImputed:
* The numImputed is now upper bounded to the shingle size.
* The impute fraction calculation, which uses numberOfImputed * 1.0 / shingleSize, now ensures the fraction does not exceed 1.
3. Decrementing numberOfImputed:
* The numberOfImputed is decremented when there is no imputation.
* Previously, numberOfImputed remained unchanged when there is an imputation as there was both an increment and a decrement, keeping the imputation fraction constant. This PR ensures the imputation fraction accurately reflects the current state. This adjustment ensures that the forest update decision, which relies on the imputation fraction, functions correctly. The forest is updated only when the imputation fraction is below the threshold of 0.5.

Testing
* Added test scenarios with various shingle sizes to verify the changes.

### DIFF
--- a/Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
+++ b/Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
@@ -81,44 +81,6 @@ public class ImputePreprocessor extends InitialSegmentPreprocessor {
     }
 
     /**
-     * the timestamps are now used to calculate the number of imputed tuples in the
-     * shingle
-     * 
-     * @param timestamp the timestamp of the current input
-     */
-    @Override
-    protected void updateTimestamps(long timestamp) {
-        /*
-         * For imputations done on timestamps other than the current one (specified by
-         * the timestamp parameter), the timestamp of the imputed tuple matches that of
-         * the input tuple, and we increment numberOfImputed. For imputations done at
-         * the current timestamp (if all input values are missing), the timestamp of the
-         * imputed tuple is the current timestamp, and we increment numberOfImputed.
-         *
-         * To check if imputed values are still present in the shingle, we use the first
-         * condition (previousTimeStamps[0] == previousTimeStamps[1]). This works
-         * because previousTimeStamps has a size equal to the shingle size and is filled
-         * with the current timestamp. However, there are scenarios where we might miss
-         * decrementing numberOfImputed:
-         *
-         * 1. Not all values in the shingle are imputed. 2. We accumulated
-         * numberOfImputed when the current timestamp had missing values.
-         *
-         * As a result, this could cause the data quality measure to decrease
-         * continuously since we are always counting missing values that should
-         * eventually be reset to zero. The second condition <pre> timestamp >
-         * previousTimeStamps[previousTimeStamps.length-1] && numberOfImputed > 0 </pre>
-         * will decrement numberOfImputed when we move to a new timestamp, provided
-         * numberOfImputed is greater than zero.
-         */
-        if (previousTimeStamps[0] == previousTimeStamps[1]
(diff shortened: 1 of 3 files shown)


Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
```diff
@@ -80,44 +80,6 @@ public float[] getScaledShingledInput(double[] inputPoint, long timestamp, int[]
...
-    @Override
-    protected void updateTimestamps(long timestamp) {
-        /*
-         * For imputations done on timestamps other than the current one (specified by
-         * the timestamp parameter), the timestamp of the imputed tuple matches that of
-         * the input tuple, and we increment numberOfImputed. For imputations done at
-         * the current timestamp (if all input values are missing), the timestamp of the
-         * imputed tuple is the current timestamp, and we increment numberOfImputed.
-         *
-         * To check if imputed values are still present in the shingle, we use the first
```
amitgalitz, line 98: At least this part of the comment is still valid right? Would be good to have as I was pretty confused what `previousTimeStamps[0] != previousTimeStamps[1]` really meant to do.
Additionally so I am understaind correct, we are saying if the last 10 values were imputed and shingle size is 8, we will most likely return false here and not allow an update until we get an actual value?

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
```diff
@@ -168,7 +142,9 @@ void updateForest(boolean changeForest, double[] input, long timestamp, RandomCu
         updateShingle(input, scaledInput);
         updateTimestamps(timestamp);
         if (isFullyImputed) {
-            numberOfImputed = numberOfImputed + 1;
+            numberOfImputed = Math.min(numberOfImputed + 1, shingleSize);
+        } else if (numberOfImputed > 0) {
+            numberOfImputed = numberOfImputed - 1;
```
amitgalitz, line 147: at this point are we guaranteed to have seen a new value that is not imputed?

amitgalitz:
> The numImputed is now upper bounded to the shingle size.

Do we want to do this because cause we only care for the fraction in relation to the current shingle?

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
kaituo: let me add the comment back. 

> we are saying if the last 10 values were imputed and shingle size is 8, we will most likely return false here and not allow an update until we get an actual value?

yes, your understanding is correct.

Java/core/src/main/java/com/amazon/randomcutforest/preprocessor/ImputePreprocessor.java
kaituo: yes.

kaituo:
> > The numImputed is now upper bounded to the shingle size.
> 
> Do we want to do this because cause we only care for the fraction in relation to the current shingle?

yes.

amitgalitz (approved):

PR #413, opened 2025-05-22
Allow processSequentially to accept explicit timestamps etc

kaituo:
*Issue #, if available:*

*Description of changes:*
This PR

* Updated ThresholdedRandomCutForest.processSequentially to accept explicit timestamps, enabling its use in STREAMING_IMPUTE mode.
* Additionally, corrected threshold initialization in ThresholdedRandomCutForest constructor: the lower RCF score threshold is now set only when autoAdjust is disabled. Previously, setting this threshold regardless of autoAdjust inadvertently disabled automatic adjustments in BasicThresholder.

Testing:
* added UT

### DIFF
--- a/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
+++ b/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
@@ -260,8 +260,10 @@ public class RCFCaster extends ThresholdedRandomCutForest {
         return processSequentially(data, timestamps, filter);
     }
 
-    public List<AnomalyDescriptor> processSequentially(double[][] data, long[] timestamps, Function<AnomalyDescriptor, Boolean> filter) {
+    public List<AnomalyDescriptor> processSequentially(double[][] data, long[] timestamps,
+            Function<AnomalyDescriptor, Boolean> filter) {
         // Precondition checks
+        checkArgument(filter != null, "filter must not be null");
         if (data != null && data.length > 0) {
             checkArgument(timestamps != null, "timestamps must not be null when data is non-empty");
             checkArgument(timestamps.length == data.length, String.format(Locale.ROOT,
@@ -286,6 +288,7 @@ public class RCFCaster extends ThresholdedRandomCutForest {
                     int length = preprocessor.getInputLength();
                     for (int i = 0; i < data.length; i++) {
                         double[] point = data[i];
+                        checkArgument(point != null, " data should not be null ");
                         checkArgument(point.length == length, " nonuniform lengths ");
                         ForecastDescriptor description = new ForecastDescriptor(point, timestamps[i], forecastHorizon);
                         augment(description);
--- a/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
+++ b/Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
@@ -20,6 +20,7 @@ import static java.lang.Math.max;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.function.Function;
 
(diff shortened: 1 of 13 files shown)


Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
```diff
@@ -246,6 +247,34 @@ public TimedRangeVector extrapolate(Calibration calibration, int horizon, boolea
...
     @Override
     public List<AnomalyDescriptor> processSequentially(double[][] data, Function<AnomalyDescriptor, Boolean> filter) {
+        if (data == null || data.length == 0) {
+            return new ArrayList<>();
+        }
+
+        long timestamp = preprocessor.getInternalTimeStamp();
+        long[] timestamps = new long[data.length];
+        for (int i = 0; i < data.length; i++) {
+            timestamps[i] = ++timestamp;
```
amitgalitz, line 257: are we just internally increasing timestamp by one for each next data point? Is that a correct assumption?

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
```diff
@@ -246,6 +247,34 @@ public TimedRangeVector extrapolate(Calibration calibration, int horizon, boolea
...
+        }
+
+        return processSequentially(data, timestamps, filter);
+    }
+
+    public List<AnomalyDescriptor> processSequentially(double[][] data, long[] timestamps, Function<AnomalyDescriptor, Boolean> filter) {
+        // Precondition checks
+        if (data != null && data.length > 0) {
```
amitgalitz, line 265: just wondering should we have checkArguments for these cases too like if data == null or we don't want to fail and we want to keep current behavior

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
```diff
@@ -254,11 +283,11 @@ public List<AnomalyDescriptor> processSequentially(double[][] data, Function<Ano
                     if (cacheDisabled) { // turn caching on temporarily
                         forest.setBoundingBoxCacheFraction(1.0);
                     }
-                    long timestamp = preprocessor.getInternalTimeStamp();
                     int length = preprocessor.getInputLength();
-                    for (double[] point : data) {
+                    for (int i = 0; i < data.length; i++) {
+                        double[] point = data[i];
```
amitgalitz, line 288: can maybe also add a new null check for point? reasonable to leave as is too though

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
amitgalitz: oh nevermind I see this is what we basically did before but now we are allowing other callers to pass in specific timestamps themselves

kaituo: yes

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
kaituo: Prefer to not fail and keep current behavior

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
kaituo: will add a null check

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
```diff
@@ -289,6 +294,65 @@ public AnomalyDescriptor process(double[] inputPoint, long timestamp, int[] miss
...
+     * the following function processes a list of vectors sequentially; the main
+     * benefit of this invocation is the caching is persisted from one data point to
+     * another and thus the execution is efficient. Moreover in many scenarios where
+     * serialization deserialization is expensive then it may be of benefit of
+     * invoking sequential process on a contiguous chunk of input (we avoid the use
+     * of the word batch -- the entire goal of this procedure is to provide
+     * sequential processing and not standard batch processing). The procedure
+     * avoids transfer of ephemeral transient objects for non-anomalies and thereby
+     * can have additional benefits. At the moment the operation does not support
+     * external timestamps.
```
amitgalitz, line 320: "At the moment the operation does not support external timestamps."

This can be removed now I think, also I am a little confused because we are saying this is not for batch processing but it is being called by SequentialAnalysis.detectAnomalies which makes it seem like its for a batch job

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
```diff
@@ -246,6 +247,34 @@ public TimedRangeVector extrapolate(Calibration calibration, int horizon, boolea
...
+        long timestamp = preprocessor.getInternalTimeStamp();
+        long[] timestamps = new long[data.length];
+        for (int i = 0; i < data.length; i++) {
+            timestamps[i] = ++timestamp;
+        }
+
+        return processSequentially(data, timestamps, filter);
+    }
+
+    public List<AnomalyDescriptor> processSequentially(double[][] data, long[] timestamps, Function<AnomalyDescriptor, Boolean> filter) {
```
amitgalitz, line 263: nit: adding a filter is not null check in preconditions

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/ThresholdedRandomCutForest.java
kaituo: Good catch. Removed.

Java/parkservices/src/main/java/com/amazon/randomcutforest/parkservices/RCFCaster.java
kaituo: added

amitgalitz (approved):

PR #414, opened 2025-09-10
feat(parkservices): preserve last descriptor fields etc

kaituo:
*Issue #, if available:*

*Description of changes:*

This PR made enhancements.
- `processSequentially`: before constructing each `AnomalyDescriptor`, scan the point for `Double.NaN` values and set `missingValues` to the indices of the missing features. This enables `processSequentially` to handle missing values.
- Preserve `currentInput` and `missingValues` during `ThresholdedRandomCutForest` serialization so the last processed descriptor survives save/restore. Enables downstream consumers to surface/impute per-feature missingness.

Testing
- Unit tests for TRCF round-trip serialization and NaN handling.
- Manual verification on sample streams.

### DIFF
--- a/Java/parkservices/src/test/java/com/amazon/randomcutforest/parkservices/TransformTest.java
+++ b/Java/parkservices/src/test/java/com/amazon/randomcutforest/parkservices/TransformTest.java
@@ -42,10 +42,20 @@ public class TransformTest {
             int numberOfTrees = 30 + rng.nextInt(20);
             int outputAfter = 32 + rng.nextInt(50);
             // shingleSize 1 is not recommended for complicated input
-            int shingleSize = 2 + rng.nextInt(15);
+            // The test sets alertOnce(true) to suppress cascades after a single injected spike.
+            // Current suppression only triggers for overlapping shingles: gap < shingleSize
+            // (in PredictorCorrector.detect), so “late” follow-on spikes (gap ≥ shingleSize)
+            // still raise grades.
+            // If shingleSize is small (e.g., 2), gap is small, anomaly would not be suppressed,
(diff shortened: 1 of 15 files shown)


Java/parkservices/src/test/java/com/amazon/randomcutforest/parkservices/TransformTest.java
```diff
@@ -42,10 +42,20 @@ public void AnomalyTest(TransformMethod method) {
...
+            // Current suppression only triggers for overlapping shingles: gap < shingleSize
+            // (in PredictorCorrector.detect), so “late” follow-on spikes (gap ≥ shingleSize)
+            // still raise grades.
+            // If shingleSize is small (e.g., 2), gap is small, anomaly would not be suppressed,
+            // may cause totalcount > numTrials.
+            int shingleSize = 3 + rng.nextInt(15);
             int baseDimensions = 1 + rng.nextInt(5);
             int dimensions = baseDimensions * shingleSize;
             long forestSeed = rng.nextLong();
+            System.out.println(" forestSeed " + forestSeed + " method " + method + " seed " + seed + " outputAfter "
```
amitgalitz, line 55: this is meant to stay for testing visibility purpose?

kaituo:
> currently now what happens when we have missing values, here I see we actually make sure to set in descriptor

processSequentially didn't work when we have missing values. Yes, I made sure to set in descriptor.

amitgalitz:
> > currently now what happens when we have missing values, here I see we actually make sure to set in descriptor
> 
> processSequentially didn't work when we have missing values. Yes, I made sure to set in descriptor.
got it

